Issue #17, opened 2021-02-04
Bootrom functions


thejpster:
Does the bootrom table lookup code work? It's just that according to Datasheet section 2.8.3, the Bootrom stores the table addresses as 16-bit values - these need to be read as 16-bit values, then expanded to 32-bit, then converted to the appropriate pointer. I suspect the code as written will read 32-bits from the bootrom, and get thus get the pointer very wrong.

AsafFisher:
Can you refer to the exact line where you see a 32-bit pointer read from the rom?

AsafFisher:
We use `rom_table_lookup` which is supplied by the bootrom, so if this function read 16-bit pointer value then we are fine :)

Hopefully it works like that. If not, well no pi pico has ever worked properly :O

AsafFisher:
<img width="983" alt="Screen Shot 2021-02-05 at 1 16 51 PM" src="https://user-images.githubusercontent.com/6796683/107027194-733bb880-67b4-11eb-830f-f767161e1acf.png">

thejpster:
```rust
const FUNC_TABLE: *const *const u16 = 0x14 as _;

rom_table_lookup(FUNC_TABLE, some_tag);

fn rom_table_lookup<T>(table: *const *const u16, tag: RomFnTableCode) -> T {
    unsafe {
        let rom_table_lookup: RomTableLookupFn<T> = core::mem::transmute(ROM_TABLE_LOOKUP_PTR);
        rom_table_lookup(*table, u16::from_le_bytes(tag) as u32)
    }
}

The deref of the table pointer with `*table`will read four bytes (a `*const _`) from memory: 0x14 .. 0x17. You should only read 0x14 and 0x15 and add two zero bytes. You should probably set FUNC_TABLE to be a *const u16, then do a core::ptr::volatile_read(), then cast the u16 to a usize, then pass that to the rom_table_function.

In your ASM listing, I think `LDR R0, [R0]` will do a 32-bit read from the address 0x14. That `LDR` should be an `LDRH`.

But maybe I'm wrong, and this code works fine as-is? Maybe I've read the datasheet wrong?

**Comment by thejpster** [comment]
Note how the SDK code does a 16-bit read:

```c
// Convert a 16 bit pointer stored at the given rom address into a 32 bit pointer
#define rom_hword_as_ptr(rom_address) (void *)(uintptr_t)(*(uint16_t *)rom_address)
```

AsafFisher:
[...]

PR #20, opened 2021-02-15
fix rom_table_lookup

jannic:
It seems like the rom_table_lookup missed one level of
pointer dereferencing.

After comparing it to the working call to reset_usb_boot() in
https://github.com/jannic/rp-microcontroller-rs/blob/master/util/rp2040-panic-usb-boot/src/lib.rs,
I changed the code until it generated basically the same assembly.

With that change, I was able to successfully call rom_data::reset_to_usb_boot()

I still don't like the type RomTableLookupFn, which just returns
some generic T without any checks, and I think rom_table_lookup
should be unsafe. But as none of those are pub, it doesn't matter too
much. So I just made the changes necessary to make the code work.

### DIFF
--- a/rp2040-hal/src/rom_data.rs
+++ b/rp2040-hal/src/rom_data.rs
@@ -144,9 +144,9 @@ pub fn copyright_string() -> &'static str {
 }
 
 /// The 8 most significant hex digits of the Bootrom git revision.
-pub fn git_revision() -> &'static str {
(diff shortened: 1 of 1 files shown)


thejpster (approved):
LGTM

thejpster:
This looks like it will close https://github.com/rp-rs/rp-hal/issues/17

PR #23, opened 2021-04-17
Add CI checks

Nitori-:
I have no idea if these are actually wanted, but I figured I'd open a PR for discussion at least. If they're desired here I can also add them to the other crates in the org.

These are pretty simple checks that I've used in other projects in the past.

Here's an example where they ran on my fork: https://github.com/Nitori-/rp-hal/runs/2370884453
I don't actually know if they'll run on this PR, or if they have to be in the main repo first in order to run against a PR


Cargo.lock
```diff
@@ -0,0 +1,118 @@
+# This file is automatically @generated by Cargo.
```
tdittr, line 1: As mentioned in the other PR: I think the lock file should only be added to binary crates.

tdittr (approved):
I think this is a great idea :+1: 
When we add unit tests we might want to add a `cargo test` run.

Nitori-:
Agreed on `cargo test`

Right now `cargo test` errors out, but once we've got unit tests it should be trivial to add

tdittr:
Oh yes I remember, for `cargo test` to work it needs to be compiled for the host instead of the pico. Although I don't remember how I did it back then :smile:

.github/workflows/check.yml
```diff
@@ -0,0 +1,15 @@
+on: [push, pull_request]
+name: CI Checks
+jobs:
+  check:
+    name: Check and Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
```
emgre, line 8: Using a fixed version to ensure nothing breaks in the future.

```suggestion
      - uses: actions/checkout@v2
```

.github/workflows/check.yml
```diff
@@ -0,0 +1,15 @@
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          target: thumbv6m-none-eabi
```
emgre, line 12: For a slightly faster execution time, this avoids installing `rust-docs` for nothing. I'm not sure the `override` is needed, but adding it makes us sure that it's using the version we want.

```suggestion
      - uses: actions-rs/toolchain@v1
        with:
          toolchain: stable
          target: thumbv6m-none-eabi
          override: true
          profile: minimal
          components: rustfmt, clippy
```

.github/workflows/check.yml
```diff
@@ -0,0 +1,15 @@
...
+      - run: "cargo check --all --examples"
+      - run: "cargo fmt -- --check"
+      - run: "cargo clippy -- -Dwarnings"
```
emgre, line 15: Using [`actions-rs/cargo`](https://github.com/actions-rs/cargo) makes the error pretty-printed in the GitHub Actions UI, instead of having to scroll through the logs. Also, the `--all` flag of `cargo check` is deprecated for `--workspace`.

```suggestion
      - uses: actions-rs/cargo@v1
        with:
          command: check
          args: --workspace
      - uses: actions-rs/cargo@v1
        with:
          command: fmt
          args: -- --check
      - uses: actions-rs/cargo@v1
        with:
          command: clippy
          args: -- -Dwarnings
```

emgre (commented):
Just my 2 cents, I've been through the setup of GitHub Actions of a Rust project recently so here are my comments.

Also, we might want to add a matrix to test it on beta and nightly at some point.

Nitori-:
I'll take it, thanks @emgre !

9names:
LGTM

tdittr (approved):

.github/workflows/check.yml
```diff
@@ -0,0 +1,27 @@
+on: [push, pull_request]
+name: CI Checks
+jobs:
+  check:
+    name: Check and Lint
+    runs-on: ubuntu-latest
```
kahara, line 6: To avoid possible issues when a new Ubuntu LTS emerges, could this be `runs-on: ubuntu-20.04`?

tdittr:
Now that I think about it again it might be nice to have `cargo test` since it will check the examples in the doc-strings (at-least for compilation errors).

tdittr:
I found how I ran the unit tests. It was simple enugh:
```bash
cargo test --target x86_64-unknown-linux-gnu
```

Although I'm not sure is the architecture running the CI fixed? Or could we sometimes run on an ARM machine?

9names:
@tdittr It will run on x86_64 unless we ask for something else.
We should add tests to CI in a separate PR, at least on my machine there's some breakage when running them currently and I don't want those changes to hold this PR up.
I'm going to merge this as-is as it's already in a good state, thanks @Nitori- for the contribution and everyone who suggested improvements.

Issue #19, opened 2021-02-07
Add a blinky example using the HAL


tdittr:
Can be implemented after the following parts have been implemented:
- [ ] GPIO (#4)
- [ ] Clocks (#6)
- [ ] Resets

Use cases:
- Starting point for fresh projects
- Getting a quick overview of the HAL

chinoto:
The checklist hasn't been updated since [main/rp2040-hal/examples/blinky.rs](https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/examples/blinky.rs) was added. Based on what I'm seeing in the example, GPIO seems done and I see a [resets.rs](https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/src/resets.rs) exists, is resetting handled on this line?
https://github.com/rp-rs/rp-hal/blob/a14cbb581907cd05208705b51b953ac773abb0d5/rp2040-hal/examples/blinky.rs#L23

I'm probably misunderstanding, but couldn't timing be handled like it is in the non-HAL example? And where does the 8e6 for AHB frequency come from? 
https://github.com/rp-rs/pico-blink-rs/blob/247df96b9089096140a9713ebb4d40bd369ef730/src/main.rs#L111
```rust
let mut delay = cortex_m::delay::Delay::new(cp.SYST, 8_000_000);
```

I'm looking forward to helping out so my friend and I can use the Picos we bought before checking Rust support 😅. I'm particularly interested in getting SPI, I2C, and possibly PWM/DMA implemented.

9names:
Apologies for not updating the checklist.
It was really used more as a guide of what to work on, and what we had started work on, but wasn't being updated on things being complete (since it was unclear how "complete" each section needed to be before being ticked off)
> is resetting handled on this line?

Hard to say with this old code, but the relevant code in the current codebase is
https://github.com/rp-rs/rp-hal/blob/711c0230b1923492bc9bd8c9614a41e92d2070cc/rp2040-hal/src/gpio/pin.rs#L786-L801
which is invoked here
https://github.com/rp-rs/rp-hal/blob/2d1086915c3c178985445edaf559dbbbe8f4b8d3/boards/pico/examples/pico_blinky.rs#L46-L51
Perhaps resetting isn't being handled at all in the base example?
We should probably check, and if it's not, fix that. It would be a good first issue for you to tackle if you are keen.

> where does the 8e6 for AHB 

[...]

Issue #4, opened 2021-01-30
Add control for GPIOs


tdittr:
The datasheet covers GPIOs in chapter 2.19. The chips has two IO banks, bank 1 is used for the QSPI flash which could be covered separately. This issue covers IO bank 0 which can be used for general purpose IO.

# Use cases
The GPIOs can be used for these functions (chapter 2.19.2):
- Control/read pin level (SIO, chapter 2.3.1)
- Connecting a pin to a peripheral (PIO, SPI, UART, I2C, PWM, ADC, USB Vbus control)
- Use as an interrupt (chapter 2.19.3)

The following settings can be changed for each pin (chapter 2.19.4):
- Output drive strength
- Output slew rate
- Schmitt trigger
- Pull-up/down
- Signal and direction inversion and override

The input buffer and output driver can be disabled (which is mandatory when used with the ADC (chapter 4.9.1)).

# Implementation ideas
## Type states
To make it harder to misuse we could add the following as type parameters (similar to [NRF HAL](https://github.com/nrf-rs/nrf-hal/blob/master/nrf-hal-common/src/gpio.rs) and the [embedded book](https://docs.rust-embedded.org/book/static-guarantees/state-machines.html)):
- Function selection (In-/Output (SIO), PIO[01], SPI[01]-(Rx|Tx|...), ...)
- Pull-Up/Down (only for Inputs)

## Traits
All pins have the same configurations listed above. So a trait that covers all these settings could make sense.

## Interrupts
ToDo...

anall:
I know the basic support just landed a few days ago, but I think refactoring this to be more like how the `atsamd` HAL does things (which seems to me to be one of the most mature HAL crates)

I hope I am not stepping on anyone's feet, but I've got a "minimum" viable example of my refactoring that I'll open as a draft PR.

9names:
GPIO and pin muxing complete as #34 merged

PR #24, opened 2021-04-17
Basic GPIO support

Nitori-:
I don't have a lot of experience with embedded Rust, so I'm definitely looking for feedback. This design is mostly inspired by other HALs like the STM32 ones.

The example code I've included seems to be enough to get the on-board Pico LED to turn on or off. I have no idea if it's legit to bring GPIO out of reset before doing clock stuff, but the clock stuff is WIP in that other PR and I don't feel anywhere near smart enough to help with that :sweat_smile: 

Things that are still TODO for this PR, that I plan to tackle if people are happy with what I've got so far:

- [x] Input
- [x] Docs

Things I'm considering out of scope for this PR:

- QSPI IO config
- Any GPIO functions other than SIO
- Bundling multiple GPIO pin updates into a single write (is this even a thing the embedded crates support? it would be handy, but seems like it could be difficult)
- Interrupts

Related to #4 and #19


rp2040-hal/src/gpio.rs
```diff
@@ -0,0 +1,132 @@
...
+                    pads: &mut pac::$PADSX,
+                    sio: &mut pac::SIO,
+                ) -> $PXi<Output> {
+                    pads.gpio[$i].modify(|_, w| w.ie().set_bit().od().clear_bit());
+                    unsafe {
+                        (*pac::$GPIOX::ptr()).gpio[$i]
+                            .gpio_ctrl
+                            .write_with_zero(|x| x.funcsel().bits(FUNCTION_SIO));
+                    }
+                    sio.gpio_oe.modify(|_, x| unsafe { x.bits(1 << $i) });
```
tdittr, line 73: If you move the ownership of the SIO into this driver you should use the atomic version of this register `GPIO_OE_SET` with `.write()` instead of `.modify()`.

Then you can also pass around pins without giving the SIO instance around with it.

tdittr (commented):
Hi,

thanks for picking this up. Your implementation looks good to me. The only major thing I would change is to move the ownership of the SIO and pads peripherals also into the driver. What do you think about that?

I might have overlooked something with the macro as I am not yet really used to their syntax. :smile:

rp2040-hal/Cargo.toml
```diff
@@ -13,4 +13,4 @@ license = "MIT OR Apache-2.0"
 cortex-m = "0.7.1"
 embedded-hal = "0.2.4"
 nb = "1.0.0"
-rp2040-pac = "0.1.1"
+rp2040-pac = { git = "https://github.com/rp-rs/rp2040-pac", branch="main" }
```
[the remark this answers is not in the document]
Nitori-: Yes we'll want a release of that crate again before this one.
Between the crates.io version of the PAC and the main branch, GPIOs seem to have been rearranged quite a bit.

rp2040-hal/src/gpio.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        $(
+            pub struct $PXi<MODE> {
+                _mode: PhantomData<MODE>,
+            }
+
+            impl<MODE> $PXi<MODE> {
+                pub fn into_output(
+                    self,
+                    pads: &mut pac::$PADSX,
```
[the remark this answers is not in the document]
Nitori-: I think we could do that. `GpioExt`'s `split` could require moves of `PADS` and `SIO`. I'll give that a shot and see if it causes grief.

I think stealing ownership of `PADS` is fine. I was hesitant about stealing `SIO` because it has a few other non-GPIO features (spinlocks and FIFOs between the cores). Things are even a bit more confusing because there's really multiple `SIO`s, one for each core. But maybe we just leave that alone for now, steal `SIO` and revisit it later once multi-core stuff is feasible (seems a bit far off for the HAL right now).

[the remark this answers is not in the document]
tdittr: Oh, I see. I did not check the PAC before about SIO. So in the long run we would need to split up the SIO registers into its parts. Not sure if that should be done on the PAC or the HAL level.

But as you said we can deal with that later.

rp2040-hal/src/gpio.rs
```diff
@@ -1,132 +1,337 @@
...
+// const FUNCTION_UART: u8 = 2;
+// const FUNCTION_I2C: u8 = 3;
+// const FUNCTION_PWM: u8 = 4;
 const FUNCTION_SIO: u8 = 5;
+// const FUNCTION_PIO0: u8 = 6;
+// const FUNCTION_PIO1: u8 = 7;
+// const FUNCTION_CLOCK: u8 = 8;
+// const FUNCTION_USB: u8 = 9;
 
 macro_rules! gpio {
```
tdittr, line 77: After a good nights sleep I think we can solve this without a macro by using const generics.

Instead of types for `Gpio0`, `Gpio1` etc. there would be just one type  `Gpio<const N: u8>`. This could make the docs easier and less redundant. And might also make it easier for some editors because the macro does not need to be expanded.

But we can also change it in a future iteration. I actually did not yet try const generics in the wild so I do not know what problems might show up.

rp2040-hal/src/gpio.rs
```diff
@@ -1,132 +1,337 @@
...
+                $(
+                    #[doc = "GPIO pin "]
+                    #[doc = $is]
+                    pub $pxi: $PXi<Unknown>,
+                )+
+            }
+
+            // Puts pad in same state as post-reset, according to datasheet
+            fn setup_pad_io(pads: &pac::$padsx::RegisterBlock, index: usize) {
+                pads.gpio[index].modify(|_, w| w.ie().set_bit()
```
tdittr, line 120: Sorry I did not notice this yesterday, but we can actually just use the [`.reset()`](https://docs.rs/rp2040-pac/0.1.1/rp2040_pac/generic/struct.Reg.html#method.reset) method to reset all the fields.

rp2040-hal/src/gpio.rs
```diff
@@ -1,132 +1,337 @@
...
+                                                .pde().set_bit()
+                                                .schmitt().set_bit()
+                                                .drive()._4m_a()
+                                                .slewfast().clear_bit());
+            }
 
-        $(
-            pub struct $PXi<MODE> {
-                _mode: PhantomData<MODE>,
+            fn set_gpio_function(gpios: &pac::$gpiox::RegisterBlock, index: usize, function: u8) {
```
tdittr, line 129: Maybe we could turn the function to an `enum` instead of a `u8`. Otherwise it gets a bit harder to explain why the `unsafe` block is actually safe.

rp2040-hal/src/gpio.rs
```diff
@@ -1,132 +1,337 @@
...
+                // them. Thus several of the methods below will reconstruct these objects
+                // as-needed
+
+                impl<MODE> $PXi<MODE> {
+                    #[doc = "Configure this pin as an output"]
+                    pub fn into_output(
+                        self,
+                    ) -> $PXi<Output> {
+                        unsafe {
+                            setup_pad_io(&*pac::$PADSX::ptr(), $i);
```
tdittr, line 155: I would move the dereferences into their own helper methods. That would move all the `unsafe` to one area. In turn all the functions using them would become easier to read.

rp2040-hal/Cargo.toml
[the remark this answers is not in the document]
Nic0w: Indeed, a new version would be needed for the PAC; commit https://github.com/rp-rs/rp2040-pac/commit/56198b7d7f795c1bf3b7c5040f3724bf6683186f was a breaking change to how GPIOs are handled but it never made it into a release. That commit also breaks the `pico-blink-rs` example based on the PAC, but I see a PR has been submitted to fix that.

rp2040-hal/src/gpio.rs
```diff
@@ -1,132 +1,337 @@
...
-                    unsafe {
-                        (*pac::SIO::ptr()).gpio_out_set.write(|x| x.bits(1 << $i));
+                impl StatefulOutputPin for $PXi<Output> {
+                    fn is_set_low(&self) -> Result<bool, Self::Error> {
+                        Ok(!self.is_set_high()?)
+                    }
+
+                    fn is_set_high(&self) -> Result<bool, Self::Error> {
+                        unsafe {
+                            Ok((*pac::SIO::ptr()).gpio_out_set.read().bits() & (1 << $i) != 0)
```
tdittr, line 208: Minor: The read value of the atomic registers is not really described in the datasheet so I would just use the "normal" register not the `set` one

```suggestion
                            Ok((*pac::SIO::ptr()).gpio_out.read().bits() & (1 << $i) != 0)
```

tdittr (changes requested):
Great to see so much progress. The implementation looks quite clean and complete to me. I only added some more minor points I noticed.

The only major point would be to use const generics instead of a macro. But I can't find a good example of another HAL already doing that. So if you want to keep the macro feel free to do so we can always come back to this in another iteration.

rp2040-hal/examples/blinky.rs
```diff
@@ -0,0 +1,29 @@
...
+#[entry]
+fn main() -> ! {
+    let mut pac = rp2040_pac::Peripherals::take().unwrap();
+
+    let pins = pac.IO_BANK0.split(pac.PADS_BANK0, pac.SIO, &mut pac.RESETS);
+    let mut led_pin = pins.gpio25.into_output();
+
+    loop {
+        led_pin.set_low().unwrap();
+        // TODO: I dare not use delays until we've got clocks running
```
9names, line 25: You could use cortex_m::asm::delay for this purpose
It does mean blink speed is arbitrary, but that's better than blinking so fast that no-one can perceive it

rp2040-hal/src/gpio.rs
Nitori-: So I'm relatively new to `unsafe`. To me the reason I didn't do that was because I basically didn't want the de-reference to escape the `unsafe` block, since it would be unsafe if we were to use it outside of this context.

rp2040-hal/src/gpio.rs
Nitori-: Yes I think for this first round it might be good to stick with what the other HALs are doing, though eliminating the macro would be ideal. The macro makes it way harder to understand what is going on.

rp2040-hal/src/gpio.rs
Nitori-: :woman_facepalming: this is why my input example stopped working. I thought I had just broken my board or something.

rp2040-hal/src/gpio.rs
tdittr: Oh I see, yes we need to make sure that the reference can't escape. We can use lifetimes to achieve that.
```rust
fn pads<'a>(&'a self) -> &'a pac::pads_bank0::RegisterBlock {
    unsafe {&*pac::$PADSX::ptr()}
}
```

Which ensures that the reference to the register block can at most live as long as the `Gpio$i`.

To make it simpler we can use [lifetime elision](https://doc.rust-lang.org/nomicon/lifetime-elision.html) which happens whenever you take one reference as parameter and return one reference from the function:
```rust
fn pads(&self) -> &pac::pads_bank0::RegisterBlock {
    unsafe {&*pac::$PADSX::ptr()}
}
```

rp2040-hal/src/gpio.rs
```diff
@@ -95,7 +95,13 @@ macro_rules! gpio {
...
 
                 fn split(self, pads: pac::$PADSX, sio: pac::SIO, resets: &mut pac::RESETS) -> Parts {
-                    resets.reset.modify(|_, w| w.$gpiox().clear_bit());
+                    resets.reset.modify(|_, w| w.$gpiox().clear_bit().$padsx().clear_bit());
+                    while resets.reset_done.read().$gpiox().bit_is_clear() {
+                        cortex_m::asm::delay(10);
+                    }
+                    while resets.reset_done.read().$padsx().bit_is_clear() {
+                        cortex_m::asm::delay(10);
+                    }
```
tdittr, line 104: At some point we could move these also to a HAL wrapper. Maybe add a ToDo note:
```suggestion
                    // TODO: Implement Resets in the HAL
                    while resets.reset_done.read().$gpiox().bit_is_clear() {
                        cortex_m::asm::delay(10);
                    }
                    while resets.reset_done.read().$padsx().bit_is_clear() {
                        cortex_m::asm::delay(10);
                    }
```

rp2040-hal/src/gpio.rs
Nitori-: Strangely the registers themselves don't seem to require `mut` to write to them. I wonder if that means we should follow the same pattern.

tdittr: Oh yes, that is weird. I would tend to take `&mut` but I'm not sure now. The `OutputPin` trait for example takes `&mut self`.

.cargo/config
```diff
@@ -1,3 +1,11 @@
 [build]
 # Instruction set of Cortex-M0+
 target = "thumbv6m-none-eabi"
+
+[target.'cfg(all(target_arch = "arm", target_os = "none"))']
```
Szpadel, line 5: Why we need those flags?

Nitori-: They seem to be the magic incantations needed to get the embedded Rust stuff to build. I copied them from the sample project.

Szpadel: `link-arg=--nmagic` is needed if flash or ram is not aligned to 0x10000

`link-arg=-Tlink.x"` is LLD linker (I think needed if you need to link C dependencies)

but i'm not sure why we need the last 2

tdittr (approved):
I think we should merge this as it is now and keep on iterating on the function over time :)

9names:
> I think we should merge this as it is now and keep on iterating on the function over time :)
Agreed. I'll fix up the simple merge conflict and bring it in.

9names:
CI failure here is because "cargo test" fails due to missing embedded-hal traits on the CI host.
Will merge this after https://github.com/rp-rs/rp-hal/pull/32 which disables running cargo test in CI

PR #25, opened 2021-04-24
Working HAL for the XOSC

Nic0w:
Hi,

This is a full, working implementation of a CrystalOscillator that allows use of the XOSC.
Typestate programming is used to transition between the different stages : Disabled > Initialized > Stable > Disabling, Dormant.

I've tested the code on my Pi Pico up to the first 3 stages (ie up to Stable) and confirmed with the Frequency Counter that the XOSC's frequency is indeed 12MHz at that point.

To be noted that the code is based on hardware_xosc/xosc.c from the official pico-sdk.

I also want to point out that I'm a novice in Rust and that this is my first ever contribution to a Rust crate, so don't hesitate to point out things that I could have done better.

Regards,


rp2040-hal/src/xosc.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// Creates a new CrystalOscillator from the underlying device.
+    pub fn new(dev: rp2040_pac::XOSC) -> Self {
+        CrystalOscillator {
+            device: dev,
+            state: Disabled
+        }
+    }
+
+    /// Initializes the XOSC : frequency range is set, startup delay is calculated and set.
+    pub fn initialize(self, frequency: Megahertz<u32>) -> Result<CrystalOscillator<Initialized>, Error> {
```
tdittr, line 94: Maybe this should take Hertz instead of Megahertz, since some oscillators have a rather weird rates, e.g. baud rate crystals.

rp2040-hal/src/xosc.rs
```diff
@@ -0,0 +1,192 @@
...
+
+        const ALLOWED_FREQUENCY_RANGE: RangeInclusive<Megahertz<u32>> = Megahertz(1)..=Megahertz(15);
+        const STABLE_DELAY: Milliseconds<u64> = Milliseconds(1_u64);
```
tdittr, line 97: This value could be different depending on the external device. But 1ms might be enough for most crystals. So maybe a const would be enough?

rp2040-hal/src/xosc.rs
```diff
@@ -0,0 +1,192 @@
...
+
+        self.device.ctrl.write(|w| {
+            w.freq_range()._1_15mhz();
+            w
+        });
+
+        let freq_hz: Hertz<u64> = frequency.into();
+        let delay_sec: Seconds<u64> = STABLE_DELAY.into();
+
+        //let startup_delay = ((freq_hz / 1000) + 128) / 256;
```
tdittr, line 112: A comment here is a good idea so the calculation is easier to understand, but I think the comment and the calculation do not match up. The divide ba 1000 seems to be missing below?

rp2040-hal/src/xosc.rs
```diff
@@ -0,0 +1,192 @@
...
+        let startup_delay = delay_sec.
+            checked_mul(freq_hz.integer()).and_then(|r|
+                 r.to_generic::<u64>(DIVIDER).ok()
+             ).
+            ok_or(Error::BadArgument)?;
```
tdittr, line 117: Might be overkill but maybe we should add a check if `startup_delay.integer()` fits into a `u16`.

Maybe something like
```
            ok_or(Error::BadArgument)?
            .integer()
            .try_into()
            .ok_ok(Error::BadArgument)?;
```

rp2040-hal/src/xosc.rs
```diff
@@ -0,0 +1,192 @@
...
+        });
+
+        Ok(self.transition(Initialized {
+            freq_hz: freq_hz
+        }))
+    }
+}
+
+/// A token that's given when the oscillator is stablilzed, and can be exchanged to proceed to the next stage.
+pub struct StableOscillatorToken {
```
tdittr, line 136: Oh this is nice, did you see this in my previous implementation or is this a common pattern? :)

rp2040-hal/src/xosc.rs
```diff
@@ -0,0 +1,192 @@
...
+        Ok(StableOscillatorToken {
+            _private: ()
+        })
+    }
+
+    /// Returns the stablilzed oscillator
+    pub fn get_stable(self, _token: StableOscillatorToken) -> CrystalOscillator<Stable> {
+        let freq_hz = self.state.freq_hz;
+        self.transition(Stable {
+            freq_hz:freq_hz
```
tdittr, line 158: This can be reduced to the following:
```suggestion
            freq_hz
```

(This can be done when the name of the field is the same as the name of the variable you want to assign to that field)

rp2040-hal/src/xosc.rs
```diff
@@ -0,0 +1,192 @@
...
+
+impl CrystalOscillator<Stable> {
+
+    /// Operating frequency of the XOSC in hertz
+    pub fn operating_frequency(&self) -> Hertz<u64> {
+        self.state.freq_hz
+    }
+
+    /// Disables the XOSC
+    pub fn disable(self) -> CrystalOscillator<Disabling> {
```
tdittr, line 171: I'm not sure what is the difference between `Disabling` and `Disabled`?

Maybe just go back to `Disabled` so the `initialize()` method can be used to re-enable the peripheral.

rp2040-hal/src/xosc.rs
```diff
@@ -0,0 +1,192 @@
...
+        self.device.ctrl.modify(|_r,w| {
+            w.enable().disable();
+            w
+        });
+
+        self.transition(Disabling)
+    }
+
+    /// Put the XOSC in dormant state
+    pub fn dormant(self) -> CrystalOscillator<Dormant> {
```
tdittr, line 181: Adding the `Dormant` mode is nice, although I would try to address the two warnings listed in the datasheet:
> WARNING: stop the PLLs before selecting dormant mode
WARNING: setup the irq before selecting dormant mode
-- Datasheet page 244

Maybe there could be a token for the IRQ? I'm not sure maybe we should not make this `pub` for now until we can make it safe?

Also there needs to be a way to get the mode from `Dormant` to `Disabled` to re-enable or free the peripheral.

tdittr (changes requested):
Hi,

thanks for taking up the work on the Oscillator. I started work on this previously, but I think you picked up from there anyway?

Overall this PR looks really good. I would just change a few minor details and maybe not yet expose the `Dormant` mode.

rp2040-hal/src/xosc.rs
```diff
@@ -0,0 +1,192 @@
...
+            return Err(Error::FrequencyOutOfRange)
+        }
+
+        self.device.ctrl.write(|w| {
+            w.freq_range()._1_15mhz();
+            w
+        });
+
+        let freq_hz: Hertz<u64> = frequency.into();
+        let delay_sec: Seconds<u64> = STABLE_DELAY.into();
```
tdittr, line 110: Looking again, this might truncate the sub-seconds so this could come out to being 0s.

rp2040-hal/src/xosc.rs
Nic0w: Yes but no ! The calculation is actually a frequency in hertz being multiplied by a time in seconds. Except the time in second is 1ms, ie 10^-3 seconds, so it can be "simplified" to a divide by 1000 (and that's how it's written in the C SDK). Originally I was doing just the same but I felt it wasn't accurate regarding the formula so I brought back the multiply.
The comment is mathematically accurate but it doesn't reflect the calculation anymore so I'm going to change it :)

rp2040-hal/src/xosc.rs
Nic0w: I kept pondering if using embedded_time types was actually a good idea because the conversion traits don't work for all core integer types and that brings "ugly" casts everywhere. Your suggestion might actually be a good way to make them look better.

Nic0w: Related to that code is the use of u64. I was wondering, for the sake of low-level/embedded, should we constraint all integer types to u32/usize ?

rp2040-hal/src/xosc.rs
Nic0w: At first I did without because I written a `_blocking` method with a `while { nop }` inside. Then, as I wanted to implement what I would call the "WouldBlock" pattern, I realized your use of these tokens was actually a very elegant solution !

rp2040-hal/src/xosc.rs
Nic0w: Good remarks. I implemented the Dormant state to be iso-functionnal with the C SDK re. XOSC but yeah you might be right about not exposing it right now (as I don't have an idea of how to deal with IRQs just yet)

rp2040-hal/src/xosc.rs
tdittr: > Related to that code is the use of u64. I was wondering, for the sake of low-level/embedded, should we constraint all integer types to u32/usize ?

I think for code that mostly runs once during an initialization it shouldn't be a big deal. In this case u32 might suffice though it would allow us to represetz up to 4GHz in Hz.

Nic0w:
> Hi,
> 
> thanks for taking up the work on the Oscillator. I started work on this previously, but I think you picked up from there anyway?
> 
> Overall this PR looks really good. I would just change a few minor details and maybe not yet expose the `Dormant` mode.

Hi,

Thanks for your feedback and code review =) I read extensively your PR and I saw you tried to bind nicely together Clocks, PLLs, Oscillators, ... with extensive typestates and such. I then had a brain meltdown trying to think about about how to pick up your work; not criticizing your code (from which I learned a lot) but the complexity of the clocks/PLL/oscillators subsystems and the complexity of "bridging" them together, with all the different abstractions that would be needed : macros, typestate, ...

So I decided on starting from scratch, taking inspiration from your PR as you saw (`StableOscillatorToken` for example) with the goal of trying to implement one subsystem at a time, starting at the bottom of the foodchain : XOSC.

Not sure if it's the right approach but I figured that it might be easier that way =)

rp2040-hal/src/xosc.rs
tdittr: Thanks :) I was actually a bit confused because most of the code looked new but the token looked familiar.

In the end I added a blocking helper method (`new_stable_blocking()`) that runs through all the steps for cases where you just want a running Oscilator without caring about all the steps. Maybe this could be a good adition to this driver as well?

tdittr:
Oh I can feel you, I had the same brain-meltdown. I think I bit off a bit to much of the problem. Writing it in smaller steps definitely is a good idea. :+1:

Maybe at some point in the future when we have all the building blocks we can actually tie them together.

rp2040-hal/src/xosc.rs
tdittr: > I kept pondering if using embedded_time types was actually a good idea [...]

Yes I had the same struggle, it seems like a nice way to add all the different units, but I ended up in generics-hell because I wanted to take *any* rate. But maybe just sticking to the smallest common unit makes it a bit easier to handle.

rp2040-hal/src/xosc.rs
tdittr: You could also make it `unsafe` noting the Warnings from the datasheet in the Safety section of the doc comment.

rp2040-hal/src/xosc.rs
Nic0w: I don't have any knowledge on physics/electronics of crystal oscillators to know if 1ms delay is sufficient for oscillators across the full range of accepted frequencies (1-15Mhz) so I left that as is.

rp2040-hal/src/xosc.rs
Nic0w: You were right ! I struggled for an hour to try and write a unit test within the crate but apparently it's not possible when the crate is no_std. None of the tricks I found online worked so here is the code of the test, that I ran on a standalone project, if you happen to know how to run it from within this crate : 

```rust
#[cfg(test)]
mod tests {
use std::convert::TryInto;
use embedded_time::{
        fraction::Fraction,
        fixed_point::FixedPoint,
        rate::{
            Hertz,
            Megahertz,
            Rate
        },
        duration::{
            Seconds,
            Milliseconds,
            Duration,
            Generic,
        }
    };
    #[test]
    fn calculate_xosc_delay() {
        let frac: Fraction = Fraction::new(256, 1);
        let milliseconds = Milliseconds(1_u32);
        let seconds: Seconds = milliseconds.into();
        let frequency: Hertz = Hertz(12_000_000_u32);
        
        let delay_to_hz: Hertz = milliseconds.to_rate().unwrap();
        assert_eq!(*delay_to_hz.integer(), 1000);

        let oper = frequency.checked_div(delay_to_hz.integer()).and_then(|r| {
            r.to_generic::<u32>(frac).ok()
        });
        assert!(oper.is_some());

        let res = oper.unwrap();

        let to_u16: Result<u16, _> = (*res.integer()).try_into();
        assert!(to_u16.is_ok());

        //Value from datasheet (Chapter 2, Section 16, §3)
        assert_eq!(to_u16.unwrap()+1, 47);
    }
}
```

tdittr: Oh yes I remember I had the same problem. The tests need to be compiled for the host instead of for the pico. But I don't remember how I did this last time.

I would try to move the calculation to its own function and then add this unit test to the HAL.

rp2040-hal/src/xosc.rs
Nic0w: And done :)

Nic0w:
This is ready for review again :)

tdittr (approved):
Looks good to me! :+1:

9names (approved):
Looks good to me too.
After merging, it would be nice to get an example showing how to use this correctly.

PR #26, opened 2021-04-25
Ignores Cargo.lock and target/*.

Nic0w:


tdittr (approved):

tdittr:
Simple enough. Though it might collide with the CI PR.

Nic0w:
Possibly, and I just realized that I messed up by including that same change in the PLL PR :/

Nic0w:
https://github.com/rp-rs/rp-hal/pull/23 does it better.

PR #27, opened 2021-04-25
Working HAL for PLLs

Nic0w:
This is a full, working implementation of a PLL HAL that allows use of both PLLs.

Typestate programming is used to transition between the different stages : Disabled > Locking > Locked.

I've tested the code on my Pi Pico using parameters from the nominal configuration outlined in the Datasheet and in the C SDK and I was able to confirme with the Frequency Counter that the `PLL_SYS` frequency is indeed 125MHz with this configuration.

The code is based on `hardware_pll/pll.c` from the official pico-sdk, with inspiration from @tdittr 's PR.

To be noted that there is something strange going on with the `clear_bit()` method for the PWR register, as I explain here : https://github.com/rp-rs/rp2040-pac/issues/16


rp2040-hal/src/pll.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    /// Bad argument : overflows, bad conversion, ...
+    BadArgument
+}
+
+impl<D: PhaseLockedLoopDevice> PhaseLockedLoop<Disabled, D> {
+
+
+    /// Instantiates and configures a new Phase-Locked-Loop device.
+    pub fn new(dev: D, refdiv: u8, vco_freq: Generic<u32>, post_div1: u8, post_div2: u8) -> Result<PhaseLockedLoop<Disabled, D>, Error> {
```
tdittr, line 104: I would merge the calculations in `initialize` and `new` into `initialize`. That way you could reconfigure the PLL later without using `free`. Also then you don't have to pass state between `new` and `initialize`.

Also having all the parameters in a struct would allow us to add a `calculate_params` function that calculates a fitting set of parameters for a given XOsc and desired output frequency. Splitting out the calculations from the configuration should also make it easier to test the module.

tdittr (changes requested):
Looks solid to me. The only thing I would change is to split the parameter calculation from the actual configuration. Makes testing easier and decouples the calculations from the hardware specific parts.

rp2040-hal/src/pll.rs
Nic0w: Done, tell we what you think :)

rp2040-hal/src/pll.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// Bad argument : overflows, bad conversion, ...
+    BadArgument
+}
+
+
+/// Parameters for a PLL.
+pub struct PLLConfig<R: Rate> {
+
+    /// Voltage Controlled Oscillator frequency.
+    pub vco_freq: R,
```
tdittr, line 101: I think I would put in the feedback divider value instead of the `vco_frequency`. Although then we need to carry around the XOsc frequency to calculate the effective output.

rp2040-hal/src/pll.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// Post Divider 1
+    pub post_div1: u8,
+
+    /// Post Divider 2
+    pub post_div2: u8
+}
+
+/// Common configs for the two PLLs. Both assume the XOSC is cadenced at 12MHz !
+/// See Chapter 2, Section 18, §2
+pub mod common_configs {
```
tdittr, line 115: I really like this pattern of common configs! :+1:

rp2040-hal/src/pll.rs
```diff
@@ -0,0 +1,254 @@
...
+
+        let vco_freq: Hertz = config.vco_freq.try_into().map_err(|_| Error::BadArgument)?;
+
+        if !VCO_FREQ_RANGE.contains(&vco_freq) {
+            return Err(Error::VCOFreqOutOfRange)
+        }
+
+        if !POSTDIV_RANGE.contains(&config.post_div2) || !POSTDIV_RANGE.contains(&config.post_div2) {
+            return Err(Error::PostDivOutOfRage)
+        }
```
tdittr, line 162: If you move these checks to a `new()` method on the `PllConfig` initialize could always succeed and there could be no "invalid" `PLLConfigs`.

rp2040-hal/src/pll.rs
```diff
@@ -0,0 +1,254 @@
...
+        const POSTDIV_RANGE: Range<u8> = 1..7;
+        const FBDIV_RANGE: Range<u16> = 16..320;
+
+        let vco_freq: Hertz = config.vco_freq.try_into().map_err(|_| Error::BadArgument)?;
+
+        if !VCO_FREQ_RANGE.contains(&vco_freq) {
+            return Err(Error::VCOFreqOutOfRange)
+        }
+
+        if !POSTDIV_RANGE.contains(&config.post_div2) || !POSTDIV_RANGE.contains(&config.post_div2) {
```
tdittr, line 160: Typo
```suggestion
        if !POSTDIV_RANGE.contains(&config.post_div1) || !POSTDIV_RANGE.contains(&config.post_div2) {
```

rp2040-hal/src/pll.rs
```diff
@@ -0,0 +1,254 @@
...
+        });
+
+        let fbdiv = vco_freq.checked_div(ref_freq_hz.integer()).
+            ok_or(Error::BadArgument)?;
+
+        let fbdiv: u16 = (*fbdiv.integer()).try_into().map_err(|_| Error::BadArgument)?;
+
+        if !FBDIV_RANGE.contains(&fbdiv) {
+            return Err(Error::FBDIVOutOfRange)
+        }
```
tdittr, line 191: Same with these checks.

tdittr (commented):
Yes that looks like what I imagined :) If you want to you could make `PllConfig` to only be able to hold valid configurations.

But we can do that in a next iteration (with my large PR it stalled at some point because I tried to do to many things at once).

Nic0w:
I feel like this is good for review again :)

tdittr (approved):
I think we can merge this and keep on iterating together with the XOSC and Clocks PRs with the UART as an example usage.

9names (commented):
LGTM.
Let me know once you've run verified that cargo clippy doesn't find any issues, and run it through cargo fmt, and I'll look to get this merged

Nic0w:
Uhhg sorry for the delay, my `cargo clippy` is not reporting those errors so I discovered those just now.

9names:
No stress about the delay. That's why it's in CI though - even if it's not flagging on your end, the server will catch it.
Thanks for the contribution!

PR #28, opened 2021-04-25
Working implementation of an UART HAL.

Nic0w:
This is a full, working implementation of an UART HAL that allows use of both UARTs.

Typestate programming is used to transition between the different stages : Disabled > Enabled.

I've tested the code on my Pi Pico with the 9600 baud common configuration exposed by the module : it works.

To be noted that in order to have UART functionning, the minimum configuration is :
1. the UARTs devices are out of reset. 
2. Peripheral Clock (clock_peri_*) is enabled at whatever stable frequency (source could be XOSC, not necessarily PLLs)
3.  two available GPIOs properly switched to the UART function and configured as Output lows.


rp2040-hal/src/uart.rs
```diff
@@ -1,3 +1,356 @@
...
+            device, config, effective_baudrate, _state: Enabled
+        }
+    }
+}
+
+impl<D: UARTDevice> UARTPeripheral<Enabled, D> {
+
+    /// Disable this UART Peripheral, falling back to the Disabled state.
+    pub fn disable(self) -> UARTPeripheral<Disabled, D> {
+        self.transition(Disabled)
```
tdittr, line 181: I think this should reset the enable signals.

rp2040-hal/src/uart.rs
```diff
@@ -1,3 +1,356 @@
...
+    fn uart_is_readable(&self) -> bool {
+        self.device.uartfr.read().rxfe().bit_is_clear()
+    }
+
+    /// Writes bytes to the UART.
+    /// This function writes as long as it can. As soon that the FIFO is full, if :
+    /// - 0 bytes were written, a WouldBlock Error is returned
+    /// - some bytes were written, it is deemed to be a success
+    /// Upon success, the number of written bytes is returned.
+    pub fn write(&self, data: &[u8]) -> nb::Result<usize, Infallible> {
```
tdittr, line 197: Would it make sense to return a slice of the not-send data instead of a usize? Then you could hand that slice back in and try again until you have an empty slice.

rp2040-hal/src/uart.rs
```diff
@@ -1,3 +1,356 @@
...
+        }
+        Ok(bytes_written)
+    }
+
+    /// Reads bytes from the UART.
+    /// This function reads as long as it can. As soon that the FIFO is empty, if :
+    /// - 0 bytes were read, a WouldBlock Error is returned
+    /// - some bytes were read, it is deemed to be a success
+    /// Upon success, the number of read bytes is returned.
+    pub fn read(&self, buffer: &mut [u8]) -> nb::Result<usize, Infallible> {
```
tdittr, line 227: Similarly here I would return a slice which only covers the filled bytes.

rp2040-hal/src/uart.rs
```diff
@@ -1,3 +1,356 @@
...
+                    return Ok(bytes_written)
+                }
+            }
+
+            bytes_written += 1;
+
+            self.device.uartdr.write(|w| unsafe {
+                w.data().bits(*c);
+                w
+            })
```
tdittr, line 217: Very-Minor: I would put the increment after the write happened.
```suggestion
            self.device.uartdr.write(|w| unsafe {
                w.data().bits(*c);
                w
            });
            
            bytes_written += 1;
```

rp2040-hal/src/uart.rs
```diff
@@ -1,3 +1,356 @@
...
+    }
+
+}
+
+/// Baudrate configuration. Code loosely inspired from the C SDK.
+fn configure_baudrate(device: &mut dyn UARTDevice, wanted_baudrate: &Baud, frequency: &Hertz) -> Baud {
+
+    let frequency = *frequency.integer();
+
+    let baudrate_div = 8 * frequency / *wanted_baudrate.integer();
```
tdittr, line 295: I would use `checked_` arithmatic here to catch any weird parameters.

rp2040-hal/src/uart.rs
```diff
@@ -1,3 +1,356 @@
...
+
+    let (baud_ibrd, baud_fbrd) = match (baudrate_div >> 7, ((baudrate_div & 0x7F) + 1) / 2) {
+
+        (0, _) => (1, 0),
+
+        (ibrd, _) if ibrd >= 65535 => (65535, 0),
+
+        (ibrd, fbrd) => (ibrd, fbrd)
+    };
```
tdittr, line 304: Similarly to the PLL I would split this calculation out into a function so it can be tested.

rp2040-hal/src/uart.rs
```diff
@@ -1,3 +1,356 @@
...
+    device.uartfbrd.write(|w| unsafe {
+        w.baud_divfrac().bits(baud_fbrd as u8);
+        w
+    });
+
+    // PL011 needs a (dummy) line control register write to latch in the
+    // divisors. We don't want to actually change LCR contents here.
+    device.uartlcr_h.write(|w| {
+        w
+    });
```
tdittr, line 320: I think write will overrite all the values. Maybe you meant `modify()`?

tdittr (changes requested):
Again a very nice driver. The only major thing I would change is to take ownership of the Tx and Rx Pins, but I think that can be changed in a later iteration when the GPIO PR is merged.

tdittr:
> To be noted that in order to have UART functionning, the minimum configuration is :
> 
> 1. the UARTs devices are out of reset.
> 2. Peripheral Clock (clock_peri_*) is enabled at whatever stable frequency (source could be XOSC, not necessarily PLLs)
> 3. two available GPIOs properly switched to the UART function and configured as Output lows.

1. You could address by taking a `&mut` to the reset block in `new()` as done in the GPIO PR.
2. I think we can adapt when the clocks are done. But it should be noted in the doc comments.
3. Could be addressed by taking ownership of the Tx and Rx Pins when the GPIO PR is merged.

rp2040-hal/src/uart.rs
Nic0w: I asked myself the same question when writing this. I'm not sure what happens when `write()` is called without actually making any changes to the register. I also tested with `modify()` and with nothing at all (commented out the code) and it does not seem to have any impact. This code is there only because I tried to mimick what's being done in the C SDK.
What they actually do is :
```c
    // PL011 needs a (dummy) line control register write to latch in the
    // divisors. We don't want to actually change LCR contents here.
    hw_set_bits(&uart_get_hw(uart)->lcr_h, 0);
```
So they write 0 ? I'm not sure what this does and how it can be translated using our PAC.

tdittr: As I understand it `.write()` only gives you the reset value of the register so you have a struct to call the methods for each field on. I'm not sure what `hw_set_bits` does but I would assume that it just is a `register |= 0` which writes to the register but with the original value. In Rust that should be `.modify(|_, w| w)`.

rp2040-hal/src/uart.rs
Nic0w: Indeed !

rp2040-hal/src/uart.rs
Nic0w: That might be a smart thing to do yeah... I'll see how that works out with the blocking function :)

Nic0w:
I can assure you your comments are very helpful to me ahah ! I wrote this code for my own use and so I didn't pay attention to some details that someone else would pay attention to. So yeah, thanks again for the reviews :)

rp2040-hal/src/uart.rs
Nic0w: I've been trying to do it and after 5 minutes or so fighting against the borrow checker because of the mut refs; I'm not sure it is doable. It works well for the write version tho !

Nic0w: Nevermind, I wrote it just a bit differently !

tdittr:
Oh one thing I forgot. You could also implement the [serial traits](https://docs.rs/embedded-hal/0.2.4/embedded_hal/serial/index.html) from embedded HAL. Then any driver written for that trait can be directly used on the Pico :smiley:

tdittr (changes requested):
Thanks fot implementing the embedded-hal traits. I would just restructure a few things:

rp2040-hal/src/serial.rs
```diff
@@ -0,0 +1,50 @@
...
+    Read,
+    Write
+};
+
+use nb::Error::WouldBlock;
+
+impl<D: UARTDevice> Read<u8> for UARTPeripheral<Enabled, D> {
+    type Error = Infallible;
+
+    fn read(&mut self) -> nb::Result<u8, Self::Error> {
```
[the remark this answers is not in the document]
Szpadel: wouldn't this mean that you cannot read anything in that case instead of getting at least something?

Or error would contain in this case what it was able to read?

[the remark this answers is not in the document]
tdittr: Yes that is true, we could still return the data in the `Err()` variant.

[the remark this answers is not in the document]
Nic0w: I'm not sure what you mean here @tdittr by the FIFO being overflowed and moreover how that state could have you wait too long between two calls to `read()`.
Still I feel like I have to disagree here. 
First, the `embedded-hal` traits' implementations "_Must erase device specific details._" (quoting the official documentation). So from that perspective, returning an error specific to the state of the FIFO doesn't stand. By the way, the FIFO might not be necessarily enabled : the default setting for the UART is to receive/send only one byte at a time.
Second, my understanding is that if the Receive FIFO is overflowed, that still means there's something to read. If there's something to read, then I feel like it would be an anti-pattern to return an error since `read()`'s goal is to read.

I feel like if there's an error to be had, that would be at the other side of the wire. If our Receive FIFO is full, then the sender's Transmit FIFO is going to overflow, and it's going to trigger a WouldBlock or equivalent on their side (and that's indeed what we do already).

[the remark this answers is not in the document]
tdittr: > I feel like if there's an error to be had, that would be at the other side of the wire. If our Receive FIFO is full, then the sender's Transmit FIFO is going to overflow, and it's going to trigger a WouldBlock or equivalent on their side (and that's indeed what we do already).

That would only hold in cases where flow control is being used. Without it either side can send data whenever it likes.

> I'm not sure what you mean here @tdittr by the FIFO being overflowed and moreover how that state could have you wait too long between two calls to read().

So if the other side sends data faster then we call `read()` at some point there will be no more space in the FIFO und thus the overrun error flag is set to high.

> Still I feel like I have to disagree here.
> First, the embedded-hal traits' implementations "Must erase device specific details." (quoting the official documentation). So from that perspective, returning an error specific to the state of the FIFO doesn't stand. 

Yes it has to erase device specific details so that a driver does not need to worry about *what* can go wrong, but it needs to be notified *that* something has gone wrong. That could also include other errors like parity errors. For example see the [`stm32f1xx_hal::serial::Error`](https://docs.rs/stm32f1xx-hal/0.7.0/stm32f1xx_hal/serial/enum.Error.html).

> By the way, the FIFO might not be necessarily enabled : the default setting for the UART is to receive/send only one byte at a time.

That does not actually change very much as then there would just be a FIFO of size 1. It will only overrun faster.

> Second, my understanding is that if the Receive FIFO is overflowed, that still means there's something to read. If there's something to read, then I feel like it would be an anti-pattern to return an error since read()'s goal is to read.

Yes that is a tough thing to decide, when to report the overrun. I would argue that it would be simpler to be reported as soon as the flag is set. Otherwise we need to keep the state of when the bit was set and count that down after any `read()` call.

Let me give an example. Assume the FIFO is turned off (size 1) and we have a simple protocol sending a coordinate like this:
```
X=-42
```

Now lets say the driver ready `X` and then is interrupted by a higher priority task. The other side sends `=` which is put into the FIFO. The driver has not yet resumed and the other side continues with `-` now there is no room in the FIFO so the `-` is dropped.

When the driver continues reading now it will read `=`. And continue with `42`. But now the `-` is lost which might be an important part of the message which now reads:
```
X=42
```

I hope that makes it more clear.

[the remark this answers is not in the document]
tdittr: Checked the datasheet again and it looks like without the FIFO activated it actually behaves like a shit-register. So the last character will be overritten by the new one.

So for my extreme example the driver should stop after reading `=` then `-` is received but overritten by `4`. :)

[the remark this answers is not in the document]
Nic0w: Thanks for the detailed explanation, and I understand how an error would be required in case of overflow. I'll try to write something and you guys can tell me what you think !

rp2040-hal/src/uart.rs
```diff
@@ -345,40 +396,78 @@ fn configure_baudrate(device: &mut dyn UARTDevice, wanted_baudrate: &Baud, frequ
...
+impl<D: UARTDevice> Read<u8> for UARTPeripheral<Enabled, D> {
+    type Error = ReadErrorType;
+
+    fn read(&mut self) -> nb::Result<u8, Self::Error> {
+        let byte: &mut [u8] = &mut [0; 1];
+
+        match self.read_raw(byte) {
+            Ok(_) => Ok(byte[0]),
+            Err(e) => match e {
+                Other(inner) => return Err(Other(inner.err_type)),
```
tdittr, line 452: I think the `return`s could be removed since the match would then just resolve to one of the values.

Maybe the two matches could also be merged into one but that is more a matter of taste :smile:

tdittr (approved):
Looks good to me :) thanks again for picking this up. If you would like you could also add an example in another PR.

Nic0w:
Re. examples, I was wondering what was the best way to proceed ? Like, where to put them ?

rp2040-hal/src/uart.rs
```diff
@@ -1,3 +1,473 @@
...
+/// UART is disabled.
+pub struct Disabled;
+
+impl State for Enabled {}
+impl State for Disabled {}
+
+/// Data bits
+pub enum DataBits {
+    /// 5 bits
+    Five,
```
9names, line 71: I think you're going a bit overboard with doc strings here.
Reader can probably infer that DataBits are Data Bits, and that Five represents 5 bits

9names (changes requested):
A couple of comments need fixing, but otherwise this looks good to me.

rp2040-hal/src/uart.rs
Nic0w: Totally agreeing with you but I think it wouldn't compile without the comments. Will check again !

rp2040-hal/src/uart.rs
```diff
@@ -1,3 +1,473 @@
...
+        Ok(())
+    }
+}
+
+/// Baudrate dividers calculation. Code inspired from the C SDK.
+fn calculate_baudrate_dividers(
+    wanted_baudrate: &Baud,
+    frequency: &Hertz,
+) -> Result<(u16, u16), Error> {
+    // baudrate_div = frequency * 8 / wanted_baudrate
```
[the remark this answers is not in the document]
Nic0w: It's not dead code, but more of a comment to show the calculation going on just under in a more readable way.
I can just as well remove the comment or explain the calculation using words. Which one would you prefer ?

rp2040-hal/src/uart.rs
Nic0w:
```
error: missing documentation for a variant
  --> rp2040-hal/src/uart.rs:70:5
   |
70 |     Five,
   |     ^^^^
   |
note: the lint level is defined here
  --> rp2040-hal/src/lib.rs:6:9
   |
6  | #![deny(missing_docs)]
   |         ^^^^^^^^^^^^

error: aborting due to previous error

error: could not compile `rp2040-hal`

```

rp2040-hal/src/uart.rs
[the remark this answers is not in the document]
9names: If the words explain why (reason why it's multiplied by 8 for example), I would prefer that.
If it's just restating the formula I would prefer you omit it.

rp2040-hal/src/uart.rs
9names: This was bad advice, my apologies.
I don't like it but it's better than writing manual exceptions to doc requirements.

9names:
I've pulled in main so you're getting all the CI checks running.
Ready to merge this as soon as you resolve the comment issue and make cargo clippy and cargo fmt happy

9names (approved):

PR #29, opened 2021-04-27
Add `cargo test` to CI run and fix an error found by it

tdittr:
Besides the normal usage of unit-tests this also ensures that
example in doc-comments do at least compile.

This expands #23


9names (approved):
LGTM

PR #30, opened 2021-04-29
Remove unnecessary unsafe block + cargo fmt XOSC

9names:
XOSC branch had a warning in build which caused CI to fail.
Also wasn't formatted with cargo fmt, so I've done that as well.


tdittr (approved):
LGTM

Szpadel (approved):

PR #31, opened 2021-05-01
Clocks

Nic0w:
Relatively thin abstraction layer that I wrote for clocks ! 
I took inspiration from existing crates and @tdittr 's work regarding macros and clock sharing.

This code is the bare minimum to have clocks working :
- set clock source where possible
- set clock auxiliary source
- set dividers where possible 
- enable/disable clock (where possible)

Example usage :
```rust

pub fn setup_clocks(clocks_dev: CLOCKS) -> Result<ClocksManager, u8> {

    let clocks = ClocksManager::new(clocks_dev);

    let mut ref_clock: ReferenceClock = clocks.ref_clock();
    ref_clock.set_xosc_src();
    ref_clock.set_int_div(1); //no division

    let mut sys_clock = clocks.sys_clock();
    sys_clock.reset_source_await();
    sys_clock.set_pll_sys_auxsrc();
    sys_clock.set_self_aux_src();
    sys_clock.await_select(0x2);

    // if we'd need to divide clocksys frequency
    //sys_clock.set_int_div(3);

    let mut peri_clock = clocks.peripheral_clock();
    peri_clock.disable();
    //change auxiliary src
    peri_clock.set_clksys_auxsrc();
    peri_clock.enable();

    Ok(clocks)
}
```

Things that I considered to be out of scope :
- Frequency Counter
- Resus

### DIFF
--- a/rp2040-hal/src/clocks/macros.rs
+++ b/rp2040-hal/src/clocks/macros.rs
@@ -1,0 +1,226 @@
+#[macro_use]
+
+macro_rules! int_division {
+     ($name:ident, $div:ident, $u:ty) => {
+
+         impl IntegerDivision for $name {
+             fn set_int_div(&mut self, div: usize) {
+                 unsafe { self.shared_dev.get() }.$div.write(|w| unsafe {
+                     w.int().bits(div as $u);
+                     w
+                 });
+             }
+         }
+    };
+}
+
+
+macro_rules! frac_division {
+     ($name:ident, $div:ident, $u:ty) => {
+
+         impl FractionDivision for $name {
+             fn set_frac_div(&mut self, div: usize) {
+                 unsafe { self.shared_dev.get() }.$div.write(|w| unsafe {
+                     w.frac().bits(div as $u);
+                     w
+                 });
+             }
+         }
+    };
+}
+
+
+macro_rules! clock_generator {
+    ($name:ident, $ctrl:ident) => {
+        impl ClockGenerator for $name {
+            fn enable(&mut self) {
+                unsafe { self.shared_dev.get() }.$ctrl.write(|w| {
+                    w.enable().set_bit();
+                    w
+                });
+            }
+
+            fn disable(&mut self) {
+                unsafe { self.shared_dev.get() }.$ctrl.write(|w| {
+                    w.enable().clear_bit();
+                    w
+                });
+            }
+
+            fn kill(&mut self) {
+                unsafe { self.shared_dev.get() }.$ctrl.write(|w| {
+                    w.kill().set_bit();
+                    w
+                });
+            }
+        }
+    };
+}
+
+
+macro_rules! xosc_source {
+    ($name:ident, $ctrl:ident) => {
+
+        impl XOSCClockSource for $name {
+            fn set_xosc_src(&mut self) {
+                unsafe { self.shared_dev.get() }.$ctrl.write(|w| {
+                    w.src().xosc_clksrc();
+                    w
+                });
+            }
+        }
+    };
+}
+
+
+macro_rules! rosc_source {
+    ($name:ident, $ctrl:ident) => {
+        impl ROSCClockSource for $name {
+            fn set_rosc_src(&mut self) {
+                unsafe { self.shared_dev.get() }.$ctrl.write(|w| {
+                    w.src().rosc_clksrc_ph();
+                    w
+                });
+            }
+        }
+    };
+}
+
+
+macro_rules! selfaux_source {
+    ($name:ident, $ctrl:ident, $self:ident) => {
+        impl SelfAuxClockSource for $name {
+            fn set_self_aux_src(&mut self) {
+                unsafe { self.shared_dev.get() }.$ctrl.write(|w| {
+                    w.src().$self();
+                    w
+                });
+            }
+        }
+    };
+}
+
+
+macro_rules! clockref_source {
+    ($name:ident, $ctrl:ident) => {
+        impl ClockREFClockSource for $name {
+            fn set_clkref_src(&mut self) {
+                unsafe { self.shared_dev.get() }.$ctrl.write(|w| {
+                    w.src().clk_ref();
+                    w
+                });
+            }
+        }
+    };
+}
+
+
+macro_rules! clocksys_auxsource {
+    ($name:ident, $ctrl:ident) => {
+        impl ClockSYSClockAuxSource for $name {
+            fn set_clksys_auxsrc(&mut self) {
+                unsafe { self.shared_dev.get() }.$ctrl.write(|w| {
+                    w.auxsrc().clk_sys();
+                    w
+                });
+            }
+        }
+    };
+}
+
+macro_rules! xosc_auxsource {
+    ($name:ident, $ctrl:ident) => {
+        impl XOSCClockAuxSource for $name {
+            fn set_xosc_auxsrc(&mut self) {
+                unsafe { self.shared_dev.get() }.$ctrl.write(|w| {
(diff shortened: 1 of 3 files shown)


9names (commented):
Looks good. Not sure how I feel about setting a clock as it's own source though.
I guess we'll find out how good an idea that is...

rp2040-hal/src/clocks/macros.rs
```diff
@@ -0,0 +1,210 @@
...
+                });
+            }
+        }
+    };
+}
+
+macro_rules! frac_division {
+    ($name:ident, $div:ident, $u:ty) => {
+        impl FractionDivision for $name {
+            fn set_frac_div(&mut self, div: usize) {
```
anall, line 19: Using `write` instead of `modify` will clobber the rest of this register. This is an issue throughout but is important here as we can't set both the integer and fractional part of the divider.

rp2040-hal/src/clocks/macros.rs
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+    };
+}
+
+macro_rules! selfaux_source {
+    ($name:ident, $ctrl:ident, $self:ident) => {
+        impl SelfAuxClockSource for $name {
+            fn set_self_aux_src(&mut self) {
+                unsafe { self.shared_dev.get() }.$ctrl.write(|w| {
```
anall, line 86: I mentioned above re. using `write` instead of `modify`, but here it completely prevents actually setting a different aux clock source besides what happens to be in `0x00`, making the `set_…_auxsrc` functions useless.

rp2040-hal/src/clocks/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+pub trait XOSCClockSource {
+    /// Set XOSC as a source.
+    fn set_xosc_src(&mut self);
+}
+/// For clocks that can have ROSC as source.
+pub trait ROSCClockSource {
+    /// set ROSC as a source.
+    fn set_rosc_src(&mut self);
+}
+/// For clocks that can have ... itself (?) as a source (is that the "glitchless mux" ?)
```
anall, line 89: This is for when clocks have a number of sources where you can switch between them in a glitchless manner, and another set which _cannot_ be switched between without causing a clock glitch.

All the ones which will glitch are an "auxiliary source", and this tells the clock to do a glitchless switch between the current glitchless-capable clock and whatever is set as the auxiliary source.

A future enhancement that might want to have a helper: The idea is that when you want to switch between one of the clocks that can't be switched glitchless: you can switch the glitchless source to something else (such as the reference oscillator), switch the aux source to whatever new source you intend, and then switch the glitchless source back to the aux source.

@9names 
>Looks good. Not sure how I feel about setting a clock as it's own source though.
>I guess we'll find out how good an idea that is...

See above.

rp2040-hal/src/clocks/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+        shared_dev.clk_sys_ctrl.write(|w| {
+            w.src().clear_bit();
+            w
+        });
+
+        self.await_select(0x1);
+    }
+
+    /// WIP - Helper function to select new source (blocking)
+    pub fn await_select(&self, clock: u8) {
```
anall, line 182: This is currently a bit of a pain to use from a user perspective. I don't need to care what number corresponds to what clock source when setting it, but I do need to know that to await it being selected.

You may have a better idea on how to improve this, but one potential way is to have the clock source switches that can be awaited just return a marker type that you can call await on (or just throw it away if you don't care)

anall (changes requested):
Looks pretty reasonable, and I don't see any real showstoppers besides the `write` vs `modify` problem.

I did notice that the following clocks are missing:

- [ ] clk_gpout0-3
- [ ] clk_adc
- [ ] clk_usb
- [ ] clk_usb

rp2040-hal/src/clocks/mod.rs
9names: Oh, this is supposed to be doing the glitchless aux switch?
Can I suggest it gets a different and more descriptive name if that's the case, because it just looks like all the other clk and aux selectors.

anall: It performs a glitchless switch to whatever the aux source/mux is set to.

Quoting the datasheet:
>All clock generators have a multiplexer referred to as the auxiliary (aux) mux. This mux has a conventional design
whose output will glitch when changing the select control. Two clock generators (clk_sys and clk_ref) have an additional multiplexer, referred to as the glitchless mux. The glitchless mux can switch between clock sources without generating a glitch on the output.

The functions ending in `_src` control that glitchless mux, the functions ending in `_auxsrc` control the auxiliary mux.

I have no objection to changing the name, but I'm not sure what would be better off the top of my head.

rp2040-hal/src/clocks/macros.rs
```diff
@@ -0,0 +1,210 @@
...
+                    w.src().rosc_clksrc_ph();
+                    w
+                });
+            }
+        }
+    };
+}
+
+macro_rules! selfaux_source {
+    ($name:ident, $ctrl:ident, $self:ident) => {
```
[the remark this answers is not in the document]
Nic0w: We sure can :)

rp2040-hal/src/clocks/macros.rs
Nic0w: Yeah, when I wrote the code I was still confused about the difference between `wirte` and `modify`. Those need to be changed.

rp2040-hal/src/clocks/mod.rs
Nic0w: Thanks for the explanation; I felt like the naming of the register's bits were quite confusing. What I understood was that those clocks had a source and an auxiliary source. But the naming of the bits sounds like the clock can be its own source, somehow ?! or its own auxiliary source. And so yeah, hence the `self` keyword and that comment !

Same here, no objection to changing the name but I'm still unsure what to call it exactly.

rp2040-hal/src/clocks/mod.rs
Nic0w: Agreed. I like the idea of the Marker type !

rp2040-hal/src/clocks/mod.rs
9names: Okay, so how about we talk about what we're trying to accomplish here - maybe the name will become obvious when we state what our goals are and how we are going to attempt to solve them.
And maybe it will become obvious if I'm missing a part of the problem...

Changing clock source (glitchless mux) is always glitchless.
Changing aux source (aux mux) can glitch, but that doesn't matter if it isn't selected as a clock source, 
So we don't want to switch aux when it's currently being used as the clock source.

1) Should we write this in a way that makes an aux-aux transition impossible while it's active?
2) Do we want a way of expressing that we want to switch the aux mux from one source to another, by switching clock source to something else (a fallback?), switching aux source, and then selecting aux as the clock source?

anall: We also have to consider the clocks without a glitchless mux. To properly switch the clocks with only an auxiliary source per the datasheet you need to disable the clock, wait 2 cycles of the source clock, switch the aux source, reenable the clock, and then finally wait an additional 2 cycles of the _new_ source clock.

If we are going to enforce the "switch glitchless mux away, switch aux mux, switch glitchless mux back" procedure, we should also enforce the "disable, switch aux mux, reenable" procedure.

The chip only has the glitchless mux for clocks that cannot be stopped.

anall:
Is there anything I can help you with to get this finished?

PR #32, opened 2021-05-04
CI improvement: separate tasks, disable cargo test, fix clippy warning

9names:
When trying to merge the UART branch the "cargo test" step was breaking the CI workflow.
CI didn't give any easy feedback as to what was broken.
This PR moves all different CI steps into their own workflow like some other HALs do, so that we can tell what the error is based on which workflow failed. This also means the tests can run in parallel which should improve run times.

I have moved the "cargo test" step in the tests_host.yml file but commented it out.

I will raise an issue to discuss whether we can fix it or we should remove it.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -12,15 +12,11 @@ jobs:
           target: thumbv6m-none-eabi
           override: true
           profile: minimal
-          components: rustfmt, clippy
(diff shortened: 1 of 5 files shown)


tdittr (approved):
Looks good to me. We can reactivate the tests when we get them fixed.

PR #34, opened 2021-05-08
Refactor GPIO to be more in line with `atsamd` HAL crates

anall:
Per my comment on #4, creating a draft PR with my "minimum" viable example.

Bring more in line with the `atsamd` HAL crates (which in my opinion seems to be one of the more mature HAL crate sets).

Instead of having types `Gpio0` ... `Gpio29` with the pin implementation have a single Type `Gpio<PinId,Mode>` and turn the `Gpio0` ... `Gpio29` into `PinId`s instead.

This allows a function that can be generic on any Output pin (in cases where we can't just use the `embedded-hal` traits):

    fn example<P : PinId>(pin : &mut Gpio<P,Output>) { ... }

This will also possibly aid in implementing the other peripherals.
For PWM, you could add marker traits to the various PinId marker types:

    impl PwmA<Channel0> for Gpio0 {}
    impl PwmB<Channel0> for Gpio1 {}
    impl PwmA<Channel0> for Gpio16 {}
    impl PwmB<Channel0> for Gpio17 {}

To enforce which pins can be assigned to which PWM channels.

This should also aid in creating a type-erased version of the PWM interface which The Rust Embedded Book also recommends.

I still need to shuffle some implementation around so some trait implementations can be generic instead of inside the macro, but "minimum" viable.

I did go a bit past "minimum" with this, including splitting out the pull-up/down resistor state into separate types (as `atsamd` does), as well as adding the ability to disable the input buffer.

This also includes a couple of other scattered changes such as making `pad(&self)`, `sio(&self)`, and `gpio_ctrl(&self)` into `pub(crate)` visibility, and making `sio(&self)` unsafe, as the caller needs to concern itself with what bits get adjusted, but if this is wanted I will probably be refactoring the register access stuff to be more in line with `atsamd`, and also to allow moving as much as possible outside the macro and into generic traits instead.


9names:
I finally got a long enough block of time to look through the atsamd gpio-v2 code to get a feel for how they implement the generic pin + degrade functionality. It looks neat! It is a *lot* of code to get it all working correctly though.
I'm assuming your goal is to make our interface as similar to the atsamd as possible so we can lift in their traits/generic fn's in with little to no modification?

TLDR: this looks good so far, the target impl looks nice, +1 from me.

anall:
More work to bring this in line with how the atsamd crate does things. Haven't implemented dynpin yet.

bradleyharden:
Hey! I'm glad you found this useful enough to copy. I take that as a huge compliment.

> It is a lot of code to get it all working correctly though.

Yeah, that's one down side to this. My only counter argument would be that, even though it's a lot of code, I think it's ultimately easier to read & understand than macros. And on top of that, it's more powerful.

Have you seen the [`bsp_pins!`](https://atsamd-rs.github.io/docs/samd21g/thumbv6m-none-eabi/doc/atsamd_hal/macro.bsp_pins.html) macro? As we've been porting our BSPs to `v2`, it has turned out to be quite useful for defining `Pin` aliases in various `PinMode`s.

anall:
@bradleyharden Yeah, `atsamd` has _really nice_ GPIO implementation from a type system perspective.

I do need to examine how you/atsamd handle assigning the various pins to UART or PWM or whatever peripherals so I can do something similar here.

@9names I think this is getting to the point where I should remove some bits I inserted for testing only, do a cleanup and documentation pass, fix clippy, and then mark this as ready for review.

I feel this is pretty much ready. The ability to assign pins to the various peripheral HAL types can come later.

bradleyharden:
@anall, how we handle SERCOM pads has undergone a lot of revision. I think the latest revision is final though. You can see the corresponding docs [here](https://atsamd-rs.github.io/docs/samd21g/thumbv6m-none-eabi/doc/atsamd_hal/sercom/v2/pad/index.html). Our new SPI module makes use of them for the [Pads type](https://atsamd-rs.github.io/docs/samd21g/thumbv6m-none-eabi/doc/atsamd_hal/sercom/v2/spi/index.html#pads). You also might be interested in a summary of different type-level techniques I put together in the [typelevel module docs](https://atsamd-rs.github.io/docs/samd21g/thumbv6m-none-eabi/doc/atsamd_hal/typelevel/index.html).

rp2040-hal/src/prelude.rs
```diff
@@ -1,3 +1 @@
 //! Prelude
-pub use crate::gpio::GpioExt;
-pub use crate::sio::Sio;
```
anall, line 3: Removed this as I realized this should never have been here. The prelude should really be traits only.

rp2040-hal/Cargo.toml
```diff
@@ -10,13 +10,14 @@ license = "MIT OR Apache-2.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-cortex-m = "0.7.1"
```
anall, line 13: can undo these version changes if needed, not really sure why I included them here

9names (commented):
Looks good. There are a couple of minor nitpicks about copy-paste oversights, but that's all I found that was worth mentioning.
The addition of the feather board support crate is nice too, was not expecting that.

rp2040-hal/src/gpio/reg.rs
```diff
@@ -0,0 +1,353 @@
...
+            }
+            Function(func) => {
+                use DynFunction::*;
+                fields.funcsel = match func {
+                    Xip => 0,
+                    Spi => 1,
+                    Uart => 2,
+                    I2C => 3,
+                    Pwm => 4,
+                    // Pio => 5
```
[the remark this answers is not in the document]
anall: That is just a placeholder saying "Sio (typo in the code) is function 5" -- I've clarified that line

anall:
>The addition of the feather board support crate is nice too, was not expecting that.

Needed something to test `bsp_pins!` on, and didn't have a Pico handy

9names (approved):
LGTM

Issue #35, opened 2021-05-09
Hardware Divider/Modulo support


9names:
See section 2.3.1.5 of the [reference manual](https://datasheets.raspberrypi.org/rp2040/rp2040-datasheet.pdf) for more info on this peripheral.

This is part of the SIO peripheral, and may require use of a mutex or the hardware spinlocks provided by the SIO to use safely. In the initial implementation, having it block until complete would be recommended.
You can also assume (for the initial implementation) that it's never used by interrupts or by the 2nd core.

Minimum deliverable should be standalone function(s) with examples that demonstrate that results match built-in / and % operators.

eutampieri:
I'd like to contribute to this issue but I'm new to embedded rust. Would this be a good first issue nonetheless?

9names:
I think it is a good first issue if you're interested in getting into Embedded Rust HALs.
If you're just starting out you might find it a bit intimidating, but please feel free to ask for help or advice if you need it.

eutampieri:
If I understand correctly it is a hardware accelerated division (or modulo), right?
I’ll then have to get the SDK and try to implement it.
How is the PR linked to this specific issue? Is it the SIO spin lock implementation?

9names:
The hardware is described in detail in the datasheet that is linked in the issue description.
We aren't using the SDK, you would be accessing the hardware through the PAC crate https://github.com/rp-rs/rp2040-pac.
This is already included as a [dependency of the HAL](https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/Cargo.toml), so you can clone this repo, add a new file to rp2040-hal/src/, and start writing the function.

9names:
The PR linked to this issue was related to splitting out the SIO peripheral into chunks.
Until then, it was used exclusively by the GPIO module.
The most common model for Embedded Rust HALs is that a single module will *own* a portion of the memory-mapped register space, as that eliminates issues around concurrent access to the registers.
The SIO is used for GPIO, divide/modulo, spinlocks, message passing. Those things don't really all belong under the GPIO module, hence the PR to allow other modules to own their own space.

9names:
Initial support was merged in PR https://github.com/rp-rs/rp-hal/pull/40
It could use more testing, and will need some work to make it safe across interrupts and multiple cores - if anyone is interested in working on these tasks please open a new issue / draft PR.

eutampieri:
Yeah sorry, I had a look but now I’m rather busy…

PR #36, opened 2021-05-10
add module to manage ownership of parts of SIO

anall:
This is mostly a random thought, and I hope this is useful.

I stuck a TODO in my proof-of-concept draft in #34 to this effect, but now that we have #35 opened I feel like this is actually needed and I took the initiative to add this. We can't have GPIO `Parts` claiming ownership of `SIO`.

I'm not sure how other HAL crates handle this, as I think SIO may be somewhat unique to the rp2040.

I'm considering everything else in SIO out of scope for this, just enough so we no longer have to pass in `pac::SIO` into creating the GPIO object.

I'm just currently handing out a "marker" type to prove ownership for `gpio_bank0` but access to other parts of SIO might actually make sense to have a "useful" type instead.


9names:
Not random at all, I was thinking the same thing when looking at the multiplier.
This is a reasonably common idiom for HALs (for clock trees, power control, etc).

rp2040-hal/src/sio.rs
```diff
@@ -0,0 +1,41 @@
+//! Single Cycle Input and Output (CIO)
```
9names, line 1: CIO?

PR #38, opened 2021-05-15
Have the `pll` and `uart` HAL modules bring the hardware out of reset

anall:
Currently, `pll` and `uart` do not bring the module out of reset before attempting to use it.

If the module is still in reset, you cannot use it.

This PR brings the UART and PLL out of reset before attempting to use them.

### DIFF
--- a/rp2040-hal/src/pll.rs
+++ b/rp2040-hal/src/pll.rs
@@ -14,6 +14,8 @@ use embedded_time::{
 
 use nb::Error::WouldBlock;
 
+use crate::resets::SubsystemReset;
+
 /// State of the PLL
 pub trait State {}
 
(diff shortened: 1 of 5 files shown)


rp2040-hal/src/pll.rs
```diff
@@ -50,6 +50,32 @@ pub struct PhaseLockedLoop<S: State, D: PhaseLockedLoopDevice> {
...
+    pub trait PhaseLockedLoopReset: PhaseLockedLoopDevice {
+        fn bring_pll_up(&self, resets: &mut rp2040_pac::RESETS);
+    }
+}
+
+impl private::PhaseLockedLoopReset for rp2040_pac::PLL_SYS {
+    fn bring_pll_up(&self, resets: &mut rp2040_pac::RESETS) {
+        resets.reset.modify(|_, w| w.pll_sys().clear_bit());
+        while resets.reset_done.read().pll_sys().bit_is_clear() {
+            cortex_m::asm::delay(10);
```
9names, line 65: The delays inside these busy-wait loops don't add any value.
Having nothing in the body does the same thing but with no code cost, and will allow us to finish sooner.

anall: Was just copying how `gpio` handled the resets. I believe you are right that an empty body would be better.

Also, a note in case it comes up: I know I'm duplicating code here, but given there are only 2 options each I feel like the duplicated code is better than introducing a maco.

anall:
Upon more consideration, I realized this is probably a better way anyway instead of having multiple traits and did the refactoring to a macro now.

This does change GPIO to reset the two subsystems (IO_BANK0 and PADS_BANK0) one after the other instead of simultaneously. I'm not sure how much this really matters for something that happens once at startup, but if it does, we could probably return the duplicated code to GPIO instead of trying to special-case that in the `resets` module.

thejpster (approved):
Looks OK to me. Thanks!

9names (approved):
LGTM

PR #39, opened 2021-05-15
Watchdog implementation

jspaulsen:
I have limited experience in this space (both embedded and rust) so I was hoping to iterate on the design and implementation of this from feedback.

I'm aware of two things that may drive design (not to say there aren't more):
* Tick generation should be enabled prior to enabling the watchdog
* RP2040-E1

In regards to RP2040-E1, should we restrict the counter to the maximum load value (0xFFFFFF / 2) and if so, how?

### DIFF
--- a/rp2040-hal/src/watchdog.rs
+++ b/rp2040-hal/src/watchdog.rs
@@ -8,15 +8,14 @@ use embedded_time::{
 use crate::pac::WATCHDOG;
 
 
-//const MAX_LOAD: u32 = 0x7FFFFF;
-
-
+///
 pub struct Watchdog {
     watchdog: WATCHDOG,
     delay_ms: u32,
 }
 
 impl Watchdog {
+    ///
     pub fn new(watchdog: WATCHDOG) -> Self {
         Self {
             watchdog,
@@ -24,6 +23,7 @@ impl Watchdog {
         }
     }
 
+    ///
     pub fn enable_tick_generation(&mut self, cycles: u8) {
         const WATCHDOG_TICK_ENABLE_BITS: u32 = 0x200;
 
@@ -57,11 +57,11 @@ impl watchdog::WatchdogEnable for Watchdog {
     /// Due to a logic error, the watchdog decrements by 2 and
     /// value must be compensated; see RP2040-E1
     fn start<T: Into<Self::Time>>(&mut self, period: T) {
-        let delay_ms = period
+        self.delay_ms = period
             .into()
             .integer() * 2;
 
-        self.load_counter(delay_ms);
+        self.load_counter(self.delay_ms);
         self.enable();
     }
 }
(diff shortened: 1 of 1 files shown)


9names:
This is looking really good so far.
I didn't know there was the option to disable the watchdog on debug, that's nifty!

> should we restrict the counter to the maximum load value?

You usually enable a watchdog for safety or reliability.
In either of those cases, you do not want to hide the fact that you are not giving the user the timeout value they asked for. So clamping the value to the maximum isn't really a good option. Ignoring it isn't really great, either.

As for what to do about it:
The alpha 1.0.0 watchdog start trait is fallible, so in that version you could just return an error from the start function
https://github.com/rust-embedded/embedded-hal/blob/e8354100b19f6068dbcd74d2a646934295406fc3/src/watchdog.rs#L36
But since we're on embedded-hal 0.2.4 and won't be moving to 1.0 until it's released, I think the easiest answers are either to ignore it (and leave a TODO in there for 1.0), or to panic in `fn start()` if the reload value is out of range.

Szpadel:
> I think the easiest answers are either to ignore it (and leave a TODO in there for 1.0), or to panic in fn start() if the reload value is out of range.

I vote for panic, silent errors errors are very hard to detect, and give false sense that everything is working correctly

rp2040-hal/src/watchdog.rs
```diff
@@ -1,3 +1,91 @@
...
+            w.pause_dbg0()
+                .bit(pause)
+                .pause_dbg1()
+                .bit(pause)
+                .pause_jtag()
+                .bit(pause)
+        })
+    }
+
+    fn load_counter(&self, counter: u32) {
```
thejpster, line 52: Isn't this unsafe? What is they give a value greater than MAX_PERIOD here?

Maybe the period should be a newtype with a const new function that panics if out of bounds, so it can blow up at compile time in a const context.

jspaulsen: Thanks for the feedback!

> Isn't this unsafe? What is they give a value greater than MAX_PERIOD here?
I'm not sure I understand what you mean by `unsafe`.  If the concern is overwriting reserved bits, the value is masked to 0xFF_FFFF [here](https://github.com/rp-rs/rp2040-pac/blob/4006f1937b8f57c4b4d6543bfd7aa7158a640fb0/src/watchdog/load.rs#L19).  Otherwise, the function is private and used exclusively by the Watchdog HAL trait implementations below (which panics if the provided period exceeds `MAX_PERIOD`)

> Maybe the period should be a newtype with a const new function that panics if out of bounds, so it can blow up at compile time in a const context.

I'm not sure I know this pattern, could you provide an example?

thejpster: Ah, OK. I hadn't spotted that was was a non-pub function and yes, I can see that the PAC limits the values to MAX_PERIOD anyway.

thejpster:
Seems OK to me, but it's draft so it can't be approved.

9names (approved):
Looks good to me

PR #40, opened 2021-05-22
Hardware divide/modulo support

Rudo2204:
A simple working implementation of hardware divide/modulo, tested on hardware.
I don't have much experience (both embedded and Rust) so there are a bunch of questions I want to ask:
* In the `sio.rs` module, there are comment that say:
```rust
// we can hand out other things here, for example:
// gpio_qspi
// divider
// interp0
// interp1
```
but I could not figure out how to fit my divmod implementation in there.
* For this initial implementation I split it into two different structs for unsigned and signed divmod operations. Is it possible to define a more generic implementation? I thought about using the trait bound `impl Into<u32>` or `impl Into<i32>` but both doesn't capture everything.
* I shamelessly copied the `sio()` function from the module `gpio.rs`. However, I'm not so sure about the safeness of it in this implementation. Also I had to duplicate the this piece of code since I split to two different structs. Not happy about this.
* I'm not sure if defining a different struct for result is even idiomatic rust. I thought about simply returning a tuple, but then the user would have to look at the documentation to know which one is the remainder/quotient.
* For signed divmod operation, I have to use `as u32` and `as i32` to cast the value around to make the compiler happy. I wonder if there is a overhead doing this. Is there a better way to do this?

Things that I also want to implement and would like some pointers for them:
* Maybe don't use blocking delay. In the datasheet it say we are free to do other operations except divmod during these 8 cycles.
* Logic to handle interrupt.
* Also implement for 2nd core.

### DIFF
--- a/rp2040-hal/src/sio.rs
+++ b/rp2040-hal/src/sio.rs
@@ -19,15 +19,30 @@ pub struct SioGpioBank0 {
     _private: PhantomData<u32>,
 }
 
+/// Marker struct for ownership of divide/modulo
+pub struct HwDivide {
+    _private: PhantomData<u32>,
+}
+
+/// Result from signed divide/modulo operation
+pub struct DivResult<T> {
+    /// The remainder result from signed divide/modulo operation
+    pub remainder: T,
+    /// The quotient result from signed divide/modulo operation
+    pub quotient: T,
+}
+
 /// Struct containing ownership markers for managing ownership of the SIO registers.
 pub struct Sio {
-    _sio: pac::SIO,
+    /// SIO
+    pub sio: pac::SIO,
 
     /// GPIO Bank 0 registers
     pub gpio_bank0: SioGpioBank0,
+    /// 8-cycle hardware divide/modulo module
+    pub hw_divide: HwDivide,
     // we can hand out other things here, for example:
     // gpio_qspi
-    // divider
     // interp0
     // interp1
 }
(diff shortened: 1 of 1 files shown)


Rudo2204:
Now I think of it, create a new `DivModUnsigned` and then call unsigned_div_mod on it to perform the operation seems really strange...

9names:
Yeah, this seems a bit too much boilerplate - I don't think people will enjoy using it.
Probably best to aim for a blocking implementation first, 8 cycles is still cheap for what we get here.
My thoughts for implementing it looked more like this:
```rust
pub struct DivResult<T> {
    pub quotient: T,
    pub remainder: T,
}

pub fn div_signed(dividend: i32, divisor: i32) -> DivResult<i32> {
    // Insert implementation here
    DivResult{quotient,remainder}
}

pub fn div_unsigned(dividend: u32, divisor: u32) -> DivResult<u32> {
    // Insert implementation here
    DivResult{quotient,remainder}
}
```
Ideally, access to these functions would be restricted to owners of an object handed out by Sio::new() similar to SioGpioBank0
https://ferrous-systems.com/blog/zero-sized-references/#singleton-ish
So if that singleton type were HwDivide, I'd name them to HwDivide.signed() and HwDivide.unsigned() since the context makes it pretty obvious what the functions are going to do.

Rudo2204:
@9names thank you for the review and the link. I have rewrite the implementation according to your feedback.
Can you please take a look at it? Thank you.

9names:
This is looking better, nice!

Passing in a ref to sio every time you call it feels a bit redundant - normally if you need something every time you make it a member function (so you can access it through &self).
In this case you don't need access to *all* of sio to do divides, and if you make the divide functions members functions of sio it isn't really great either - it will become a very busy namespace, plus you'll often run into ownership issues.

One answer to this is to steal access to the peripheral instead, and use a marker object to ensure you're the only one accessing the registers at a time.
To do this, bring back the HwDivide struct with it's PhantomData, then make a pointer to the SIO in your functions so you can access the hardware:

```rust
impl HwDivide {
    /// Perform hardware unsigned divide/module operation
    pub fn div_unsigned(&self, dividend: u32, divisor: u32) -> DivResult<u32> {
        let sio = unsafe { &(*pac::SIO::ptr()) };
        sio.div_sdividend.write(|w| unsafe { w.bits(dividend) });
        ...
```

Then, users of of the HAL can write code like this:
```rust
let sio = Sio::new(pac.SIO);
let divider = sio.hwdivide;
let result = divider.div_unsigned(10,2);
```
This way it's obvious who has the hardware divider and doesn't make you repeat yourself

anall:
A note that you don't even need to do `PhantomData`.

I realized after the Sio PR that `()` is a completely valid type in this case and will still prevent someone outside from constructing it, and is cleaner than using PhantomData with a random type. In fact, I thought I actually fixed that for SioGpioBank0 -- but it seems I forgot to push the commit to my branch or something?

Rudo2204:
@9names I think this is ready to be reviewed again. The last `2fc0dc7` commit changes `PhantomData` to an empty tupple like @anall suggested. I'm not sure if we should split that into another PR or not, but I have included it here.

9names (approved):
LGTM

PR #43, opened 2021-06-20
split rom_data into safe and unsafe functions

anall:
Most of the functions provided in `rom_data` can do unsafe things.

You can use `memset`/friends to modify Rust data structures without limits. It is safe Rust to get a raw (const or not) pointer to something, and it is also safe Rust to turn a `*const T` into a `*mut T`. It is only unsafe to dereference the raw pointer.

Any operations from the ROM that allow pointer operations on memory (ergo a "dereference") should be unsafe.
I believe the operations that change how the flash is set up should also be considered unsafe, as those may cause undefined behavior in the code (especially, but not only, if you are using the QSPI pins as GPIO.)

`wait_for_vector` has some preconditions for when it should be used, so should probably be unsafe as well.

I see no unsafe behavior from `popcount32`, `reverse32`, `clz32`, or `ctz32` unless there is some weird behavior in the ROM I am unaware of.

I am not sure if `reset_to_usb_boot` should be considered unsafe. I do not believe that can cause undefined behavior in the following code as it forces a reset, but I am not sure if we want to gate that behind an unsafe as an "are you sure"?

I am not sure if there's a better way to do this with the macro, as Rust macros are still not my strong point.

### DIFF
--- a/rp2040-hal/src/rom_data.rs
+++ b/rp2040-hal/src/rom_data.rs
@@ -52,6 +52,25 @@ macro_rules! rom_funcs {
     }
 }
 
+macro_rules! rom_funcs_unsafe {
+    (
+        $(
+            $(#[$outer:meta])*
+            $c:literal $name:ident (
+                $( $aname:ident : $aty:ty ),*
+            ) -> $ret:ty ;
+        )*
+    ) => {
+        $(
+            $(#[$outer])*
+            pub unsafe fn $name($( $aname:$aty ),*) -> $ret{
+                let func:  extern "C" fn( $( $aty ),* ) -> $ret = rom_table_lookup(FUNC_TABLE, *$c);
+                func($( $aname ),*)
+            }
+        )*
+    }
+}
+
 rom_funcs! {
     /// Return a count of the number of 1 bits in value.
     b"P3" popcount32(value: u32) -> u32;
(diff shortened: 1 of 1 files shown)


9names:
This looks good to me. 
I don't think we should mark things as unsafe unless they actually are. A reset is a normal part of a system lifecycle, we should assume that a user that calls it knows what they're doing.
I am also not a macro expert, so I'm not going to cast any judgement here.

anall:
So then all I'm really not 100% sure about are the flash functions (but those can do raw memory writes, so probably should be unsafe) and `wait_for_vector` (but I believe the only "harm" that can do is lock up the processor if called on the wrong core)

The flash related functions can also mess up the QSPI-as-GPIO pin configuration if you are using that.

I think both should stay unsafe, but I am on the fence about `wait_for_vector`

9names:
My feeling: flash funcs are unsafe for the reasons you've stated.
wait_for_vector is technically safe as far as I'm concerned, but I'm also on the fence about whether it's worth moving it into the safe funcs list. I don't think it's something we should spend too much thought on - it's likely to only be called once at startup.
Maybe we should wait until we try using it before deciding where to put it?

anall:
>I don't think it's something we should spend too much thought on - it's likely to only be called once at startup.

We shouldn't even need to call it in normal operation. The bootrom calls it for us on the second core. I think the only reason we would need to call that function (on the second core) is if we want to effectively reset the second core (from the second core).

>Maybe we should wait until we try using it before deciding where to put it?

Reasonable.

anall:
It seems that this somehow got accidentally reverted at some point. Shall I just open a new PR to un-revert it?

9names:
Yep, new PR please

PR #44, opened 2021-06-22
PWM implementation

uncheckedexception:
Not an expert at embedded rust, but I needed PWM functionality for a project, so decided to implement it myself.

Includes almost every PWM method the RP2040 allows for (the entire list is at the top of the src/pwm.rs file)

If there are any major issues, it's probably centered around the fact that some pins share PWM blocks but can set different duty cycles. For example, the set_duty(), get_duty(), set_inv(), and clr_inv() write to unique registers for each PWM pin, but the set_top(), set_ph_correct(), and clr_ph_correct() methods write to the same register for PWM pins 0 and 1, since they share the same internal PWM block.

There's also a quick example at examples/pwm_blink.rs that toggles a PWM pin between duty cycles of about 25% and 50%


rp2040-hal/src/pwm.rs
```diff
@@ -1,3 +1,312 @@
...
+//! min_config() leaves those registers in the state they were before it was called (Careful, this can lead to unexpected behavior)
+//! It's recommended to only call min_config() after calling default_config() on a pin that shares a PWM block.
+
+#[doc = "Trait to be implemented on the HAL, containing methods to extract PWM pins"]
+pub trait PwmExt<PADS, IO> {
+    #[doc = "The type that PWM pins are stored in"]
+    type Parts;
+
+    #[doc = "Method that extracts PWM pins into HAL objects"]
+    fn split(self, pads: PADS, io: IO, reset: &mut rp2040_pac::RESETS) -> Self::Parts;
```
anall, line 55: Following the [C-CTOR recommendation](https://docs.rust-embedded.org/book/design-patterns/hal/predictability.html) in the Rust Embedded book, we shouldn't have a `split` method, and instead, define a type we can call `new` on.

rp2040-hal/src/pwm.rs
```diff
@@ -1,3 +1,312 @@
...
+}
+
+macro_rules! pwm {
+    ($PWMX:ident, $pwmx:ident, $GPIOX:ident, $gpiox:ident, $PADSX:ident, $padsx:ident, $pwmxs:expr, [
```
anall, line 59: This takes over the GPIO pins, meaning we can use GPIO _or_ PWM, but not both.

My massive GPIO refactor (#34) should make it easier to not need to do something like this, and instead be able to do something like:

```
let led_pin = pins.gpio10.into_mode::<FunctionPwm>();
let pwm5 = pwm.pwm5;
pwm5.configure( …, led_pin );
pwm5.set_duty(15000);
```

That way GPIO and PWM can coexist (and UART, and I2C, and whatever else we have)

rp2040-hal/src/pwm.rs
```diff
@@ -1,3 +1,312 @@
...
+                }
+            )+
+        }
+    };
+}
+
+// This is stupidly long because rust's concat_idents feature is only available in nightly builds.
+// TODO: See if this can be condensed by creating an array of idents somehow? (no idea if that's possible)
+pwm!(
+    PWM, pwm, IO_BANK0, io_bank0, PADS_BANK0, pads_bank0, "PWM", [
```
anall, line 284: I'm not sure what you are doing here with 28 PWM objects. The RP2040 only has 8 PWM channels.

With this, you can try and use both `pwm0` and `pwm16` at the same time but that will conflict and try to use the same PWM channel for 2 different pins.

anall:
Thinking about this more: the SVD in `rp2040-pac` should probably be patched to cluster the various `ch*_…` PWM registers to make accessing them easier.

I can make the change and send the PR for that if you'd like, but that will break your code here.

tyler-stowell:
Thanks for all the advice! I made changes to conform to the C-CTOR recommendations and reduced the number of PWM's from 28 to 8.

In terms of the issue with the PWM interface taking over GPIO pins, the PWM pins only use the GPIO pins for initial setup, so I temporarily adjusted it so that it takes temporary references instead of ownership. This is definitely not a good fix, since it only lets you create PWM pins before GPIO pins, and there's nothing to stop you from trying to use PWM and GPIO on the same pin. However, it should make it easier to adjust the code to match the structure you were envisioning once the GPIO refactor is complete.

Unfortunately, up until the GPIO refactor is complete, there's probably not a good way to finish this PWM implementation in a safe way, but thanks for the advice anyways!

Also, in terms of adjusting the rp2040-pac to group the various ch*_... registers, I don't know if it's necessary, but if you were to make the change, I'd be happy to adjust my code here to conform. Thank you!

anall:
And the registers have been clustered

tyler-stowell:
Clustering the registers definitely helped to clean up this code. Thanks a lot!

rp2040-hal/src/pwm.rs
```diff
@@ -88,7 +87,7 @@ impl $PXi {
     // will instead receive a single gpio pin.
     fn init_io(&self, pwm: &mut pac::$PWMX, pad : &mut pac::PADS_BANK0, io : &mut pac::IO_BANK0, resets: &mut pac::RESETS) -> () {
         //TODO: Merge these into gpio.rs split function after GPIO refactor. At the moment, this is here because these need to be reset for
-        // the PWM to work. However, because they're here, they'll be reset every time a new PWM pin is created (which is bad). Definitely needs to change once GPIO is redone.
+        // the PWM to work. However, because they're here, they'll be reset every time a new PWM pin is created (BAD).
         pwm.reset_bring_up(resets);
```
anall, line 91: This actually shouldn't reset them a second time. `reset_bring_up` only brings the peripheral out of reset. Should be a no-op (and a waste of a few cycles) if the peripheral has already been brought up.

leo60228:
> Unfortunately, up until the GPIO refactor is complete, there's probably not a good way to finish this PWM implementation in a safe way, but thanks for the advice anyways!

That PR has been merged!

9names (approved):
LGTM, we'll do the GPIO-related cleanup in seperate PR.
Thanks your contribution @uncheckedexception

PR #45, opened 2021-06-25
Fix gpio docs

hmvp:
The sio argument does not work with a bare pac SIO struct


9names (approved):
Yep, good catch.
LGTM

PR #48, opened 2021-06-27
Add missing clocks

hmvp:
Also reorderd the macro invocations to match the datasheet

### DIFF
--- a/rp2040-hal/src/clocks/mod.rs
+++ b/rp2040-hal/src/clocks/mod.rs
@@ -42,6 +42,34 @@ impl ClocksManager {
         self.clocks
     }
 
+    /// Getter for the GPIO Output 0 Clock.
+    pub fn gpio_output0_clock(&self) -> GpioOutput0Clock {
+        GpioOutput0Clock {
+            shared_dev: self.shared_clocks,
+        }
+    }
+
+    /// Getter for the GPIO Output 1 Clock.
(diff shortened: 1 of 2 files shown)


rp2040-hal/src/clocks/mod.rs
```diff
@@ -101,6 +150,30 @@ pub trait ClockSYSClockAuxSource {
...
     fn set_clksys_auxsrc(&mut self);
 }
+
+/// For clocks that can have the USB Clock as an auxilliary source.
+pub trait ClockUSBClockAuxSource {
+    /// Set USB Clock as source.
+    fn set_clkusb_auxsrc(&mut self);
+}
+
+/// For clocks that can have the USB Clock as an auxilliary source.
```
9names, line 160: Copy paste error on the comment for the ADC, RTC, Ref clock blocks

9names (approved):
Looks good to me. A couple of comment errors due to copy+paste, but other than that this looks good to me.
If you want to fix these go ahead, otherwise we'll do a cleanup post-merge.
Thanks for the PR

rp2040-hal/src/clocks/mod.rs
hmvp: no problem, i will clean them up in a moment

hmvp: Fixed

Issue #47, opened 2021-06-27
Clocks improvements


anall:
Collected suggested improvements from #31:

* Rename `selfaux_source` macro. See: https://github.com/rp-rs/rp-hal/pull/31#discussion_r628847889
* Better name/handling around aux sources and glitchless mux: See discussion: https://github.com/rp-rs/rp-hal/pull/31#discussion_r632477569
  * Better name for `set_self_aux_src`
  * Possibly: Prevent aux-aux transition while the aux source is active on the clocks with glitchless muxes. (Per datasheet: clock glitches on these should be avoided at all costs as the CPU or other important logic is possibly currently running off of the clock.)
  * Possibly: Prevent clock source switches on a running clock w/o the glitchless mux. (Per datasheet: disable the clock, wait 2 cycles of source clock, switch the (aux) source, re-enable clock, wait 2 clock cycles of new source)
*  `await_select` is a bit of a pain from a user perspective, as it requires knowing the internal clock number. Clock source switches probably should return a type that you can call `await` on. See: https://github.com/rp-rs/rp-hal/pull/31#discussion_r632165914

PR #49, opened 2021-06-28
Clock init

hmvp:
Preliminary init function to setup the clocks the same as the C SDK does. I am not sure about the structure but at least it gets everything going...

### DIFF
--- a/rp2040-hal/src/clocks/mod.rs
+++ b/rp2040-hal/src/clocks/mod.rs
@@ -1,11 +1,40 @@
 //! Clocks (CLOCKS)
-// See [Chapter 2 Section 15](https://datasheets.raspberrypi.org/rp2040/rp2040_datasheet.pdf) for more details
+//!
+//!
+//!
+//! Usage:
+//! ```rust
+//! let mut p = rp2040_pac::Peripherals::take().unwrap();
+//! let mut watchdog = Watchdog::new(p.WATCHDOG);
+//! let mut clocks = ClocksManager::new(p.CLOCKS, &mut watchdog);
+//! // Enable the xosc
+//! let xosc = setup_xosc_blocking(p.XOSC, XOSC_MHZ.Hz()).ok().unwrap();
+//!
+//!
+//! // Configure PLLs
+//! //                   REF     FBDIV VCO            POSTDIV
+//! // PLL SYS: 12 / 1 = 12MHz * 125 = 1500MHZ / 6 / 2 = 125MHz
+//! // PLL USB: 12 / 1 = 12MHz * 40  = 480 MHz / 5 / 2 =  48MHz
+//! let pll_sys =
+//!     setup_pll_blocking(p.PLL_SYS, 12.MHz().into(), PLL_SYS_125MHZ, &mut clocks, &mut p.RESETS).ok().unwrap();
+//! let pll_usb =
+//!     setup_pll_blocking(p.PLL_USB, 12.MHz().into(), PLL_USB_48MHZ, &mut clocks, &mut p.RESETS).ok().unwrap();
+//! clocks.init(&xosc, &pll_sys, &pll_usb);
+//! ```
+//!
+//! See [Chapter 2 Section 15](https://datasheets.raspberrypi.org/rp2040/rp2040_datasheet.pdf) for more details
 
-use crate::pac::*;
+use crate::{
+    pll::{Locked, PhaseLockedLoop},
+    watchdog::Watchdog,
+    xosc::{CrystalOscillator, Stable},
+};
+use core::convert::TryInto;
+use embedded_time::rate::*;
+use pac::{clocks, CLOCKS, PLL_SYS, PLL_USB};
 
 #[macro_use]
 mod macros;
-
 #[derive(Copy, Clone)]
 /// Provides refs to the CLOCKS block.
 pub struct ShareableClocks {
@@ -22,6 +51,17 @@ impl ShareableClocks {
     }
 }
 
+const XOSC_MHZ: u32 = 12_000_000_u32;
+
+fn make_div<S: TryInto<Hertz<u32>>, F: TryInto<Hertz<u32>>>(
+    src_freq: S,
+    freq: F,
+) -> Result<u32, ()> {
+    let src_freq = u64::from(*src_freq.try_into().map_err(|_| ())?.integer());
+    let freq = (*freq.try_into().map_err(|_| ())?.integer()).into();
+    let div: u64 = (src_freq << 8).wrapping_div(freq);
+    Ok(div as u32)
+}
 /// Abstraction layer providing Clock Management.
 pub struct ClocksManager {
     clocks: CLOCKS,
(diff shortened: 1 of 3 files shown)


rp2040-hal/src/clocks/mod.rs
```diff
@@ -1,11 +1,40 @@
...
+//!
+//!
+//!
+//! Usage:
+//! ```rust
+//! let mut p = rp2040_pac::Peripherals::take().unwrap();
+//! let mut watchdog = Watchdog::new(p.WATCHDOG);
+//! let mut clocks = ClocksManager::new(p.CLOCKS, &mut watchdog);
+//! // Enable the xosc
+//! let xosc = setup_xosc_blocking(p.XOSC, XOSC_MHZ.Hz()).unwrap();
```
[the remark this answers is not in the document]
hmvp: Ah yes i used `.ok().unwrap()` in my actual implementation

rp2040-hal/src/clocks/mod.rs
```diff
@@ -22,26 +51,141 @@ impl ShareableClocks {
...
         ClocksManager {
             clocks: clocks_block,
             shared_clocks,
         }
     }
 
+    /// Initialize the clocks
+    pub fn init(
+        &self,
+        _: CrystalOscillator<Stable>,
```
hmvp, line 89: These should probably be taken as `&`

rp2040-hal/src/clocks/mod.rs
```diff
@@ -22,21 +51,108 @@ impl ShareableClocks {
...
+        _: &CrystalOscillator<Stable>,
+        _: &PhaseLockedLoop<Locked, PLL_SYS>,
+        _: &PhaseLockedLoop<Locked, PLL_USB>,
+    ) {
+        // Configure clocks
+        // CLK_REF = XOSC (12MHz) / 1 = 12MHz
+        let mut ref_clock = self.ref_clock();
+        let div = make_div(12u32.MHz(), 12u32.MHz()).unwrap();
+        if div > ref_clock.get_div() {
+            ref_clock.set_div(div);
```
9names, line 98: The Pico SDK you copied this from had a comment explaining why div is conditionally set once and then set again after changing clocks, but you didn't copy it across.
I recommend that you include this note because the reason for doing this is not immediately obvious to someone reading this code.

rp2040-hal/src/clocks/mod.rs
```diff
@@ -22,21 +51,108 @@ impl ShareableClocks {
...
+
+        // CLK ADC = PLL USB (48MHZ) / 1 = 48MHz
+        let mut adc_clock = self.adc_clock();
+        let div = make_div(48u32.MHz(), 48u32.MHz()).unwrap();
+        if div > adc_clock.get_div() {
+            adc_clock.set_div(div);
+        }
+        adc_clock.disable();
+        adc_clock.set_pll_usb_auxsrc();
+        adc_clock.enable();
```
9names, line 134: Ah, wouldn't you want to enable the ADC clock inside the ADC peripheral driver?
Same with the other endpoint clocks here. Most HALs will enable the clock on peripheral creation, where you have the most knowledge about how the system is going to be used, and where an error would lead you to the correct place.

rp2040-hal/src/clocks/mod.rs
```diff
@@ -327,30 +483,29 @@ pll_usb_auxsource!(ReferenceClock, clk_ref_ctrl);
...
-        self.await_select(0x1);
+        self.await_select(0x0);
     }
 
     /// WIP - Helper function to select new source (blocking)
     pub fn await_select(&self, clock: u8) {
         let shared_dev = unsafe { self.shared_dev.get() };
 
-        while shared_dev.clk_sys_selected.read().bits() != clock as u32 {
+        while (shared_dev.clk_sys_selected.read().bits() & 1 << clock) == 0 {
```
9names, line 508: Some parens around 1 << clock would be nice just so people don't have to remember whether << or & is evaluated first. Or assign it to a variable first, that would be even easier to read.

rp2040-hal/src/clocks/mod.rs
hmvp: Yes definitely, however I started with this to scratch my own itch and to get something out of the temp sensor. The adc is quite easy to do with just the pac. But for the clocks I basically used the C sdk defaults.  

I feel the clock subsystem needs a lot of rework to support this. But by adding the missing clocks and providing a somewhat sane default at least the rest can work..

rp2040-hal/src/clocks/mod.rs
hmvp: Yeah due to how the current code is structured and how the pac works I could not get an C SDK equivalent set_clock function to work. So I figured the next best thing is to unroll it for every clock so there is at least an easy way to enable them... 

I will add that comment back. But alll that stuff should be abstracted over in a proper function

hmvp:
I improved the remarks. However this subsystem need a bit of rework. If I feel like it I will try to improve it. But that might also need help from the pac

rp2040-hal/src/clocks/mod.rs
9names: Sorry, should have been more clear that I wanted
```
shared_dev.clk_sys_selected.read().bits() & (1 << clock)
```
here, to make it obvious that 1 << clock resolves before being bitwise AND'd

rp2040-hal/src/clocks/mod.rs
```diff
@@ -22,21 +51,113 @@ impl ShareableClocks {
...
+        ref_clock.set_div(div);
+
+        // CLK SYS = PLL SYS (125MHz) / 1 = 125MHz
+        let mut sys_clock = self.sys_clock();
+        let div = make_div(125u32.MHz(), 125u32.MHz()).unwrap();
+        // If increasing divisor, set divisor before source.
+        if div > sys_clock.get_div() {
+            sys_clock.set_div(div);
+        }
+        sys_clock.set_pll_sys_auxsrc();
```
anall, line 112: Probably should make sure we're not on the aux mux before switching this. I know I've managed to get one of my test boards to reset without it resetting the clocks.

rp2040-hal/src/pll.rs
```diff
@@ -264,3 +265,28 @@ impl<D: PhaseLockedLoopDevice> PhaseLockedLoop<Locking, D> {
...
+    dev: D,
+    xosc_frequency: Generic<u32>,
+    config: PLLConfig<R>,
+    clocks: &mut ClocksManager,
+    resets: &mut RESETS,
+) -> Result<PhaseLockedLoop<Locked, D>, Error>
+where
+    R: Into<Hertz<u64>>,
+{
+    // Before we touch PLLs, switch sys and ref cleanly away from their aux sources.
```
anall, line 280: Do we need to do something with the USB or ADC clocks as well here?

anall:
I'm not sure what options we'll have for improving the PAC here unfortunately given some of the limits of svd2rust but if you've got a list of useful improvements I can look into that.

There is #47 which collected requested improvements for this subsystem from the initial PR which we merged to allow future iteration.

rp2040-hal/src/pll.rs
hmvp: Not sure, tbh I feel this is a bit of a hack

rp2040-hal/src/clocks/mod.rs
hmvp: Yeah there should be a generic switch mux function that takes care of that like the C sdk has.. However Rust types wont let me do that easily...

hmvp:
> I'm not sure what options we'll have for improving the PAC here unfortunately given some of the limits of svd2rust but if you've got a list of useful improvements I can look into that.
> 
> There is #47 which collected requested improvements for this subsystem from the initial PR which we merged to allow future iteration.

So this was meant to be just adding the missing clocks and to make a quick init function (that assumes the device just came out of reset)

I can add some improvements to this or merge this and change them in a next pull request.. I am not sure how long that takes..

anall:
You don't need to do any of that here. I was just pointing out #47 because you mentioned "However this subsystem need a bit of rework"

rp2040-hal/src/clocks/mod.rs
9names: Here's the comment you added back

**// If increasing divisor, set divisor before source.**

Compare that with the C SDK comment

**// If increasing divisor, set divisor before source. Otherwise set source before divisor.
// This avoids a momentary overspeed when e.g. switching to a faster source and increasing divisor to compensate.**

The second line there is much more valuable that the first, because it's describing *why* the code is performing actions that way.
The code can help us understand _what_ and _how_, but it does not provide any insight as to _why_ that method is important.

Permalink to the C code, for reference
https://github.com/raspberrypi/pico-sdk/blob/bfcbefafc5d2a210551a4d9d80b4303d4ae0adf7/src/rp2_common/hardware_clocks/clocks.c#L55

We'll fix it up post merge, but please try to keep in mind that a person reading your code has less knowledge of the problem than you do, and it's in your best interest to help them understand what you're trying to do.

9names:
LGTM, thanks!
We'll pick up any improvements in new PRs to keep this moving

PR #50, opened 2021-07-02
Spi

hmvp:
Basic implementation based on C SDK code. Does not seem to work yet

### DIFF
--- a/rp2040-hal/src/spi.rs
+++ b/rp2040-hal/src/spi.rs
@@ -1,3 +1,218 @@
 //! Serial Peripheral Interface (SPI)
-// See [Chapter 4 Section 4](https://datasheets.raspberrypi.org/rp2040/rp2040_datasheet.pdf) for more details
-// TODO
+//!
+//! See [Chapter 4 Section 4](https://datasheets.raspberrypi.org/rp2040/rp2040_datasheet.pdf) for more details
+//!
+//! ## Usage
+//!
+//! ```rust
+//! let sio = Sio::new(p.SIO);
+//! let gpio_pins = Pins::new(p.IO_BANK0, p.PADS_BANK0, sio.gpio_bank0, &mut p.RESETS);
+//!
+//! // Color
+//! let _ = gpio_pins.spi_sclk.into_mode::<FunctionSpi>();
+//! let _ = gpio_pins.spi_mosi.into_mode::<FunctionSpi>();
+//!
+//! let spi = Spi::<_, _, 8>::new(p.SPI0).init(&mut p.RESETS, 125_000_000, 16_000_000, &MODE_0);
+//! ```
+
+use crate::resets::SubsystemReset;
+use core::{convert::Infallible, marker::PhantomData, ops::Deref};
+use embedded_hal::blocking::spi;
+use embedded_hal::spi::{FullDuplex, Mode, Phase, Polarity};
+use pac::RESETS;
+
+/// State of the SPI
+pub trait State {}
+
+/// Spi is disabled
+pub struct Disabled {
+    __private: (),
+}
+
+/// Spi is enabled
+pub struct Enabled {
+    __private: (),
+}
+
+impl State for Disabled {}
+impl State for Enabled {}
+
+/// Pac SPI device
+pub trait SpiDevice: Deref<Target = pac::spi0::RegisterBlock> + SubsystemReset {}
+
+impl SpiDevice for pac::SPI0 {}
+impl SpiDevice for pac::SPI1 {}
+
+/// Data size used in spi
+pub trait DataSize {}
+
+impl DataSize for u8 {}
+impl DataSize for u16 {}
+
+/// Spi
+pub struct Spi<S: State, D: SpiDevice, const DS: u8> {
+    device: D,
+    state: PhantomData<S>,
+}
+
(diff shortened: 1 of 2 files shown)


rp2040-hal/src/spi.rs
```diff
@@ -1,3 +1,184 @@
...
+            }
+        }
+
+        impl Spi for $name {
+            fn init(&mut self, resets: &mut RESETS, baudrate: u32) {
+                self.spi_dev.reset_bring_down(resets);
+                self.spi_dev.reset_bring_up(resets);
+
+                self.set_baudrate(baudrate);
+                self.spi_dev
```
anall, line 72: This and the following write to `sspcsr0` clobber the result of `set_baudrate`

rp2040-hal/src/spi.rs
```diff
@@ -1,3 +1,184 @@
...
+                        break;
+                    }
+                }
+
+                // We might not find a prescale value that lowers the clock freq enough, so we leave it at max
+                debug_assert_ne!(prescale, u8::MAX);
+
+                // Find largest post-divide which makes output <= baudrate. Post-divide is
+                // an integer in the range 1 to 256 inclusive.
+                for postdiv_option in 256u32..1 {
```
anall, line 107: `256u32..1` is an empty range, you need to do something like `(1..=256u32).rev()`

anall (commented):
Just adding some notes for your future reference. Those changes + some extra support for additional `embedded_hal` traits got the OLED screen on the test board (Adafruit Macropad) working

rp2040-hal/src/spi.rs
hmvp: Ah yes some debugging code.. Should be fixed now

rp2040-hal/src/spi.rs
hmvp: Fixed

rp2040-hal/src/spi.rs
```diff
@@ -1,3 +1,218 @@
...
+}
+
+impl<D: SpiDevice, const DS: u8> Spi<Enabled, D, DS> {
+    fn is_writable(&self) -> bool {
+        self.device.sspsr.read().tnf().bit_is_set()
+    }
+    fn is_readable(&self) -> bool {
+        self.device.sspsr.read().rne().bit_is_set()
+    }
+    // fn is_busy(&self) -> bool {
```
9names, line 167: Commented code should have a comment to explain why it is commented out

9names (changes requested):
Pretty good port of the C SDK code, just a couple of things I'd like cleaned up.

If you wouldn't mind, could you please not rewrite git history after a review?
I can squash the PR into a single commit on merge if you like, but it's hard to follow review history and review-related fixes when they aren't separate commits on top of your existing work.

rp2040-hal/src/spi.rs
```diff
@@ -1,3 +1,218 @@
...
+    pub fn new(device: D) -> Spi<Disabled, D, DS> {
+        Spi {
+            device,
+            state: PhantomData,
+        }
+    }
+    /// Set baudrate based on system clock
+    ///
+    /// Typically the systemclock is set to 125_000_000
+    fn set_baudrate(&mut self, sys_freqency: u32, baudrate: u32) -> u32 {
```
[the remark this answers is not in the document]
anall: This is based on the peripheral clock, not the system clock. By default, they are the same but do not need to be. On my test board, I'm running the system clock at 125MHz but have the peripheral clock running at 48MHz based on the USB PLL.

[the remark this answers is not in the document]
hmvp: Ah yes I am not entirely happy about that part. I feel the clocks struct should keep track of the desired frequency of all clocks so users can just ask for the right frequency and the code can figure out the dividers. But that needs a lot more work in the clocks section.

For now I will fix the comment and the variable name

rp2040-hal/src/spi.rs
```diff
@@ -1,3 +1,214 @@
...
+    pub fn new(device: D) -> Spi<Disabled, D, DS> {
+        Spi {
+            device,
+            state: PhantomData,
+        }
+    }
+    /// Set baudrate based on peripheral clock
+    ///
+    /// Typically the peripheral clock is set to 125_000_000
+    fn set_baudrate(&mut self, peri_frequency: u32, baudrate: u32) -> u32 {
```
hmvp, line 82: Should I use embedded_time::rate::*Hertz here for the frequency?

anall: That would make this easier to use, at least.

anall:
I'll look at this more after work today, but this works fine in my (Adafruit Macropad) test program with the only changes required being to support the different calling conventions.

I wonder if we want to add the ability to set the frame format too?

9names (approved):

9names:
I think we will want different frame formats, but we can wait until we have a need for them before we implement them.
I'll wait for @anall to do additional testing before merging this.

anall:
No need to hold on my review as this does work on my test board and we can
iterate later.

(Also: using the TI frame format does make the timing a little less gross
when (mis-)using SPI to drive WS2812 LEDS/"Neopixel"s, but that can be a
future improvement)

9names:
Great, merging as-is then.
Thanks y'all!

PR #51, opened 2021-07-03
Extra board support crates

hmvp:


boards/pico/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+extern crate rp2040_hal as hal;
+
+#[cfg(feature = "rt")]
+extern crate cortex_m_rt;
+#[cfg(feature = "rt")]
+pub use cortex_m_rt::entry;
+
+pub use hal::pac;
+
+hal::bsp_pins!(
```
leo60228, line 12: Are these definitions correct? The `feather_rp2040` BSP defines all pins, not just the ones with dedicated functions.
I'm not sure if this is necessary, though.

hmvp: Ah you might be right!

boards/pico/Cargo.toml
```diff
@@ -1,13 +1,15 @@
 [package]
-name = "pico-bsp"
+name = "pico"
```
anall, line 2: `pico` feels a little too generic if this ever ends up on `crates.io` -- maybe `pico-rp2040`?

boards/pico_explorer/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+        name: i2c_sda,
+        aliases: { FunctionI2C: Sda }
+    },
+    Gpio21 {
+        name: i2c_scl,
+        aliases: { FunctionI2C: Scl }
+    },
+    Gpio22 {
+        name: i2c_int,
+        aliases: { FunctionI2C: Int }
```
anall, line 91: The I2C peripheral doesn't support an interrupt. (Gpio22 is even on a different I2C channel than 20/21)

boards/pico/Cargo.toml
leo60228: Maybe `pi-pico`, or `rp-pico` if the idea is to namespace it? `pico-rp2040` feels a bit redundant to me.

boards/pico_explorer/src/lib.rs
hmvp: Well that is interesting. I have the board in front of me and that is how its silk screened on... It also matches with the schematic: https://cdn.shopify.com/s/files/1/0174/1800/files/pico_explorer_schematic.pdf?v=1619077703

The i2C_int pin is used for the breakout garden boards: https://shop.pimoroni.com/collections/pimoroni-breakouts?sort=new I see that some do have an interrupt pin. I assume they want you to add a (G)PIO interrupt on that pin...

hmvp: I removed the Function alias

boards/pico/Cargo.toml
hmvp: Yeah I also feel more for prefixing with manufacturer

9names (approved):
LGTM.
More BSPs are definitely welcome.

PR #52, opened 2021-07-04
Initial ADC impl

hmvp:


rp2040-hal/src/adc.rs
```diff
@@ -1,3 +1,104 @@
...
+    WORD: From<u16>,
+    PIN: Channel<Adc, ID = u8>,
+{
+    type Error = ();
+
+    fn read(&mut self, _pin: &mut PIN) -> nb::Result<WORD, Self::Error> {
+        let chan = PIN::channel();
+
+        if chan == 4 {
+            self.device.cs.modify(|_, w| w.ts_en().set_bit())
```
9names, line 89: You turn on the temperature sensor here. Do you ever turn it off again?

rp2040-hal/src/adc.rs
```diff
@@ -1,3 +1,104 @@
...
+    device: ADC,
+}
+
+impl Adc {
+    /// Create new adc struct and bring up adc
+    pub fn new(device: ADC, resets: &mut RESETS) -> Self {
+        device.reset_bring_down(resets);
+        device.reset_bring_up(resets);
+
+        // Enable adc and temp sensor
```
9names, line 28: This isn't enabling the temp sensor, could you reduce this comment to just "Enable adc"
I know the section in the manual is "ADC and temperature sensor", but it's really two separate components and you're only dealing with one of them here.

rp2040-hal/src/adc.rs
```diff
@@ -1,3 +1,104 @@
...
+        device.cs.write(|w| w.en().set_bit());
+
+        // Wait for adc ready
+        while !device.cs.read().ready().bit_is_set() {
+            cortex_m::asm::nop();
```
9names, line 33: Performing a nop in a spin loop does not accomplish anything.
Unless you're going to put the processor to sleep waiting for interrupt (wfi) you should leave the body of the while empty.

rp2040-hal/src/adc.rs
```diff
@@ -1,3 +1,104 @@
...
+channel!(Gpio29, 3);
+
+/// Internal temperature sensor type
+pub struct TempSense;
+
+impl Channel<Adc> for TempSense {
+    type ID = u8; // ADC channels are identified numerically
+
+    fn channel() -> u8 {
+        4
```
9names, line 74: This should be a constant, so that when it's used later on line 88 it's obvious why channel 4 is a special case.
```const TEMPERATURE_SENSOR_CHANNEL: u8 = 4;```

rp2040-hal/src/adc.rs
hmvp: Yeah that was bothering me also.. Improved it a bit

rp2040-hal/src/adc.rs
hmvp: Fixed

hmvp:
Fixed the review comments

9names (approved):
LGTM

PR #53, opened 2021-07-06
Add 'rt' feature [for the HAL to pass to the PAC]

anall:
For the HAL, currently just passes the feature to the PAC.

Also pass the 'rt' feature through to the HAL for all the BSP crates.

This is needed to make using interrupts easier (feature use is global in Cargo, so if you include the PAC as a dependency in your application with the 'rt' feature, that will pollute the dependency inside the HAL -- but we probably shouldn't rely on that)

And either way, BSPs (should) re-export the HAL (as `bsp::hal`), and the HAL re-exports the PAC (as `hal::pac`) -- so there should be no reason to actually directly depend on the PAC (or the HAL if you are using a BSP)

### DIFF
--- a/boards/feather_rp2040/Cargo.toml
+++ b/boards/feather_rp2040/Cargo.toml
@@ -16,4 +16,4 @@ cortex-m-rt = { version = "0.6.14", optional = true }
 
 [features]
 default = ["rt"]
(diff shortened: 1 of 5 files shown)


9names (approved):
Agreed.
LGTM

PR #54, opened 2021-07-07
Refactor clocks

### LINKED ISSUE #47: Clocks improvements

hmvp:
This makes setting clocks much safer and easier

Still to do for a next change is to give the clocks some type state so you cannot change a clock if there are dependents (or they change automatically as well)

Closes #47

### DIFF
--- a/rp2040-hal/src/clocks/available_clocks.rs
+++ b/rp2040-hal/src/clocks/available_clocks.rs
@@ -1,0 +1,253 @@
+//! Available clocks
+
+use super::{ClocksManager, ShareableClocks};
+use crate::{
+    gpio::{
+        bank0::{Gpio20, Gpio22},
+        FunctionClock, Pin,
+    },
+    pll::{Locked, PhaseLockedLoop},
+    rosc::{Enabled, RingOscillator},
+    typelevel::Sealed,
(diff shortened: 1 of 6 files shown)


9names (approved):
LGTM

PR #56, opened 2021-07-07
I2c

JoNil:
This is the beginning of my i2c implementation, warning i have not tested it yet. Waiting for a pico order.

It is more or less a literal translation of the c code in the rp pico sdk.

TODO:
- [x] Test
- [x] Set the pin mode for i2c
- [x] Get the clock frequency from somewhere

### DIFF
--- a/rp2040-hal/src/i2c.rs
+++ b/rp2040-hal/src/i2c.rs
@@ -67,6 +67,22 @@ impl SclPin<I2C0> for Gpio21 {}
 impl SdaPin<I2C1> for Gpio26 {}
 impl SclPin<I2C1> for Gpio27 {}
 
+/// I2C pin pair
+pub trait I2CPinPair {}
+
+impl I2CPinPair for (Gpio0, Gpio1) {}
+impl I2CPinPair for (Gpio2, Gpio3) {}
+impl I2CPinPair for (Gpio4, Gpio5) {}
+impl I2CPinPair for (Gpio6, Gpio7) {}
+impl I2CPinPair for (Gpio8, Gpio9) {}
+impl I2CPinPair for (Gpio10, Gpio11) {}
+impl I2CPinPair for (Gpio12, Gpio13) {}
+impl I2CPinPair for (Gpio14, Gpio15) {}
+impl I2CPinPair for (Gpio16, Gpio17) {}
+impl I2CPinPair for (Gpio18, Gpio19) {}
+impl I2CPinPair for (Gpio20, Gpio21) {}
+impl I2CPinPair for (Gpio26, Gpio27) {}
+
 /// I2C peripheral operating in master mode
 pub struct I2C<I2C, Pins> {
     i2c: I2C,
@@ -94,6 +110,7 @@ macro_rules! hal {
                     Scl: SclPin<$I2CX>,
                     SdaMode: PinMode + ValidPinMode<Sda>,
                     SclMode: PinMode + ValidPinMode<Scl>,
+                    (Sda, Scl): I2CPinPair,
                 {
                     let sda_pin = sda_pin.into_mode::<FunctionI2C>();
                     let scl_pin = scl_pin.into_mode::<FunctionI2C>();
--- a/rp2040-hal/src/i2c.rs
+++ b/rp2040-hal/src/i2c.rs
@@ -15,14 +15,7 @@ use rp2040_pac::{I2C0, I2C1, RESETS};
 #[non_exhaustive]
 #[derive(Debug)]
 pub enum Error {
-    /// Bus error
-    Bus,
-    /// Arbitration loss
-    Arbitration,
-    // Overrun, // slave mode only
-    // Pec, // SMBUS mode only
-    // Timeout, // SMBUS mode only
-    // Alert, // SMBUS mode only
+    Abort(u32),
 }
 
 #[doc(hidden)]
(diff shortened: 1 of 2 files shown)


rp2040-hal/src/i2c.rs
```diff
@@ -1,3 +1,384 @@
...
+#[doc(hidden)]
+mod sealed {
+
+    use crate::gpio::pin::bank0::{
+        Gpio0, Gpio1, Gpio10, Gpio11, Gpio12, Gpio13, Gpio14, Gpio15, Gpio16, Gpio17, Gpio18,
+        Gpio19, Gpio2, Gpio20, Gpio21, Gpio26, Gpio27, Gpio3, Gpio4, Gpio5, Gpio6, Gpio7, Gpio8,
+        Gpio9,
+    };
+
+    pub trait Sealed {}
```
anall, line 31: We already have `crate::typelevel::Sealed`, which all of these `PinId`s already implement

rp2040-hal/src/i2c.rs
```diff
@@ -1,3 +1,384 @@
...
+    impl Sealed for Gpio27 {}
+}
+
+/// SCL pin
+pub trait SclPin<I2C>: sealed::Sealed {}
+
+/// SDA pin
+pub trait SdaPin<I2C>: sealed::Sealed {}
+
+impl SdaPin<I2C0> for Gpio0 {}
```
anall, line 64: This is a good way to handle this

rp2040-hal/src/i2c.rs
```diff
@@ -1,3 +1,384 @@
...
+fn i2c_reserved_addr(addr: u8) -> bool {
+    (addr & 0x78) == 0 || (addr & 0x78) == 0x78
+}
+
+macro_rules! hal {
+    ($($I2CX:ident: ($i2cX:ident),)+) => {
+        $(
+            impl<SCL, SDA> I2c<$I2CX, (SCL, SDA)> {
+                /// Configures the I2C peripheral to work in master mode
+                pub fn $i2cX<F>(i2c: $I2CX, pins: (SCL, SDA), freq: F) -> Self
```
anall, line 115: `pins` accepts two `PinId`s, not two `Pin`s -- this should be something like `Pin<SCL,FunctionI2C>`, which also handles making sure the pin is in the correct mode for being used as an i2c pin

rp2040-hal/src/i2c.rs
```diff
@@ -1,3 +1,384 @@
...
+                    SCL: SclPin<$I2CX>,
+                    SDA: SdaPin<$I2CX>,
+                {
+                    let freq = freq.into().0;
+                    assert!(freq <= 1_000_000);
+                    assert!(freq > 0);
+
+                    unsafe {
+                        // Reset i2c hardware
+                        let resets = &*RESETS::ptr();
```
anall, line 127: You should be passing this in as &mut RESETS, also see `resets::SubsystemReset` which handles the below bit.

rp2040-hal/src/i2c.rs
```diff
@@ -1,3 +1,384 @@
...
+}
+
+fn i2c_reserved_addr(addr: u8) -> bool {
+    (addr & 0x78) == 0 || (addr & 0x78) == 0x78
+}
+
+macro_rules! hal {
+    ($($I2CX:ident: ($i2cX:ident),)+) => {
+        $(
+            impl<SCL, SDA> I2c<$I2CX, (SCL, SDA)> {
```
anall, line 113: `I2C`?

anall (changes requested):
I see this is a draft, but a few notes.

I'd probably base this off the v2 bits of `atsamd-hal` instead of the stm stuff, as this HAL's GPIO is heavily based on the `atsamd-hal` crate

It seems they haven't yet converted I2C to v2, but here's SPI: https://github.com/atsamd-rs/atsamd/blob/master/hal/src/thumbv7em/sercom/v2/spi.rs

rp2040-hal/src/i2c.rs
anall: Also probably take this as two arguments instead of as a tuple. You can always continue to use a tuple internally.

rp2040-hal/src/i2c.rs
JoNil: Fixed!

rp2040-hal/src/i2c.rs
JoNil: Nice! Fixed :)

rp2040-hal/src/i2c.rs
JoNil: :)

rp2040-hal/src/i2c.rs
JoNil: Nice! Fixed!

rp2040-hal/src/i2c.rs
JoNil: Fixed!

JoNil:
@anall Can you check the type level shenanigans for the i2cX function. 

What i think the code does is that it allows input of pins in any mode as long as they are convertible into FunctionI2C pins then it converts them into FunctionI2C pins and the struct only stores FunctionI2C pins.

In the example pi skd i2c example https://github.com/raspberrypi/pico-examples/blob/146680d625bb2238a84c01dee118351efb5ede4e/i2c/lcd_1602_i2c/lcd_1602_i2c.c#L138
The pins are set into pull up mode. Will sda_pin.into_mode::<FunctionI2C>() also change the pins into pull up mode?

Right now i have my test code here: https://github.com/JoNil/pico_plant/blob/9b4e5ac9c6ac0be6e51838c6f383f38e4657b1fc/src/main.rs#L28

I will add an example to this pr before it is ready :)

anall:
>Can you check the type level shenanigans for the i2cX function.

I'll look deeper tomorrow.

>What i think the code does is that it allows input of pins in any mode as long as they are convertible into FunctionI2C pins then it converts them into FunctionI2C pins and the struct only stores FunctionI2C pins.

It might just be easier to take a Pin<…> already in the mode, and rely on the caller calling `.into_mode()`?

>The pins are set into pull up mode. Will sda_pin.into_mode::() also change the pins into pull up mode?

Not currently. Do we want that always done?


Also re. the Pin Pair constraint. Does the datasheet say these are paired? As far as I can tell you can use any valid SDA for that channel with any other valid SCL -- unless I missed something in the datasheet.

JoNil:
> It might just be easier to take a Pin<…> already in the mode, and rely on the caller calling `.into_mode()`?

I changed it so that the code only accepts pin already in the FunctionI2C mode. How does the user choose if the pin should be a pull-up or not when the pin is in FunctionI2C mode? I can't seem to figure this out from the gpio code.

> Also re. the Pin Pair constraint. Does the datasheet say these are paired? As far as I can tell you can use any valid SDA for that channel with any other valid SCL -- unless I missed something in the datasheet.

You are most likely correct! 😅 I will test this when i get my next batch of picos.

JoNil:
I have now successfully run a 1306 oled i2c display with this code :D

9names (approved):
LGTM

PR #58, opened 2021-07-19
Deny warnings in ci

JoNil:
I have moved the deny warnings error to ci instead of every build.

See https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md for a rationale.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -4,6 +4,8 @@ jobs:
   check:
     name: cargo-check
     runs-on: ubuntu-20.04
+    env:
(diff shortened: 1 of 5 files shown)


9names:
In case anyone wants a concrete example where this already hurts:
If we update to the latest svd2rust, it will change write_with_zero() to be unsafe.
Currently, I cannot add unsafe blocks to the HAL before updating svd2rust, because the [unused_unsafe] lint will cause the build to error out rather than just warn me that the unsafe isn't necessary.

9names (approved):
LGTM

PR #59, opened 2021-07-19
Attempt at simpler interface  for novice embeded developers

JoNil:
When playing around with the adc temp sensor i ran into a hang because i hadn't configured the adc clock.

This is an attempt at a simplified interface to the rp2040 that still allows all the usages of the usual way to write embedded rust.

Do you think this is a good idea? Do you want me to develop this further? Any ideas suggestions of how to make it better?

Blinky before:
```Rust
#![no_std]
#![no_main]

use cortex_m_rt::entry;
use embedded_hal::digital::v2::OutputPin;
use hal::pac;
use hal::sio::Sio;
use panic_halt as _;
use rp2040_hal as hal;

#[link_section = ".boot2"]
#[used]
pub static BOOT2: [u8; 256] = rp2040_boot2::BOOT_LOADER;

#[entry]
fn main() -> ! {
    let mut pac = pac::Peripherals::take().unwrap();

    let sio = Sio::new(pac.SIO);
    let pins = hal::gpio::Pins::new(
        pac.IO_BANK0,
        pac.PADS_BANK0,
        sio.gpio_bank0,
        &mut pac.RESETS,
    );
    let mut led_pin = pins.gpio25.into_push_pull_output();

    loop {
        led_pin.set_high().unwrap();
        cortex_m::asm::delay(500_000);
        led_pin.set_low().unwrap();
        cortex_m::asm::delay(500_000);
    }
}
```

Blinky after:

```Rust
#![no_std]
#![no_main]

use cortex_m_rt::entry;
use embedded_hal::digital::v2::OutputPin;
use hal::Rp2040;
use panic_halt as _;
use rp2040_hal as hal;

#[link_section = ".boot2"]
#[used]
pub static BOOT2: [u8; 256] = rp2040_boot2::BOOT_LOADER;

#[entry]
fn main() -> ! {
    let mut rp = Rp2040::take().unwrap();

    let mut led_pin = rp.pins.gpio25.into_push_pull_output();

    loop {
        led_pin.set_high().unwrap();
        cortex_m::asm::delay(500_000);
        led_pin.set_low().unwrap();
        cortex_m::asm::delay(500_000);
    }
}
```

Here is an example that uses Adc, I2c and Gpio: https://github.com/JoNil/pico_plant/blob/ffe3e6bb8bf07b79a038e826bb354577af9c4df1/src/main.rs

Since this is a prototype i have not documented it so it only builds with the deny warnings removed.

### DIFF
--- a/rp2040-hal/src/lib.rs
+++ b/rp2040-hal/src/lib.rs
@@ -10,6 +10,7 @@ extern crate cortex_m;
 extern crate embedded_hal as hal;
 extern crate nb;
 pub use paste;
+pub use rp2040::Rp2040;
 
 pub extern crate rp2040_pac as pac;
 
@@ -23,6 +24,7 @@ pub mod pwm;
 pub mod resets;
 pub mod rom_data;
 pub mod rosc;
+mod rp2040;
 pub mod rtc;
 pub mod sio;
 pub mod spi;
--- a/rp2040-hal/src/rp2040.rs
+++ b/rp2040-hal/src/rp2040.rs
@@ -1,0 +1,170 @@
+use crate::{
+    clocks::ClocksManager,
+    gpio::Pins,
+    pll::{
+        common_configs::{PLL_SYS_125MHZ, PLL_USB_48MHZ},
+        setup_pll_blocking, Locked, PhaseLockedLoop,
+    },
+    sio::Sio,
+    watchdog::Watchdog,
+    xosc::{setup_xosc_blocking, CrystalOscillator, Stable},
+};
+use embedded_time::rate::Extensions;
+use pac::{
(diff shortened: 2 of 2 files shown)


9names:
This is a nice idea, but it really belongs in a BSP (board support package) since things like external high speed oscillator may be different or missing between boards.
A BSP will also have nice names for pins, so for example you can refer to gpio25 as led.
We already have a minimal BSP for the Pico in this repo:
https://github.com/rp-rs/rp-hal/blob/main/boards/pico

Here's what a blinky example looks like targeting the BSP on the micro:bit
https://github.com/nrf-rs/microbit/blob/main/examples/gpio-hal-blinky/src/main.rs

Here's the high-level factory they're calling to configure the HAL
https://github.com/nrf-rs/microbit/blob/458d70584f3b46e075b4dfc21377a98cb65795da/microbit-common/src/v2/board.rs#L131

JoNil:
Nice i'll look more into the bsp, code :)

PR #60, opened 2021-07-21
Add HD44780 example

Marcuss2:
I think this is a good example showing how to use libraries compatible with embedded-hal.

Make sure to squash commits, I forgot to format.

Also, is there a better way to get the AHL bus speed from the HAL?


9names:
There's currently no way to request the system bus (AHB) frequency from the HAL.
I think it's safe to assume it will be 12MHz on reset and will most often set to 125MHz when the clock infra is used, but it would be great to have a way to interrogate this at runtime to ensure everything is correct.

9names (approved):
LGTM, thanks!

PR #61, opened 2021-07-21
Pwm improvements

hmvp:
I was not entirely satisfied with the way PWM worked so I started to modify it a bit. That turned into a full rewrite. However I am not satisfied yet.

The upside is that this models the actual peripheral better where some settings are shared between channels and multiple pins can be used simultaneous by the same channel.

The downside is that you might want to "start" with the gpio pin and "just" find the right pwm slice that belongs to that. I found no way to do that except for an macro that needs to be used in the same scope as the `Slices` instance..
Another issue is that once you borrow a channel (like in the example just before the loop) the slice is no longer available to change settings. We could probably change that by making all methods take `&self` instead of `&mut self` but I am not sure whats best here...

Any insights are welcome


hmvp:
@anall I took a lot of inspiration from the gpio system. Maybe you have some comments?

anall:
> @anall I took a lot of inspiration from the gpio system. Maybe you have some comments?

I'll try and take a look this weekend.

rp2040-hal/src/pwm/mod.rs
```diff
@@ -0,0 +1,581 @@
+//! Pulse Width Modulation (PWM)
+//!
+//! To access the PWM pins you must call the 'split' method on the PWM. This will return a
```
henkkuli, line 3: This documentation along with the examples needs to be updated.

rp2040-hal/src/pwm/mod.rs
```diff
@@ -0,0 +1,581 @@
...
+pub trait SliceId: Sealed {
+    /// Corresponding [`DynSliceId`](super::DynSliceId)
+    const DYN: DynSliceId;
+    /// [`DivMode`] at reset
+    type Reset;
+}
+
+macro_rules! slice_id {
+    ($Id:ident, $NUM:literal, $reset : ident) => {
+        #[doc = "Slice ID representing slice "]
```
henkkuli, line 156: The documentation could contain the numeric ID.
```suggestion
        #[doc = std::concat!("Slice ID representing slice ", $NUM)]
```

rp2040-hal/src/pwm/mod.rs
```diff
@@ -0,0 +1,581 @@
...
+}
+
+//==============================================================================
+//  Registers
+//==============================================================================
+
+/// Provide a safe register interface for [`Slice`]s
+///
+/// This `struct` takes ownership of a [`SliceId`] and provides an API to
+/// access the corresponding regsiters.
```
henkkuli, line 173:
```suggestion
/// access the corresponding registers.
```

rp2040-hal/src/pwm/mod.rs
```diff
@@ -0,0 +1,581 @@
...
+        Pwm2: (2, [Gpio4, Gpio5, Gpio20, Gpio21], 2),
+        Pwm3: (3, [Gpio6, Gpio7, Gpio22, Gpio23], 3),
+        Pwm4: (4, [Gpio8, Gpio9, Gpio24, Gpio25], 4),
+        Pwm5: (5, [Gpio10, Gpio11, Gpio26, Gpio27], 5),
+        Pwm6: (6, [Gpio12, Gpio13, Gpio28, Gpio29], 6),
+        Pwm7: (7, [Gpio14, Gpio15], 7)
+    ]
+}
+
+///
```
henkkuli, line 311:
```suggestion
/// Marker trait for valid PWM input pins.
```

rp2040-hal/src/pwm/mod.rs
```diff
@@ -0,0 +1,581 @@
...
+pub trait ValidPwmInputPin<S: SliceId>: Sealed {}
+///
```
henkkuli, line 313:
```suggestion
/// Marker trait for valid PWM output pins.
```

rp2040-hal/src/pwm/mod.rs
```diff
@@ -0,0 +1,581 @@
...
+pub trait ValidPwmOutputPin<S: SliceId, C: ChannelId>: Sealed {}
+
+/// Marks this Channel as having output (Both Channels A and B)
+pub trait PwmOutputChannel {
```
henkkuli, line 317: Is this trait used anywhere? How about `PwmInputChannel`?

rp2040-hal/src/pwm/mod.rs
```diff
@@ -0,0 +1,581 @@
...
+        self.regs.enable(true);
+    }
+
+    /// Disable slice
+    #[inline]
+    pub fn disable(&mut self) {
+        self.regs.enable(true)
+    }
+
+    #[doc = "Enables phase correct mode"]
```
henkkuli, line 415: Why using `#[doc = ""]` and not `///`? Also
```suggestion
    #[doc = "Enable phase correct mode"]
```

rp2040-hal/src/pwm/mod.rs
```diff
@@ -0,0 +1,581 @@
...
+        match C::DYN {
+            DynChannelId::A => self.regs.write_cc_a(duty),
+            DynChannelId::B => self.regs.write_cc_b(duty),
+        }
+    }
+}
+
+impl<S: SliceId, M: SliceMode + ValidSliceMode<S>> Channel<S, M, A> {
+    ///
+    pub fn output_to<
```
henkkuli, line 517: This should probably return `PwmPinToken` because otherwise there is no way to recover the pin back from PWM.

rp2040-hal/src/pwm/dynslice.rs
```diff
@@ -0,0 +1,370 @@
...
+}
+
+//==============================================================================
+//  DynRegisters
+//==============================================================================
+
+/// Provide a safe register interface for [`DynSlice`]s
+///
+/// This `struct` takes ownership of a [`DynSliceId`] and provides an API to
+/// access the corresponding regsiters.
```
henkkuli, line 43:
```suggestion
/// access the corresponding registers.
```

rp2040-hal/src/pwm/dynslice.rs
```diff
@@ -0,0 +1,370 @@
...
+    #[inline]
+    #[allow(clippy::wrong_self_convention)]
+    pub fn into_mode(&mut self, mode: DynSliceMode) {
+        if mode != self.mode {
+            self.regs.do_change_mode(mode);
+            self.mode = mode;
+        }
+    }
+
+    /// Enable Slice
```
henkkuli, line 150:
```suggestion
    /// Enable slice
```

henkkuli (commented):
Even though I'm not a maintainer of this crate, I decided to leave a few comments on this PR as I need it. In general I like what this PR is doing and the API matches my expectations.

As a general note, all public methods should have documentation. Now some methods have empty doc string.

rp2040-hal/src/pwm/reg.rs
```diff
@@ -0,0 +1,120 @@
...
+        let num = self.id().num as usize;
+        let ch = unsafe { &(*pac::PWM::ptr()).ch[num] };
+        ch.ctr.write(|w| unsafe { w.ctr().bits(value) });
+    }
+
+    #[inline]
+    fn write_cc_a(&mut self, value: u16) {
+        let num = self.id().num as usize;
+        let ch = unsafe { &(*pac::PWM::ptr()).ch[num] };
+        ch.cc.write(|w| unsafe { w.a().bits(value) });
```
henkkuli, line 86: This overwrites the value written to channel b. Use
```suggestion
        ch.cc.modify(|_, w| unsafe { w.a().bits(value) });
```
instead

hmvp:
> Even though I'm not a maintainer of this crate, I decided to leave a few comments on this PR as I need it. In general I like what this PR is doing and the API matches my expectations.
> 
That's cool, thanks for the comments. I wasn't nearly finished with this so there are some obvious omissions as you pointed out. 
I put it out here in this because I have some doubts about the direction (See first comment) and didn't want to finish something that might be the wrong direction al together.. 

@9names @anall Did you have a chance to look at the direction I took this?

9names:
I had a look, and I like the API you're aiming for better than the previous one.
The original pin-level PWM abstraction really doesn't match the hardware well

I'd would like to know the reason why you want type-erased versions of PWM slices though.
I understand the cost/benefit in doing this for GPIO/pins (though I do not like the cost), but I'm not convinced that there is a *need* for this for PWM slices.
Happy to have this if it solves a problem, I just don't know what that problem is right now.
Also: I'm not nearly confident enough in type-state programming to provide any criticism or help here.

As far as interaction with the PWM peripheral goes:
I'm reading it correctly you "disable" a channel by setting it's duty cycle to 0. A common way to disable a single channel of a PWM pair when there is no channel disable bit is to disconnect the pin from the peripheral, so that the channel output is no longer driving the pin.
This relieves you of the need to store the duty cycle anywhere except the register, and is immediate.

hmvp:
> I had a look, and I like the API you're aiming for better than the previous one.
> The original pin-level PWM abstraction really doesn't match the hardware well
> 
Yeah
> I'd would like to know the reason why you want type-erased versions of PWM slices though.
> I understand the cost/benefit in doing this for GPIO/pins (though I do not like the cost), but I'm not convinced that there is a _need_ for this for PWM slices.
> Happy to have this if it solves a problem, I just don't know what that problem is right now.
> Also: I'm not nearly confident enough in type-state programming to provide any criticism or help here.
> 
To be honest I mostly copied gpio. But I'm still struggling with two issues: 
- One is how to get a pwm slice/channel based on a gpio pin. It seems that would make it more user friendly.
- The other one is the enable multiple slices at the same time. I have not found a way to return multiple slices in some way..
It feels a bit like the Dyn stuff might help with that. Its just that am still looking to see if I can check that at compile time (using types)

> As far as interaction with the PWM peripheral goes:
> I'm reading it correctly you "disable" a channel by setting it's duty cycle to 0. A common way to disable a single channel of a PWM pair when there is no channel disable bit is to disconnect the pin from the peripheral, so that the channel output is no longer driving the pin.
> This relieves you of the need to store the duty cycle anywhere except the register, and is immediate.
That is a better solution, but you need to know which pin is used (and what to set it to). all channels can output to two pins. Even both at the same time... Maybe I should implement the PwmPin trait on the PwmPinToken... But that makes it harder to implement working on both pins...

rp2040-hal/src/pwm/dynslice.rs
```diff
@@ -0,0 +1,370 @@
...
+    #[inline]
+    pub fn enable(&mut self) {
+        self.regs.enable(true);
+    }
+
+    /// Disable slice
+    #[inline]
+    pub fn disable(&mut self) {
+        self.regs.enable(true)
```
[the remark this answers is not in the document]
hmvp: haha yes, I am not even ready to actually test the stuff yet..

hmvp:
I pushed a rebased and heavily changed version. I had more time to test this and am reasonably happy with it.. The two commented functions are still something I feel might need to added. But I cannot get it to work with Rusts constraints...

anall:
I'll try to give this a look later this week. My day job and other commitments have been very demanding recently.

> I'd would like to know the reason why you want type-erased versions of PWM slices though.
> I understand the cost/benefit in doing this for GPIO/pins (though I do not like the cost)

I believe in a lot of cases the optimizer should get rid of the runtime cost (at least for GPIO) if you are not using the type-erased variants. There will obviously be a runtime cost if the type-erased version is used.

But I agree it is probably overkill if it is not needed (I have yet to give this PR a look)

>One is how to get a pwm slice/channel based on a gpio pin. It seems that would make it more user friendly.

I'll have to give this a deeper look, but I feel like there is something clever you can do.

rp2040-hal/src/pwm/mod.rs
9names: Where is the split method? What is a `_____` struct?
I think this first bit of the docs needs little more attention.

rp2040-hal/src/pwm/mod.rs
```diff
@@ -0,0 +1,655 @@
...
+//! ```
+//!
+//! The following configuration options are also available:
+//!
+//! ```no_run
+//! # use rp2040_hal::{prelude::*, pwm::Slices};
+//! # let mut pac = rp2040_pac::Peripherals::take().unwrap();
+//! # let pwm_slices = Slices::new(pac.PWM, &mut pac.RESETS);
+//! # let mut pwm = pwm_slices.pwm4;
+//! pwm.set_ph_correct(); // Run in phase correct mode
```
9names, line 58: I think it would be good to show here how to bring the pwm slice into one of the other modes as well, since it's not obvious.
```rust
/// Set this channel 
let pwm: &mut hal::pwm::Slice<hal::pwm::Pwm4, hal::pwm::InputHighRunning> =
        &mut pwm_slices.pwm4.into_mode();
```

9names (changes requested):
I finally found enough time to test and review this code - sorry about the delay!
The implementation seems quite robust and easy to use, I really like that you can set multiple input/output pins easily.
Happy to merge this once you've improved the docs where requested.

9names (approved):
LGTM

PR #64, opened 2021-07-25
Improve clock frequency stuff for uninitialized clocks and add some examples

hmvp:
Improve clock frequency stuff for uninitialized clocks
    
    - Made clocks singletons so the frequency handling actually works
    - Added initial frequencies
    - Improved the docs
    - Added a Clock trait
    - Add a easier init function
    - Added some Pico BSP examples. Also a XOSC Crystal freq constant. (Could be different for other BSPs)


rp2040-hal/src/clocks/macros.rs
```diff
@@ -107,17 +166,31 @@ macro_rules! clock {
...
+            impl Clock for $name {
+                type Variant = [<$reg:camel SrcType>];
+
+                #[doc = "Get operating frequency for `"$name"`"]
+                fn freq(&self) -> Hertz {
+                    self.frequency
+                }
 
-                /// Configure this clock based on a clock source and desired frequency
-                pub fn configure_clock<S:ClockSource + ValidSrc<$name, Variant=[<$reg:camel SrcType>]>>(&mut self, src: &S, freq: Hertz) -> bool{
```
anall, line 112: Not that you are changing this but this should probably return a `Result<(),ClockError>` or something. Magic return values aren't really that idiomatic in Rust.

rp2040-hal/src/clocks/macros.rs
```diff
@@ -148,7 +221,8 @@ macro_rules! clock {
                     self.set_div(div);
 
                     // Store the configured frequency
-                    self.frequency = src_freq / div;
+                    // div contains both the integer part and the fractional part so we need to shift the src_freq equally
+                    self.frequency = (shifted_src_freq / div as u64).try_into().unwrap();
```
anall, line 225: :+1:

rp2040-hal/src/clocks/macros.rs
```diff
@@ -107,17 +166,31 @@ macro_rules! clock {
...
-                /// Configure this clock based on a clock source and desired frequency
-                pub fn configure_clock<S:ClockSource + ValidSrc<$name, Variant=[<$reg:camel SrcType>]>>(&mut self, src: &S, freq: Hertz) -> bool{
-                    let src_freq: Hertz  = src.get_freq();
+                #[doc = "Configure `"$name"`"]
+                fn configure_clock<S: ValidSrc<$name>>(&mut self, src: &S, freq: Hertz) -> bool{
+                    let src_freq: Hertz<u64> = src.get_freq().into();
 
-                    if freq .gt(& src_freq){
+                    if freq.gt(&src_freq){
                         return false;
```
anall, line 184: See above, this should be something like `Err(ClockError::CantIncrease)` or whatever. Don't have a strong opinion on the name.

rp2040-hal/src/clocks/macros.rs
```diff
@@ -1,3 +1,62 @@
...
+            impl ClocksManager {
+                /// Exchanges CLOCKS block against Self.
+                pub fn new(mut clocks_block: CLOCKS) -> Self {
+                    // Disable resus that may be enabled from previous software
+                    clocks_block.clk_sys_resus_ctrl.write_with_zero(|w| w);
+
+                    let shared_clocks = ShareableClocks::new(&mut clocks_block);
+                    ClocksManager {
+                        clocks: clocks_block,
+                        $(
```
anall, line 35: :+1:

anall (commented):
Was actually about to open an issue about the problems I tripped over trying to update my test program, but you seem to be fixing them here.

Just a few additional comments.

README.md
```diff
@@ -71,6 +71,24 @@ Use this space to show useful examples of how a project can be used. Additional
 For more examples, please refer to the [Documentation](https://github.com/rp-rs/rp-hal)
 
 
+### Run examples
+
+Install [`uf2conv`](https://github.com/sajattack/uf2conv-rs) and [`cargo-binutils`](https://github.com/rust-embedded/cargo-binutils)
+
+```sh
+cargo install uf2conv cargo-binutils
```
9names, line 79: I wrote up a similar script to this yesterday, was considering adding it directly into the repo.
You've missed a step here, which is to run
rustup component add llvm-tools-preview

hmvp: Ah yes will add that

rp2040-hal/src/clocks/macros.rs
hmvp: Done

rp2040-hal/src/clocks/macros.rs
hmvp: Done

9names (approved):
LGTM

PR #66, opened 2021-07-26
Explorer board

hmvp:


9names:
Can you remove the linker script change from this PR?
It will require more discussion and consideration than the board support work.

9names (approved):

PR #69, opened 2021-07-29
Fix unsigned hardware divide/modulo

Rudo2204:
There was a mistake when I copy pasted the code in #40. My apologies.


9names:
I missed it in review too, in spite of looking for things like this

9names (approved):

PR #70, opened 2021-07-30
Simplify uf2 flashing

hmvp:
It might even be a good idea to set the runner in de committed cargo config. But I am not sure if that interferes with debugging using a probe etc..

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -9,3 +9,4 @@ rustflags = [
     "-C", "inline-threshold=5",
     "-C", "no-vectorize-loops",
 ]
+runner = "elf2uf2-rs -d"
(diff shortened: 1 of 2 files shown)


9names:
This is much cleaner than the previous method and my attempt at this.
I think this should be the default runner, as out-of-the box experience matters. Even if you plan to always use a debugger, validating that you can run code using UF2 first is a good idea.
And, someone using a debugger is going to need to add probe/debug software specific config into their config.toml anyway, so requiring them to remove the elf2uf2-rs runner is not a big deal.

hmvp:
Changed!

9names (approved):

PR #73, opened 2021-08-06
Add UART example + core::fmt support to UART

### LINKED ISSUE #62: UART example
Hi,

Can anyone provide example for setting up UART?
I'm having issue that I located (using led) to be in `UartPeripheral::enable` (code probably panics or hangs the cpu)
I located that it cannot be invalid frequency (as I didn't unwrap error) and I didn't found anything obvious inside that could cause that.

Thanks

9names:
HAL UART code did not have an example, and the documentation did not have sample code either.
This PR adds both of these.

Resolves #62


9names:
Added core::fmt support as it was trivial, and it's almost always what end-users want anyway.

henkkuli:
Implementing `core::fmt::Write` directly on `UartPeripheral` is definitely an important quality-of-life improvement. Currently it is implemented only indirectly for `dyn embedded_hal::serial::Write` which makes using it a real pain in the ass:
```rust
use core::fmt::Write;
let uart0: &mut UartPeripheral<Enabled, UART0> = ...;
let uart0: &mut dyn embedded_hal::serial::Write<u8, Error = Infallible> = uart0;
write!(uart0, "Hello world!").unwrap();
```

rp2040-hal/examples/uart.rs
```diff
@@ -0,0 +1,70 @@
...
+        clocks.peripheral_clock.into(),
+    )
+    .unwrap();
+
+    // UART TX (characters sent from pico) on pin 1 (GPIO0) and RX (on pin 2 (GPIO1)
+    let _tx_pin = pins.gpio0.into_mode::<gpio::FunctionUart>();
+    let _rx_pin = pins.gpio1.into_mode::<gpio::FunctionUart>();
+
+    // We need a short delay here otherwise the first few characters are garbled.
+    // TODO: work out why
```
henkkuli, line 60: This is probably due to missing frame start. While the UART channel is not in use, the line should be in marking state, i.e. high, but the above lines turn the line from high impedance state to high state and hence the line might not be in the marking state for long enough for the receiver to correctly detect this. Maybe you could initialize the pin as output before initializing the UART peripheral and see whether that helps?

9names: Ah, of course.
I was observing this while testing by physically detaching and re-attaching the Pico while leaving my USB-UART attached, so it would make sense that TX wasn't in a good state prior to transmission.
It isn't exhibiting this issue when resetting the program via a debugger or by via the RUN pin, so I think it's safe to remove this delay.

PR #76, opened 2021-08-08
Fix doc examples and add checking

### LINKED ISSUE #33: Running tests on CI
After merging PR https://github.com/rp-rs/rp-hal/pull/29 we started running `cargo test` on the CI host
This CI stage started failing on https://github.com/rp-rs/rp-hal/pull/24 with error message: Error: language item required, but not found: `eh_personality`
Link to CI failure https://github.com/rp-rs/rp-hal/pull/24/checks?check_run_id=2498172416

We would like to run tests that exercise embedded_hal traits, but these will not run in CI without providing them.
I could not find any other HAL attempting to run cargo test on the hosts CI, there are security/reliability issues with running HIL(hardware-in-loop) with CI on one of the developers systems, and I don't think linux-embedded-hal is sufficiently advanced that we can use it for this purpose.

My current proposal is that we stop running `cargo test` as part of CI until we can determine how to do so without causing more harm than good.
I've created a PR https://github.com/rp-rs/rp-hal/pull/32 that does that, as well as breaking our current CI workload into multiple workflows to aid discovery

hmvp:
Resolves #33

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -20,3 +20,7 @@ jobs:
         with:
           command: check
           args: --workspace --examples
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --doc --target x86_64-unknown-linux-gnu
--- a/rp2040-hal/src/uart.rs
+++ b/rp2040-hal/src/uart.rs
@@ -5,19 +5,30 @@
 //! ## Usage
 //!
 //! See [examples/uart.rs](https://github.com/rp-rs/rp-hal/tree/main/rp2040-hal/examples/uart.rs) for a more complete example
-//! ```rust
+//! ```no_run
+//! use rp2040_hal::{clocks::init_clocks_and_plls, gpio::{Pins, FunctionUart}, pac, sio::Sio, uart::{self, UartPeripheral}, watchdog::Watchdog};
+//!
+//! let mut peripherals = pac::Peripherals::take().unwrap();
+//! let sio = Sio::new(peripherals.SIO);
+//! let pins = Pins::new(peripherals.IO_BANK0, peripherals.PADS_BANK0, sio.gpio_bank0, &mut peripherals.RESETS);
+//!
+//! let mut watchdog = Watchdog::new(peripherals.WATCHDOG);
+//!
+//! const XOSC_CRYSTAL_FREQ: u32 = 12_000_000; // Typically found in BSP crates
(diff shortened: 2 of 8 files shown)


.github/workflows/check.yml
```diff
@@ -20,3 +20,7 @@ jobs:
         with:
           command: check
           args: --workspace --examples
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --doc --target x86_64-unknown-linux-gnu
```
9names, line 26: Since this is a test workflow, and it's targeting the host, could you add it to
```.github/workflows/tests_host.yml```
instead of
```.github/workflows/check.yml```

File: rp2040-hal/src/uart.rs [hunk 2]
```diff
@@ -5,19 +5,30 @@
...
+//! let sio = Sio::new(peripherals.SIO);
+//! let pins = Pins::new(peripherals.IO_BANK0, peripherals.PADS_BANK0, sio.gpio_bank0, &mut peripherals.RESETS);
+//!
+//! let mut watchdog = Watchdog::new(peripherals.WATCHDOG);
+//!
+//! const XOSC_CRYSTAL_FREQ: u32 = 12_000_000; // Typically found in BSP crates
+//! let mut clocks = init_clocks_and_plls(XOSC_CRYSTAL_FREQ, peripherals.XOSC, peripherals.CLOCKS, peripherals.PLL_SYS, peripherals.PLL_USB, &mut peripherals.RESETS, &mut watchdog).ok().unwrap();
+//!
 //! // Need to perform clock init before using UART or it will freeze.
 //! // Skipping it here for brevity
```
9names, line 21: The comments about not performing clock init can be removed now, since you're clearly doing that init on the lines above

9names:
Thanks for getting the doc tests up and running again, and for fixing up the examples.
Could you add "Resolves #33" somewhere in the description so that issue gets automatically closed when this merges?

.github/workflows/check.yml
hmvp: Are you sure? I changed all examples to no_run since we can only compile check them. So I feel  this is basically a check for them. I interpreted tests_host as relevant for running actual tests on either iron or the ci host (with mocked hardware?)

9names: You are correct, they do logically fit with the checks so let's leave them there

9names (approved):
LGTM

PR #77, opened 2021-08-08
pwm rewrite

abgeana:
I rewrote the pwm module as part of another hobby project. I am submitting my version for review and any potential feedback. I tried to keep the implementation simple while following the original pico-sdk conventions.

I know that there is yet another [pwm rewrite](https://github.com/rp-rs/rp-hal/pull/61) and that is fine. Whichever makes it in the end is fine. I am however genuinely interested in any feedback.

### DIFF
--- a/rp2040-hal/src/pwm.rs
+++ b/rp2040-hal/src/pwm.rs
@@ -1,260 +1,324 @@
 //! Pulse Width Modulation (PWM)
 //!
-//! To access the PWM pins you must call the 'split' method on the PWM. This will return a
-//! `_____` struct with access to each PWM pin:
+//! To access the PWM pins you must instantiate the `Pwm` type, by giving it ownership of the
+//! rp2040_pac::PWM intance retrieved via `rp2040_pac::Peripherals::take()`:
 //!
 //! ```rust
-//! use rp2040_hal::prelude::*;
-//! let mut pac = rp2040_pac::Peripherals::take().unwrap();
-//! let pin_num = 0;
-//! let mut pwm_pin = Pwm0::new(pin_num);
+//! use rp2040_hal:: {
+//!     gpio::{pin::*, Pins},
+//!     pac,
+//!     pwm::{Pwm, PwmChannel, PwmOutput, PwmSlice},
+//!     sio::Sio,
+//! };
+//! let pac = pac::Peripherals::take().unwrap();
+//! let pwm = Pwm::new(pac.PWM, &mut pac.RESETS);
 //! ```
 //!
-//! Once you have the PWM pins struct, you can take individual pins and configure them:
+//! Once you have the `Pwm` instance, you must access individual slices and configure them via
+//! `PwmSlice` instances:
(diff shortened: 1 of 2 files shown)


9names:
Thanks for the PR!
I'll try to get around to reviewing it in the next couple of days.

rp2040-hal/src/pwm.rs
```diff
@@ -1,260 +1,324 @@
...
-    fn ctr(&self) -> &pac::$pwmx::ch::CTR {
-        unsafe {
-            &(*pac::$PWMX::ptr()).ch[$i].ctr
-        }
+    /// Enable the phase correct mode of the PWM slice.
+    #[inline(always)]
+    pub fn enable_phase_correct(&self) {
+        self.pwm.regs.ch[self.idx]
+            .csr
+            .write(|w| w.ph_correct().set_bit());
```
9names, line 129: `write` is not what you use for setting individual bits in a register if you want to preserve the rest of the register fields.
```rust
self.pwm.regs.ch[self.idx].csr.write(|w| w.ph_correct().set_bit());
```
Inside the closure, w starts with the reset value of the register, and then you set the ph_correct bit, and that gets written back to the register (clearing all the other set bits of the register).
You want to use `modify` in cases such as this:
```rust
self.pwm.regs.ch[self.idx].csr.modify(|_,w| w.ph_correct().set_bit());
```
Note how we don't care about the value of r in this case, so it is replaced with `_` in the capture clause and is inaccessible inside the closure body. This lets others  know that it is unused.
You can read more here about this here: https://docs.rust-embedded.org/book/start/registers.html#modifying

rp2040-hal/src/pwm.rs
```diff
@@ -1,260 +1,324 @@
...
-        pad.gpio[self.pin].write(|w| w.ie().set_bit());
-        pad.gpio[self.pin].write(|w| w.od().clear_bit());
-        unsafe {
-            io.gpio[self.pin].gpio_ctrl.write_with_zero(|w| w.funcsel().pwm_a_0());
+    /// Generate a `PwmSlice` instance if the provided slice index is valid. Note that the index is
+    /// 0 based.
+    pub fn slice(&self, idx: usize) -> Option<PwmSlice> {
+        if idx > 7 {
+            // The rp2040 PWM block has only 8 slices indexed from 0 to 7
+            return None;
```
9names, line 98: It is more idiomatic in Rust to return values from each clause of a condition than to use a guard clause like this.
I've flipped the positive condition first, though this is entirely a style choice so keeping as idx > 7 
```rust
pub fn slice(&self, idx: usize) -> Option<PwmSlice> {
    // The rp2040 PWM block has only 8 slices indexed from 0 to 7
    if idx > 7  {
        None
    } else {
        Some(PwmSlice::new(self, idx))
    }
}
```

9names:
Closing this as we merged #61 instead.
Thanks again for the PR, your effort is appreciated.

PR #78, opened 2021-08-08
Correctly configure watchdog tick generation

### LINKED ISSUE #72: Watchdog timer set to garbage
It looks like watchdog is configured incorrectly in `init_clocks_and_plls`
https://github.com/rp-rs/rp-hal/blob/ffa97842e295b19a01fb6438f700c683e9d90ce1/rp2040-hal/src/clocks/mod.rs#L320

in datasheet it is that it should work at 1MHz (and alarms are using the same clock), therefore this should be set here to:
`(xosc_crystal_freq / 1_000_000) as u8`
currently it is trying write very big number (12MHz) to 8bits and as it is truncated it sets it to garbage value

derekw023:
Resolves #72 by applying suggested edits


9names (approved):
LGTM, thanks for the PR

PR #91, opened 2021-08-21
Adafruit feather rp2040

taunusflieger:
Added a blinky example for the Adafruit Feather RP2040 board which uses the flash support added in rp-rs/rp2040-boot2-rs#9 . The example is based on the Pico board example with modifications to utilize the board support crate for the Feather RP2040 board. The Cargo.toml references the specific version rp-rs/rp2040-boot2-rs#9 , this should be changed once we have a new release for rp2040-boot2-rs.


9names (approved):
LGTM, thanks!

PR #93, opened 2021-08-21
Add ADC example and doc-example

### LINKED ISSUE #83: Missing hal example and doc-example for ADC
We have an example of using the ADC as a temperature sensor as part of the pico_explorer_showcase example: https://github.com/rp-rs/rp-hal/blob/main/boards/pico_explorer/examples/pico_explorer_showcase.rs

But we have no standalone example for it in the HAL and no doc-example in adc.rs
https://docs.rs/rp2040-hal/0.2.0/rp2040_hal/adc/index.html

9names:
Closes #83

### DIFF
--- a/rp2040-hal/examples/adc.rs
+++ b/rp2040-hal/examples/adc.rs
@@ -1,0 +1,88 @@
+//! Read ADC samples from the temperature sensor and pin and
+//! output them to the UART on pins 1 and 2 at 9600 baud
+#![no_std]
+#![no_main]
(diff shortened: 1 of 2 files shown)


9names:
Note that I'm not sure the ADC implementation is correct.
Running the example while leaving the input put floating, I observe the temperature sensor and the input pin read the same value. 
I did not dig into this further, but it would be worth following up now that it's easier to observe this issue.

PR #98, opened 2021-09-07
Implement usb_device for rp2040

ithinuel:
This PR depends on https://github.com/rp-rs/rp2040-pac/pull/36 in order to dramatically simplify the driver's implementation. See the PR's description for details.

It has been tested on a pi pico board with the demo application features in this PR.

### Limitations

- RP2040-E5 (Device enumeration workaround) is not implemented
- Suspend/resume is not implemented
- VBus detection is not tested and may not be fully implemented

### DIFF
--- a/rp2040-hal/src/usb.rs
+++ b/rp2040-hal/src/usb.rs
@@ -1,3 +1,530 @@
 //! Universal Serial Bus (USB)
 // See [Chapter 4 Section 1](https://datasheets.raspberrypi.org/rp2040/rp2040_datasheet.pdf) for more details
-// TODO
+//! ## Usage
+//!
+//! Initialize the Usb Bus forcing the VBUS detection.
+//! ```no_run
+//! use usb_device::{class_prelude::*, prelude::*};
+//! use rp2040_hal::usb::UsbBus;
+//! let mut peripherals = pac::Peripherals::take().unwrap();
+//! let usb_bus = UsbBusAllocator::new(UsbBus::new(
+//!         pac.USBCTRL_REGS,
+//!         pac.USBCTRL_DPRAM,
+//!         clocks.usb_clock,
+//!         true,
+//!         &mut pac.RESETS,
+//!     ));
+//! // Use the usb_bus as usual.
+//! ```
+//!
+//! See [pico_usb_serial.rs](https://github.com/rp-rs/rp-hal/tree/main/boards/pico/examples/pico_usb_serial.rs) for more complete examples
+
+use core::cell::RefCell;
+
+use crate::clocks::UsbClock;
+use crate::pac::RESETS;
+use crate::pac::USBCTRL_DPRAM;
+use crate::pac::USBCTRL_REGS;
+use crate::resets::SubsystemReset;
+
+use cortex_m::interrupt::{self, Mutex};
+
+use usb_device::{
+    bus::{PollResult, UsbBus as UsbBusTrait},
+    endpoint::{EndpointAddress, EndpointType},
+    Result as UsbResult, UsbDirection, UsbError,
+};
+
+fn ep_addr_to_ep_buf_ctrl_idx(ep_addr: EndpointAddress) -> usize {
+    ep_addr.index() * 2 + (if ep_addr.is_in() { 0 } else { 1 })
+}
+#[derive(Debug)]
+struct Endpoint {
+    ep_type: EndpointType,
+    max_packet_size: u16,
+    buffer_offset: u16,
+}
+impl Endpoint {
+    unsafe fn get_buf_parts(&self) -> (*mut u8, usize) {
+        const DPRAM_BASE: *mut u8 = USBCTRL_DPRAM::ptr() as *mut u8;
+        if self.ep_type == EndpointType::Control {
+            (DPRAM_BASE.offset(0x100), 64usize)
+        } else {
+            (
+                DPRAM_BASE.offset(0x180 + (self.buffer_offset * 64) as isize),
+                self.max_packet_size as usize,
+            )
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


rp2040-hal/src/usb.rs
```diff
@@ -1,3 +1,528 @@
...
+
+            let buff_status = inner.ctrl_reg.buff_status.read().bits();
+            if buff_status != 0 {
+                // IN Complete shall only be reported once.
+                inner
+                    .ctrl_reg
+                    .buff_status
+                    .write(|w| unsafe { w.bits(0x5555_5555) });
+
+                for i in 0..32u32 {
```
9names, line 500: maybe 2 separate ranges here, 1 that does ep_in_complete and the other that does ep_out.
have them step 2.
I don't think it will produce better assembly but I think it reads a little easier.
```rust
for i in (0..32u32).step_by(2) {
  let m = 1 << i;
  ep_in_complete |= ((buff_status & m) >> (i / 2)) as u16;
}
for i in (1..32u32).step_by(2) {
  let m = 1 << i;
  ep_out |= ((buff_status & m) >> (i / 2 + 1)) as u16;
}
````

**Review by 9names — CHANGES_REQUESTED** [review]
This looks pretty good!
Definitely want that errata workaround implemented ASAP (not necessarily in this PR), took me a while to realize the cause of my enumeration errors as I usually plug everything through USB hubs.
Fix up the comments, pull in the latest upstream PAC and get CI happy and this is okay by me.

File: rp2040-hal/src/usb.rs (hunk 1 above)
Comment (ithinuel) [reply rp2040-hal/src/usb.rs:500]: Because this code is run very often and do not actually contribute much to the usb stack processing I wanted to make it run as fast as possible (and avoid multiple loops).
I agree that it is convoluted and I don't really like it either.

I tried to compare your suggestion with the existing solution and a few alternative.

```rust
// Alternative 1:
while buff_status != 0 {
    let first_bit = buff_status.trailing_zeros();
    
    let u32_mask = 1 << first_bit;
    let u16_mask = 1 << (first_bit / 2);
    
    // clear the flag
    buff_status ^= u32_mask;
    // assign to the appropriate direction.
    if (first_bit & 1) == 0 {
        ep_in_complete |= u16_mask;
    } else {
        ep_out |= u16_mask;
    }
}
```
and
```rust
// Alternative 2
for i in 0..32u32 {
    let mask = 1 << i;
    if (buff_status & mask) == mask {
        // Even index are IN endpoints.
        if (i & 1) == 0 {
            ep_in_complete |= 1 << (i / 2);
        } else {
            ep_out |= 1 << (i / 2);
        }
    }
}
```

Here's what `cargo size --release --example pico_usb_serial` reports:

| snippet | size (bytes) | comment |
| --- | --- | --- |
| current | 29248 | hardly readable
| 9names' suggestion |  29168 | enhance readability but reusing the convoluted bit shifting/masking or the previous solution. |
| alternative 1 | 29168 | Using a ctz that do not exist as an instruction on cm0+. |
| alternative 2 | 29088 | Seems to be the best solution, easier to read and also smaller! |

Do you think the Alternative 2 would do?

9names: alt2 does read better.
Is it functionally equivalent to the original code though, or was the +1 in the ep_out shift a bug in the original?

It's hard for anyone else to determine if you fixed or introduced a bug in the fix, as the commit containing the "current" version has been replaced - please just add more commits to your PR once a review has started, it makes it easier to determine what you've changed.
We can squash them when merging to remove fix-up commits.
Here's a compiler explorer version of the above versions + the original version (from my archives) for anyone else that wants to review the code + the performance
https://rust.godbolt.org/z/qMKxdTfWK

current and alt2 produce very similar code, alt1 produces good code for thumbv7 but terrible for thumbv6, my suggestion was significantly worse than all the others.
So alt2 (assuming it's still correct) is a good option.

anall: Have not looked at the context here, but would using the ROM `ctz` be an option here? https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/src/rom_data.rs#L65

9names: I don't think we have enough of a performance problem here to justify making this more complex.

Even if we did: In the worst-case (every endpoint needs handling) 32 calls to the ROM ctz (11 or 10 cycles each depending on ROM version) is going to be slower than 32 calls to check if the current bit is zero.

If you were looking for quick gains in the common case, I think using clz to set the end of the range would be better
since the number of endpoints used by most people will be very low.
It does remove the fixed loop size though, which the optimizer can use to make the loop faster, so you are punishing the worst case if you do this

ithinuel: >  Is it functionally equivalent to the original code though, or was the +1 in the ep_out shift a bug in the original?

It is functionally equivalent. The +1 was not a bug either. The goal is to map the bits from the position `i` in `buff_status` to the position `i/2` in `ep_out`, so the original version achieved that by moving the bit to the left by `n+1` (regardless if this bit was 0 or 1).
In alt2 we checked if that bit was one so we directly shift 1 to the right by `i/2` and or it into `ep_out`.

> It's hard for anyone else to determine if you fixed or introduced a bug in the fix, as the commit containing the "current" version has been replaced

True, sorry about that. I got used to Gitlab that handles that better.

> would using the ROM `ctz` be an option here?

I had a glance at it but AFAIU the lookup function would have to be every time and I haven't looked at what's in there.

9names: My tired brain missed that it was shifted from the mask position, and that's the source of the +1. makes sense.
This is much better though so I'm going to mark this as resolved

9names (approved):
Looks good to me

PR #103, opened 2021-09-09
Fix adc giving wrong values on sequential reads of multiple channels

JoNil:
Wait for the adc to complete the read before reading the resault


9names (approved):
This works much better, thanks!

PR #104, opened 2021-09-10
Usb fixes

9names:
The check (is_ep0 && (max_packet_size != 64)) doesn't seem to be necessary, and was blocking an example I ported (#105) from working.
The validate buffer size part was upsetting clippy (two paths with the same code) so I simplified it into a single boolean expression - and also added the text from the datasheet explaining what the buffer lengths should be.

### DIFF
--- a/rp2040-hal/src/usb.rs
+++ b/rp2040-hal/src/usb.rs
@@ -149,10 +149,12 @@ impl Inner {
             return Err(UsbError::InvalidEndpoint);
         }
 
-        // validate buffer size
-        if let (EndpointType::Isochronous, true) = (ep_type, max_packet_size > 1023) {
-            return Err(UsbError::Unsupported);
-        } else if max_packet_size > 64 {
+        // Validate buffer size. From datasheet (4.1.2.5):
+        // Data Buffers are typically 64 bytes long as this is the max normal packet size for most FS packets.
+        // For Isochronous endpoints a maximum buffer size of 1023 bytes is supported.
+        // For other packet types the maximum size is 64 bytes per buffer.
+        if (ep_type != EndpointType::Isochronous && max_packet_size > 64) || max_packet_size > 1023
+        {
             return Err(UsbError::Unsupported);
         }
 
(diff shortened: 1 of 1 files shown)


camrbuss:
@9names @ithinuel Thank you so much for the work on the USB peripheral!

I was banging my head trying to get @ithinuel fork working with a hid device, but after #98 went in and I saw this PR open, the fix is working great.

I've got [Keyberon](https://github.com/TeXitoi/keyberon/) working with a custom keyboard and the rp2040!

9names:
@camrbuss Would you mind sharing a link to your port?
I'm currently moving the interrupt flag handling code into the peripheral driver, would be nice to have another codebase to test against.

camrbuss:
@9names I am in a pretty [bad place right now](https://github.com/camrbuss/pinci/blob/9de93e789a39b6c66b7882d94f1ac25103011ac6/firmware/src/main.rs) with just a timer read in a loop. I will definitely move to an interrupt tomorrow, but might go all the way to RTIC at that point.

9names:
Good - part of what I'm trying to preserve is "timer read in a loop" working correctly.
The whole thing works really well with interrupts so I want to make that easy, but I don't want to do it at the cost of manually calling poll without them.

9names:
Since it's working for you Cam I think I'll merge this as-is, and make a separate PR for the interrupt changes.

ithinuel:
Hi guys, what host platform are you testing this with?

That check was kind of a workaround for a limitation in [usb-device](https://github.com/mvirkkunen/usb-device) (see this https://github.com/mvirkkunen/usb-device/pull/50).

9names:
I tested on Linux, will test it out on Windows right now.

9names:
Yep, didn't work on Windows when it works fine on Linux.

Okay, we'll need to work out some way of handling this - I think the existing way wasn't clear enough about what to do about the problem.
I think it's related to https://github.com/hathach/tinyusb/issues/184
Will discuss this in the matrix room

PR #106, opened 2021-09-11
Handle usb interrupt flags in driver

9names:
Move enabling of buff_status, buff_reset and setup_req into UsbBus::enable() and clear the flags from within the UsbBus::poll() so that user code doesn't need to touch the registers.
This way you can either manually poll from within a forever loop, or you can unmask the interrupt and call poll from within the USBCTRL_IRQ.
I've updated the interrupt examples to use the latter, as a reference for how much simpler this is.

I also moved the setup_req portion of the poll handler higher up so that will happen even in the presence of a reset request. This matches the behaviour of the driver in the C USB library that the official SDK uses.

### DIFF
--- a/rp2040-hal/src/usb.rs
+++ b/rp2040-hal/src/usb.rs
@@ -407,6 +407,17 @@ impl UsbBusTrait for UsbBus {
             // at this stage ep's are expected to be in their reset state
             // TODO: is it worth having a debug_assert for that here?
 
+            // Enable interrupt generation when a buffer is done, when the bus is reset,
+            // and when a setup packet is received
+            // this should be sufficient for device mode, will need more for host.
+            inner.ctrl_reg.inte.modify(|_, w| {
+                w.buff_status()
+                    .set_bit()
+                    .bus_reset()
+                    .set_bit()
+                    .setup_req()
+                    .set_bit()
+            });
             // enable pull up to let the host know we exist.
             inner
                 .ctrl_reg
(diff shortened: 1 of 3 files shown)


rp2040-hal/src/usb.rs
```diff
@@ -525,11 +544,13 @@ impl UsbBusTrait for UsbBus {
...
-            // Only report setup if OUT has been cleared.
-            if sie_status.setup_rec().bit_is_set() {
-                ep_setup |= 1;
-                inner.read_setup = true;
+            if sie_status_flags.bus_reset().bit_is_set() {
+                // Clear interrupt flag for reset
+                inner
+                    .ctrl_reg
+                    .sie_status
+                    .modify(|_, w| w.bus_reset().set_bit());
```
ithinuel, line 552: This bit is cleared line [433](https://github.com/rp-rs/rp-hal/blob/af862424633107515fc3ac5f2f044474999f7511/rp2040-hal/src/usb.rs#L433)

rp2040-hal/src/usb.rs
```diff
@@ -498,14 +509,22 @@ impl UsbBusTrait for UsbBus {
...
             let (mut ep_out, mut ep_in_complete, mut ep_setup): (u16, u16, u16) = (0, 0, 0);
 
+            // check for setup request
+            // Only report setup if OUT has been cleared.
+            if sie_status_flags.setup_rec().bit_is_set() {
+                // Clear the setup_rec interrupt flag
+                inner
+                    .ctrl_reg
+                    .sie_status
+                    .modify(|_, w| w.setup_rec().set_bit());
```
ithinuel, line 522: This bit is cleared line [286](https://github.com/rp-rs/rp-hal/blob/af862424633107515fc3ac5f2f044474999f7511/rp2040-hal/src/usb.rs#L286).
Also it must not be cleared until it has been treated (see [here](https://docs.rs/usb-device/0.2.8/usb_device/bus/enum.PollResult.html#variant.Data.field.ep_setup))

ithinuel (changes requested):
The check for reset should not be moved as there is no point in parsing the status if everything's going to be reset anyway.

Note that when Suspend/Resume support will be implemented, they'll also need to take place before that loop (and after the check for reset).

rp2040-hal/src/usb.rs
9names: I'll revert this section to how it was previously

rp2040-hal/src/usb.rs
9names: I'll revert this section to how it was previously

ithinuel (approved):
I haven't tested it but code wise this LGTM

9names:
I tested all the examples we have - it shouldn't break anyone else unless they unmask the interrupt and don't call usb::poll so I think we're safe.

PR #107, opened 2021-09-11
Update pac so we can use rt 0.7

JoNil:


9names:
I've pushed a new PAC release, could you update to
rp2040-pac = "0.1.5"
instead of the current git version?

Also, could you update the BSP crates that use 0.6.14 to cortex-m-rt 0.7 as well?
Thanks!

JoNil:
Done!

9names (approved):
LGTM

PR #109, opened 2021-09-14
Add Pico RTIC Timer Blinky Example

### LINKED ISSUE #108: Add RTIC Example
I have started trying to use [RTIC](https://github.com/rtic-rs/cortex-m-rtic) v0.6.x with this hal with limited success. It would be great to have an example using RTIC with an external interrupt.

If someone has already done this in another project I would be happy to write the example, but my current work is failing.

camrbuss:
Created RTIC example that blinks gpio25 at an interval of 1 second from a interrupt with the RTIC framework. Uses local and shared data from RTIC with pac access through the RTIC Mutex.

I couldn't get RTIC 0.5.x working as the `Nr` trait was needed where as 0.6.x uses `InterruptNumber`

Happy to fix any suggestions!

Resolves #108

### DIFF
--- a/boards/pico/Cargo.toml
+++ b/boards/pico/Cargo.toml
@@ -13,6 +13,7 @@ license = "MIT OR Apache-2.0"
 cortex-m = "0.7.2"
 rp2040-hal = { path = "../../rp2040-hal", version = "0.2.0"}
 cortex-m-rt = { version = "0.6.14", optional = true }
+cortex-m-rtic = "0.6.0-alpha.5"
 embedded-time = "0.12.0"
 usb-device= "0.2.8"
 usbd-serial = "0.1.1"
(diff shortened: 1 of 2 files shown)


9names:
We should really have used the rt version span to allow either rt 0.6 or 0.7 and let the user choose.
https://github.com/embassy-rs/rp2040-pac2/blob/9ad7223a48a065e612bc7dc7be5bf5bd0b41cfc4/Cargo.toml#L8
Not really important here - targetting rtic 0.6.0 seems like a much better option, since the only thing holding back on general release of that is better documentation.

rtic should only be a dev-dependency, we don't it to build for all users of the HAL.

9names (approved):
This is good initial rtic example, thanks!
I think there's room for use of more abstractions - once we have alarms in the timer driver, for example, and pico::Pins::led since you're in the pico bsp examples.
Not worth holding up merging for those, so I'm going to do that right now.

Issue #100, opened 2021-09-09
Timer module support


9names:
We haven't implemented the alarms at as all, and the current counter read implementation is not thread safe (as it uses the latched registers, which could race).
It needs to be changed to use the raw read registers and check for overflow like the [C SDK](https://github.com/raspberrypi/pico-sdk/blob/bfcbefafc5d2a210551a4d9d80b4303d4ae0adf7/src/rp2_common/hardware_timer/timer.c#L38) and  [embassy-rp](https://github.com/embassy-rs/embassy/blob/9082e5f6c5982e7d220f54c8e64e0dd27c6c579d/embassy-rp/src/timer.rs#L36)

9names:
Counter read is now thread safe, still need to implement handling of alarm generation.

PR #110, opened 2021-09-14
Add embedded_hal::timer support

ithinuel:
This PR implements support for the [`embedded_hal::timer`](https://docs.rs/embedded-hal/0.2.6/embedded_hal/timer/index.html) traits.

### DIFF
--- a/boards/pico/Cargo.toml
+++ b/boards/pico/Cargo.toml
@@ -23,6 +23,7 @@ panic-halt= "0.2.0"
 embedded-hal ="0.2.5"
 cortex-m-rtic = "0.6.0-alpha.5"
 rp2040-boot2 = "0.1.2"
+nb = "1.0"
 
 [features]
 default = ["rt"]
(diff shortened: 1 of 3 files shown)


9names:
Hey, thanks for this. If it's not to much could I get you to write a standalone example that uses these traits, so it's obvious for new users how the timer should be used?
Also, not sure if you saw #100 but the current implementation of get_counter could also use some small modifications.

ithinuel:
I updated the `get_counter` function and added an example for the pico board.

I also have https://github.com/ithinuel/rp-hal/tree/add-pro-micro-ws2812-example that adds an example for the pro micro rp2040.
This example combines PIO & Timer (as CountDown trait). (also happy to move https://github.com/ithinuel/ws2812-pio-rs/ under this org)

9names (approved):
LGTM

9names:
Not sure what we should do with regards to PIO driver crates just yet.
Might be worth discussing it in the matrix channel?

PR #111, opened 2021-09-14
Fixed u8 overflow in spi with large postdiv

JoNil:
Fixed bug where a postdiv of 256 would not be represented as a u8 resulting in a panic on subtraction underflow

### DIFF
--- a/rp2040-hal/src/spi.rs
+++ b/rp2040-hal/src/spi.rs
@@ -112,9 +112,9 @@ impl<D: SpiDevice, const DS: u8> Spi<Disabled, D, DS> {
 
         // Find largest post-divide which makes output <= baudrate. Post-divide is
         // an integer in the range 1 to 256 inclusive.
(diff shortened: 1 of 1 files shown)


9names (commented):
If it's only ever going to be effectively 0-255 anyway, could we keep it as u8 instead?
How do you feel about
```rust
for postdiv_option in (0..=255u32).rev() {
  if freq_in / (prescale as u32 * postdiv_option) > baudrate {
// snip
.modify(|_, w| unsafe { w.scr().bits(postdiv) });
// snip
(freq_in / (prescale as u32 * (1 + postdiv as u32))).Hz()
```

JoNil:
Your solution looks good to me too!

JoNil:
Updated after your suggestion, i also made the loop counter u8 so we get a warning if it is changed to be out of range

9names (approved):
LGTM

PR #116, opened 2021-09-15
Made spi set_baudrate public

JoNil:
When using the library embedded-sdmmc i needed a way to reset the clock of the spi peripheral without having ownership of the spi device. This is the way i came up with.

Note: The diff looks kind of weird the new function has not moved but the set_baudrate and set_format functions has moved to be available internally both in enabled and disabled state.

Example usage code can be found here:
https://github.com/JoNil/pico_plant/blob/fc872a3b145f54c2835442413a1a1673e1803d01/src/main.rs#L208

### DIFF
--- a/rp2040-hal/src/spi.rs
+++ b/rp2040-hal/src/spi.rs
@@ -72,20 +72,11 @@ impl<S: State, D: SpiDevice, const DS: u8> Spi<S, D, DS> {
     pub fn free(self) -> D {
         self.device
     }
-}
 
-impl<D: SpiDevice, const DS: u8> Spi<Disabled, D, DS> {
-    /// Create new spi device
-    pub fn new(device: D) -> Spi<Disabled, D, DS> {
-        Spi {
-            device,
-            state: PhantomData,
-        }
-    }
     /// Set baudrate based on peripheral clock
     ///
     /// Typically the peripheral clock is set to 125_000_000
-    fn set_baudrate<F: Into<Hertz<u32>>, B: Into<Hertz<u32>>>(
+    pub fn set_baudrate<F: Into<Hertz<u32>>, B: Into<Hertz<u32>>>(
         &mut self,
         peri_frequency: F,
         baudrate: B,
@@ -128,6 +119,16 @@ impl<D: SpiDevice, const DS: u8> Spi<Disabled, D, DS> {
 
         // Return the frequency we were able to achieve
         (freq_in / (prescale as u32 * (1 + postdiv as u32))).Hz()
+    }
+}
+
+impl<D: SpiDevice, const DS: u8> Spi<Disabled, D, DS> {
+    /// Create new spi device
+    pub fn new(device: D) -> Spi<Disabled, D, DS> {
+        Spi {
+            device,
+            state: PhantomData,
+        }
     }
 
     /// Set format and datasize


ithinuel:
Your pico_plant repo may be private it 404's me.

9names:
I'm very curious about why resetting the clock is useful for this, looking forward to seeing your use-case

JoNil:
Now it is public!

This is my first time talking to sd cards so I'm by no means an expert. Form what i understand sd card only talk spi if you ask them in a specific way. And this involves sending commands at a low spi clock, Then after you have inited the card to use spi then you can increase the spi clock to get better performance.

From the embedded-sdmmc-rs init function docs:
/// This routine must be performed with an SPI clock speed of around 100 - 400 kHz.
/// Afterwards you may increase the SPI clock speed.

Then they have a spi function to get access to the underlying spi device with this interface:
/// Get a temporary borrow on the underlying SPI device. Useful if you
/// need to re-clock the SPI after performing `init()`.
pub fn spi(&mut self) -> core::cell::RefMut<SPI> ...

ithinuel:
Yes, speed negotiation is rather common on SPI/I²C and serial (for the most common ones).
I wonder if a full re-init is necessary or if a limited `set_frequency` would be enough?
Reinit kind of implies reset (bus state, clearing fifo, reset dma settings) to me which seems to be what you implemented but it's probably not necessary, simply changing the clock divider should do the trick.

9names:
You could also keep a handle to the spi bus using https://crates.io/crates/shared-bus, which would ensure your reclocking code has exclusive access at the time it changed the frequency

9names:
oh, sdmmc already has a method for accessing the underlying spi interface.
https://docs.rs/embedded-sdmmc/0.3.0/embedded_sdmmc/struct.SdMmcSpi.html#method.spi

9names:
We could make spi::set_baudrate() public, then you could use that.
Would that work for you JoNil?

JoNil:
That would work for me! Is it ok to change that register wile the hardware is enabled?

9names:
I think it should be fine even when there's data being transmitted, thought I think for now we should recommend that you only change that register when there is no data in flight.

JoNil:
I now updated this pr to make the set_baudrate method public

ithinuel (approved):
LGTM

> Is it ok to change that register wile the hardware is enabled?

Usually, when disabling a peripheral, it finishes any on-going symbol transmission before actually being disabled.
I'd be nice to check that with raspberrypi if this the case for this peripheral (I couldn't find the info in the doc).
Wrapping the configuration (line 113-119 in the new revision of the file)  with disable/enable could be enough.

```rust
self.device.sspcr1.modify(|_,w| w.sse().clear_bit());
while self.device.sspcr1.read().sse().bit_is_set() { cortex_m::asm::nop(); }
self.device.sspcpsr.write(|w| unsafe { w.cpsdvsr().bits(prescaler) });
self.device.sspcr0.modify(|_,w| unsafe { w.scr().bits(postdiv) });
self.device.sspcr1.modify(|_,w| w.sse().set_bit());
```

9names (approved):
PR looks good to me too.
Looking at the prescaler update sequence there's going to be a small window where bus speed may end up being unintentionally faster. Not an issue when not transmitting (current use case) and we could work around it if it turns out to be a problem.
I'll create an issue with this discussion so we remember to look into it.

PR #118, opened 2021-09-16
Add Pro Micro rainbow example (uses PIO & Timer's Countdown)

ithinuel:
Everything is in the title.

It has a dependency on https://github.com/ithinuel/ws2812-pio-rs.


9names (approved):
LGTM

PR #119, opened 2021-09-18
Work around enumeration issues with small ep0 max packet size.

ithinuel:
Prevents `DataIn` of the previous request from being sent to the host when a new setup is received.


9names (approved):
Looks good to me.

PR #120, opened 2021-09-18
Implements USB enumeration workaround (RP2040-E5).

ithinuel:
This is PR presents a first attempt at implementing RP2040-E5.
It still lacks a way to force the init of the timer. Any suggestion is welcome.

I don't have a way to replicate the issue so I can't check if the implementation is correct.
Anyone able to replicate the issue is invited to test it and let me know of the result :)

### DIFF
--- a/rp2040-hal/src/usb.rs
+++ b/rp2040-hal/src/usb.rs
@@ -10,7 +10,6 @@
 //! const XOSC_CRYSTAL_FREQ: u32 = 12_000_000; // Typically found in BSP crates
 //!
 //! let mut pac = pac::Peripherals::take().unwrap();
-//! let sio = Sio::new(pac.SIO);
 //! let mut watchdog = Watchdog::new(pac.WATCHDOG);
 //! let mut clocks = init_clocks_and_plls(
 //!     XOSC_CRYSTAL_FREQ,
@@ -39,7 +38,7 @@
 //!
 //! During enumeration Windows hosts send a `StatusOut` after the `DataIn` packet of the first
 //! `Get Descriptor` resquest even if the `DataIn` isn't completed (typically when the `max_packet_size_ep0`
-//! is less than 18bytes). The next request request is a `Set Address` that expect a `StatusIn`.
+//! is less than 18bytes). The next request is a `Set Address` that expect a `StatusIn`.
 //!
 //! The issue is that by the time the previous `DataIn` packet is acknoledged and the `StatusOut`
 //! followed by `Setup` are received, the usb stack may have already prepared the next `DataIn` payload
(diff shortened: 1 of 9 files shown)


rp2040-hal/src/usb.rs
```diff
@@ -50,6 +50,85 @@ use usb_device::{
...
+
+    // Switch back to USB phy
+    pac.USBCTRL_REGS
+        .usb_muxing
+        .write(|w| w.to_phy().set_bit().softcon().set_bit());
+
+    // Get rid of DP pullup override
+    pac.USBCTRL_REGS
+        .usbphy_direct_override
+        .write(|w| w.dp_pullup_en_override_en().set_bit());
```
9names, line 81: Shouldn't this be clear_bit? Also using modify() instead of write(), since the SDK code reference used hw_clear_alias

rp2040-hal/src/usb.rs
```diff
@@ -50,6 +50,85 @@ use usb_device::{
...
+
+    // J state is a differential 1 for a full speed device so
+    // DP = 1 and DM = 0. Don't actually need to set DM low as it
+    // is already gated assuming it isn't funcseld.
+    io.gpio_ctrl.modify(|_, w| w.inover().high());
+
+    // Force PHY pull up to stay before switching away from the phy
+    pac.USBCTRL_REGS
+        .usbphy_direct
+        .write(|w| w.dp_pullup_en().set_bit());
```
9names, line 114: The SDK code uses hw_set_alias here, so you should be preserving the other bits in the register by calling modify() instead of write()

rp2040-hal/src/usb.rs
```diff
@@ -50,6 +50,85 @@ use usb_device::{
...
+    pac.USBCTRL_REGS
+        .usbphy_direct_override
+        .write(|w| w.dp_pullup_en_override_en().set_bit());
```
9names, line 117: The SDK code uses hw_set_alias here, so you should be preserving the other bits in the register by calling modify() instead of write()

9names (commented):
Found a couple of discrepancies between the C SDK code and yours.
It's still not working correctly with these changes applied.

rp2040-hal/src/usb.rs
ithinuel: Ah, that's what I missed! While writing is I found odd that this bit was set both in start and finish.

rp2040-hal/src/usb.rs
ithinuel: good catch!

rp2040-hal/src/usb.rs
```diff
@@ -54,6 +53,56 @@
...
+//!
+//! ## Issue on RP2040B0 and RP2040B1: USB device fails to exit RESET state on busy USB bus.
+//!
+//! The feature `rp2040-e5`implements the workaround described by [RP2040-E5](https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf#%5B%7B%22num%22%3A630%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C115%2C158.848%2Cnull%5D).
+//!
+//! The workaround requires the GPIO block to be released from its reset and has for side effect
+//! that GPIO15 will be stolen for a few hundred microseconds each time a Reset is detected on the
+//! USB bus.
+//!
+//! Although the external state of the pin will not be affected (the pin it put in "bus keep"), a
```
jannic, line 65: I don't think this is exactly true: If the pin was configured as an active output, the external state will switch from "actively driven" to "weak pull-up/down".

ithinuel: TBH, I'm don't know enough about HW to easily check that.
I tried to find a reference to "bus keep" in the datasheet in vain (I saw that feature on FPGA design but I'm not sure about what it does).

I based that comment on the reference implementation in the [pico-sdk](https://github.com/raspberrypi/pico-sdk/blob/afc10f3599c27147a6f34781b7102d86f58aa5f6/src/rp2_common/pico_fix/rp2040_usb_device_enumeration/rp2040_usb_device_enumeration.c#L97-L98).

jannic: Yes, that matches my understanding. There is surprisingly little documentation, it seems like the bus-keep feature is not mentioned in the datasheet at all.
The [SDK](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_gpio/include/hardware/gpio.h#L203-L204) contains a comment "On the RP2040, setting both pulls enables a "bus keep" function, i.e. a weak pull to whatever is current high/low state of GPIO.", that's all the documentation I can find. But still, a "weak pull" is different from an actively driven output.

ithinuel: A thing I don't know is if the bus is currently high and something comes to pull it low.
Is the pin still driven high by the bus keep or does it switch?

I'll try (I hope there won't be magic smoke)

jannic: It should switch. The general idea of such a bus keep feature is described here: https://en.wikipedia.org/wiki/Bus-holder
(The actual implementation on the rp2040 might be different, but the result should be the same.)

ithinuel: It does indeed :) so the comment seems to be correct :D

jannic: Only as long as nothing is connected to the pin - which seems like a rather uninteresting situation.
But for example, if you have  an external pull-up resistor, it is probably much stronger than the bus-keep mechanism. So if the pin was configured as an active output with level low, it will temporarily switch to high during the e5-workaround.

I'd prefer a wording like:
"The pin will be temporarily put in "bus keep" mode, weakly pulling the
output towards its current logic level. In absence of external loads,
the current logic level will be maintained.
A user will lose control of the pin's output and reading from it may
not reflect the actual state of the external pin."

jannic:
I still can't get it work with the pico_usb_twitchy_mouse example. Does that one work for you?

ithinuel:
@jannic I added the same feature (`rp2040-e5`) to the rp-pico board crate.
You need to build with the feature enabled for the work around to work there (this makes the pico_usb_twitchy_mouse take the GPIO BANK0 out of reset).

`cargo run --example pico_usb_twitchy_mouse --features="rp2040-e5"` should do the trick.

jannic:
Yes, that was what I did wrong: I used `--features rp2040-hal/rp2040-e5` and not `--features rp2040-e5`, so the feature was not enabled for the board crate itself, and the GPIO un-reset wasn't activated.

Now it works.

jannic (approved):

PR #121, opened 2021-09-19
Update changelog and bump verson for hal release 0.3.0

9names:
It's been a bit more than a month, I think we're due for a release.
Lots of new code + docs and it would be nice for it to be on crates.io and docs.rs


thejpster (approved):

thejpster:
Are we doing cargo-publish on the BSPs, or just the HAL?

9names:
We haven't published any of the BSPs yet, but I think it's time that we should.

PR #123, opened 2021-09-19
Implement GPIO override

henkkuli:
Interrupt status, input, output enable and output can be overridden. See documentation for GPIO0_CTRL for more info.

As a background, I happen to need input inversion as my optocoupler inverts the signal and I don't want to modify my existing PIO code.

### DIFF
--- a/rp2040-hal/src/gpio/mod.rs
+++ b/rp2040-hal/src/gpio/mod.rs
@@ -62,3 +62,55 @@ pub enum OutputSlewRate {
     /// Slew fast
     Fast,
 }
+
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+/// Interrupt override state.
+pub enum InterruptOverride {
+    /// Don't invert the interrupt.
+    DontInvert = 0,
+    /// Invert the interrupt.
+    Invert = 1,
+    /// Drive interrupt low.
+    AlwaysLow = 2,
+    /// Drive interrupt high.
+    AlwaysHigh = 3,
+}
+
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+/// Input override state.
+pub enum InputOverride {
+    /// Don't invert the peripheral input.
+    DontInvert = 0,
+    /// Invert the peripheral input.
+    Invert = 1,
+    /// Drive peripheral output low.
+    AlwaysLow = 2,
+    /// Drive peripheral output high.
+    AlwaysHigh = 3,
+}
+
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+/// Output enable override state.
+pub enum OutputEnableOverride {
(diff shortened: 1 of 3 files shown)


rp2040-hal/src/gpio/mod.rs
```diff
@@ -62,3 +62,55 @@ pub enum OutputSlewRate {
...
+}
+
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+/// Input override state.
+pub enum InputOverride {
+    /// Don't invert the peripheral input.
+    DontInvert = 0,
+    /// Invert the peripheral input.
+    Invert = 1,
+    /// Drive peripheral output low.
```
9names, line 86: I assume these will drive the input low, not the output?

rp2040-hal/src/gpio/mod.rs
```diff
@@ -62,3 +62,55 @@ pub enum OutputSlewRate {
...
+
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+/// Output  override state.
+pub enum OutputOverride {
+    /// Use the original output  signal from selected peripheral.
+    DontInvert = 0,
+    /// Invert the output  signal from selected peripheral.
+    Invert = 1,
+    /// Drive output low.
+    Disable = 2,
```
9names, line 113: These should be AlwaysLow and AlwaysHigh like the input enums

rp2040-hal/src/gpio/pin.rs
```diff
@@ -517,12 +520,36 @@ where
...
+    /// Set the slew rate for the pin.
     #[inline]
     pub fn set_slew_rate(&mut self, rate: OutputSlewRate) {
         self.regs.write_slew_rate(rate)
     }
 
+    /// Set the interrupt override.
+    #[inline]
+    pub fn set_interrupt_override(&mut self, r#override: InterruptOverride) {
+        self.regs.set_interrupt_override(r#override);
```
9names, line 532: can we avoid the raw label on the argument? maybe something like override_value or over_ride instead.

rp2040-hal/src/gpio/mod.rs
henkkuli: That's also how I understand it.

rp2040-hal/src/gpio/mod.rs
henkkuli: Exactly!

rp2040-hal/src/gpio/pin.rs
henkkuli: I changed these to `override_value` now.

henkkuli:
Thanks for the review, and sorry for the clear bugs. Apparently trying to write documentation after 12 hours of physically active volunteer work isn't optimal :sweat_smile:

9names (approved):
Looks good to me!

PR #125, opened 2021-09-20
Add README to Pro Micro RP2040 BSP.

jonathanpallant:
Also add BSP to top-level README, and some extra comments in the
example.


9names (approved):
Lgtm

PR #131, opened 2021-09-25
implement embedded-hal 1.0.0-alpha.5

jannic:
With this change, it is possible to use rp2040-hal with embedded-hal 1.0.0-alpha.5.
The new trait implementations are hidden behind a feature flag, `eh1_0_alpha`, to avoid semver breaking changes as long as that flag is not set.

When using that flag, breaking changes can be expected, as embedded-hal 1.0 is not yet released. This is documented in README.md.

When embedded-hal 1.0 is released, we have several options:
- We could activate it unconditionally (and just remove the feature flag).
- We could rename the feature flag to `eh1_0` or similar.
- We could hide support for embedded-hal 0.2 behind a feature flag and support 1.0 by default.
- We could remove support for embedded-hal 0.2 and replace it with support for version 1.0.

Currently I'd prefer the first option, as long as support for 1.0 can be activated without breaking changes for existing users. The last to options are added for completeness, I don't think it would be sensible to remove support for embedded-hal 0.2 in the near future.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: check
-          args: --workspace --examples
+          args: --workspace --examples --features eh1_0_alpha
       - uses: actions-rs/cargo@v1
         with:
           command: test
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -12,7 +12,7 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 cortex-m = "0.7.2"
 embedded-hal = { version = "0.2.5", features = ["unproven"] }
-eh1_0_alpha = { version = "1.0.0-alpha.5", package="embedded-hal", optional=true }
+eh1_0_alpha = { version = "=1.0.0-alpha.5", package="embedded-hal", optional=true }
 embedded-time = "0.12.0"
 itertools = { version = "0.10.1", default-features = false }
 nb = "1.0"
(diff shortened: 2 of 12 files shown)


jannic:
Some CI checks failed due to an unrelated issue.
(See https://github.com/rp-rs/rp-hal/pull/132 )

9names:
If we're including the alpha hal, i recommend we use an exact match specifier to avoid semver issues.
```
eh1_0_alpha = { version = "=1.0.0-alpha.5", package="embedded-hal", optional=true }
```
There will not be multiple releases of an alpha, and each one is considered backwards incompatible so the safest option is to depend on that specific tag and only upgrade when we know it is safe.
There are other HALs that are already doing multiple HALs simultaneously (see lpc-8xx), I see no reason to hide this behind a config flag. It just makes it more likely to break without anyone noticing.

jannic:
> If we're including the alpha hal, i recommend we use an exact match specifier to avoid semver issues.

Very good point, I'll change that.

> There are other HALs that are already doing multiple HALs simultaneously (see lpc-8xx), I see no reason to hide this behind a config flag. It just makes it more likely to break without anyone noticing.

During alpha or after eh-1.0 is released? The cargo reference explicitly states: "It is recommended to not use pre-release dependencies in a library if your library is not also a pre-release." I agree and would prefer to keep the dependency behind a feature flag while it is an alpha version.

9names:
The safest assumption is that 1.0 will not release soon.
At the very least, if this is behind a feature flag we need to have a test building this code with this feature enabled as part of our CI or it will certainly suffer from bit rot.

jannic:
Good point, an automated test build would be good. Is it ok to enable the feature unconditionally for the test build (as I did in the commit a few seconds ago) or do we need separate tests with and without that feature enabled?

9names:
I'm happy to have it always on

9names (approved):
I'm okay with this PR getting merged, with the caveat that if it should be removed if it becomes a maintenance burden.
LGTM

PR #132, opened 2021-09-25
fix pico_usb_twitchy_mouse with usbd-hid >= 0.5.1

jannic:
usbd-hid 0.5.1 contained an accidental semver breaking change

fix example accordingly


9names (approved):
Ah that's unfortunate.
Nice catch though. LGTM.

PR #133, opened 2021-09-25
Add a simple example for PIO.

### LINKED ISSUE #114: Missing example for PIO
Now that we've landed PIO support, we should get a basic example up and going showing how to use it.
This is separate to any other, complex examples using PIO for specific tasks too.

jannic:
Just a simple blinky, this time implemented using PIO.
This fixes #114.

### DIFF
--- a/rp2040-hal/examples/pio_blink.rs
+++ b/rp2040-hal/examples/pio_blink.rs
@@ -1,6 +1,6 @@
-//! Blinks the LED on a Pico board, using a PIO program
+//! This example toggles the GPIO25 pin, using a PIO program.
 //!
-//! This will blink an LED attached to GP25, which is the pin the Pico uses for the on-board LED.
+//! If a LED is connected to that pin, like on a Pico board, the LED should blink.
 #![no_std]
 #![no_main]
 
--- a/rp2040-hal/examples/pio_blink.rs
+++ b/rp2040-hal/examples/pio_blink.rs
@@ -39,13 +39,17 @@ fn main() -> ! {
     let mut wrap_source = a.label();
     // Set pin as Out
     a.set(pio::SetDestination::PINDIRS, 1);
+    // Define begin of program loop
     a.bind(&mut wrap_target);
     // Set pin low
     a.set_with_delay(pio::SetDestination::PINS, 0, MAX_DELAY);
     // Set pin high
     a.set_with_delay(pio::SetDestination::PINS, 1, MAX_DELAY);
-    // NOP
+    // Define end of program loop
     a.bind(&mut wrap_source);
+    // The labels wrap_target and wrap_source, as set above,
+    // define a loop which is executed repeatedly by the PIO
+    // state machine.
     let program = a.assemble_with_wrap(wrap_source, wrap_target);
 
     // Initialize and start PIO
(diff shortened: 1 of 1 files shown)


9names (commented):
I've tested this and it works for me.
Looks good, though I'd like someone more familiar with PIO to review to verify.
Some comments about how the wrapping labels form a loop would be nice as well, and how that NOP is created?

jannic:
Oops, sorry, that NOP is a leftover comment. I had NOPs in earlier versions (similar to the one here https://github.com/ithinuel/ws2812-pio-rs/blob/main/src/lib.rs#L56 ) but removed them later, to keep the example as simple as possible.

mgottschlag:
I found this very helpful as-is. I will create a pull request for pio-rs w.r.t. pio_proc::pio!() to fix usage in no-std code - once that works, I'd suggest adding an example for that macro as well, because it is far more comfortable.

jonathanpallant:
Please could you add this to https://github.com/rp-rs/rp-hal/blob/main/boards/pico/README.md, so it all says in sync? Thanks!

jannic:
Sure, I wasn't aware there was a list of examples in the README.

jannic:
@jonathanpallant, as the example is located in `rp2040-hal/examples/pio_blink.rs`, not `boards/pico/examples/...`, is the README you mentioned really the right place to add it to?
Or should the example be moved?
(I'm not sure about that: On the one hand, the PIO is not specific to the pico board, so the point of the example applies to all boards using an RP2040. On the other hand, it does use the pico boot loader and GPIO25 for the LED, which would not be correct for other boards. But in that sense, most examples in `rp2040-hal/examples` would not run unmodified on all boards.)

jonathanpallant:
This is a good point! Sorry, I did not spot which folder the example was in.

I think this example should be in the `pico` folder as it is specifically designed to work on the Pico and not all RP2040 boards have an LED on Pin 25. However, I can also see a use for an example in the `rp2040-hal/examples` folder which demonstrates the use of the PIO.

Perhaps an example in `rp2040-hal/examples` could talk in generic terms about setting up a PIO program to use an arbitrary pin, and an example in the `boards/pico/examples` folder could talk in concrete terms in blinking an LED on GPIO25.

If you're happy to do that, then please do update this PR. If not, we can merge this as-is and I can split it into two examples - one designed to run on a Pico and one that shows how to use the PIO crate.  I will also make a note to review the other examples and ensure nothing in `rp2040-hal` makes any assumption about which board it runs on.

jannic:
How could a generic example work? Sure, it could toggle some arbitrary GPIO pin. But that's not better than the current example, just that this arbitrary pin happens to be connected to a LED on the Pico.
Perhaps I could leave the example in `rp2040-hal/examples` but change the description? Like, "this example toggles a GPIO pin. If it is connected to a LED, like on the rp pico, the LED should blink"?

jonathanpallant:
It would be an example for reading rather than an example for executing - because as you say, it can't make any assumptions about the pin arrangements. I'll make it clear in the rp-hal README that all the examples are of that type.

Yes the code is fine as-is - the subtlety is in the documentation. Your proposed wording is good.

jonathanpallant (approved):

jonathanpallant:
Brilliant - thanks!

PR #134, opened 2021-09-25
Implement bus keep inputs, as documented in C SDK

jannic:
When both pull-up and pull-down are enabled, the RP2040 enters a
so-called "bus keep" function, which uses a weak pull to tue current
high/low state of the GPIO.

See for example
https://raspberrypi.github.io/pico-sdk-doxygen/group__hardware__gpio.html#gab6bf9552da32b3dd0a5d0db45d8374fc

As this change adds new variants to public enums, it is a semver breaking change and therefore needs an appropriate version increment, when merged.

### DIFF
--- a/rp2040-hal/src/gpio/dynpin.rs
+++ b/rp2040-hal/src/gpio/dynpin.rs
@@ -91,6 +91,7 @@ pub enum DynDisabled {
     Floating,
     PullDown,
     PullUp,
+    BusKeep,
 }
 
 /// Value-level `enum` for input configurations
@@ -100,6 +101,7 @@ pub enum DynInput {
     Floating,
     PullDown,
     PullUp,
+    BusKeep,
 }
 
 /// Value-level `enum` for output configurations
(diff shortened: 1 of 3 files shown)


9names (commented):
LGTM.
Since we've already bumped version for the next release but have not actually pushed that release, I don't think we need to bump the version again.
If I understand it correctly, the point of bumping versions on pub enum changes is that it would break exhaustive matching - do you believe that there is user code with match on pin configuration?

jannic:
Right, another version bump is not necessary if this gets merged before 0.3.0 is released.
Regarding a match on pin configuration: While I don't see an obvious use case, I don't think it's completely unreasonable. In any case, I tend to follow the semver compatibility chapter of the cargo book: https://doc.rust-lang.org/cargo/reference/semver.html#enum-variant-new

9names (approved):
If we're not waiting for another release first, I think this is ready to merge.
LGTM

PR #135, opened 2021-09-25
Implement RngCore for RingOscillator

robyoung:
Implement [`rand_core::RngCore`](https://docs.rs/rand_core/0.6.3/rand_core/trait.RngCore.html) for `RingOscillator` from the `get_random_bit` function. This is not suitable for security purposes so [`rand_core::CryptoRng`](https://docs.rs/rand_core/0.6.3/rand_core/trait.CryptoRng.html) has not been implemented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- implement `rng_core::RngCore` for `RingOscillator`
+
 ## [0.3.0] - 2021-09-20
 
 ### Added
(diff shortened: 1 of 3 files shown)


9names (commented):
This looks great, thanks!

Just for fun I thought I'd see if i could make a more optimal version of this. By putting it into a function with a local variable it ended up a little tighter in the inner loop, at least in the optimize-for-size case.
Harder to tell just by looking if the O3 version is any better though.

https://rust.godbolt.org/z/bnTx9sMnM

robyoung:
Thanks for taking a look. I'd be happy to go either way. I don't have enough experience to really have an opinion on which is better. If you think it would be an improvement I can make the change when I get home later today.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- implement `rng_core::RngCore` for `RingOscillator`
```
9names, line 11: Typo here: rand_code::RngCore

9names (approved):
I don't think it's necessary to make the code change, especially when I haven't really tested it for correctness or performance.
The typo in the changelog is largely inconsequential, I'll clean it up in my next PR if no-one else gets there first.
LGTM

PR #136, opened 2021-09-26
Adds Reset for Timer

Lichtso:
otherwise it would not run

### DIFF
--- a/boards/pico/examples/pico_countdown_blinky.rs
+++ b/boards/pico/examples/pico_countdown_blinky.rs
@@ -61,7 +61,7 @@ fn main() -> ! {
     .unwrap();
 
     // Configure the Timer peripheral in count-down mode
-    let timer = hal::timer::Timer::new(pac.TIMER);
+    let timer = hal::timer::Timer::new(pac.TIMER, &mut pac.RESETS);
     let mut count_down = timer.count_down();
 
     // The single-cycle I/O block controls our GPIO pins
--- a/boards/pico/examples/pico_usb_serial.rs
+++ b/boards/pico/examples/pico_usb_serial.rs
@@ -89,7 +89,7 @@ fn main() -> ! {
         .device_class(2) // from: https://www.usb.org/defined-class-codes
         .build();
 
-    let timer = hal::timer::Timer::new(pac.TIMER);
+    let timer = hal::timer::Timer::new(pac.TIMER, &mut pac.RESETS);
     let mut said_hello = false;
     loop {
         // A welcome message at the beginning
--- a/boards/pro_micro_rp2040/examples/pro_micro_rainbow.rs
+++ b/boards/pro_micro_rp2040/examples/pro_micro_rainbow.rs
@@ -71,7 +71,7 @@ fn main() -> ! {
 
     let _led: Pin<_, FunctionPio0> = pins.led.into_mode();
 
-    let timer = Timer::new(pac.TIMER);
+    let timer = Timer::new(pac.TIMER, &mut pac.RESETS);
     let mut delay = timer.count_down();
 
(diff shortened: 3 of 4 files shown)


9names:
The timer demos were already working - are we resetting the timer somewhere else?

Lichtso:
I was wondering about that too, as I could not get any of the timer examples to run.
So, at least in my builds the timer does not get reset anywhere else.

jannic:
I do not understand why this could be necessary:
According to the data sheet, the chip starts with timer reset asserted. Then, as the very first thing done by the boot ROM, the timer reset flag gets cleared:
https://github.com/raspberrypi/pico-bootrom/blob/master/bootrom/bootrom_main.c#L227-L231
So at the point where the firmware get called, the timer should be in a defined state and out of reset.
It may tick very slowly if the clocks are not properly configured, though. But a timer reset wouldn't help to fix that.

There is another condition for the timer to work: "The  Watchdog  tick  (see  Section  4.7.2)  must  be  running  for  the  timer  to  start  counting.  The  SDK  starts  this  tick  as part of the platform initialisation code". The watchdog tick is also enabled at bootup.

jonathanpallant:
@Lichtso - how are you loading the firmware on to the RP2040?

Lichtso:
I am not uploading to Flash, but RAM directly.
And yes, that might make a difference in what the firmware initializes or leaves out.

jannic:
How do you upload directly to RAM? Using a debug probe? I guess the debugger is jumping directly to your entry point, without any initialization by the boot ROM. That explains why the timer reset flag is not cleared.
I'm still not sure that this is the right solution. Some initialization helper method you could call in case your software runs from RAM would be more general, as we could add other missing setup steps there, as well.

Lichtso:
No I am not using JTAG or anything, just the UF2 with a different destination address and upload it via USB in boot select mode.

> Some initialization helper method you could call in case your software runs from RAM would be more general, as we could add other missing setup steps there, as well.

If there are other "things" missing, sure. But so far this has been the only thing I noticed.

jannic:
Ok I can reproduce it. And indeed, it's not that the timer needs an additional reset, but it needs to be taken out of reset. Those two lines are sufficient to make the timer run:
```
    pac.RESETS.reset.modify(|_, w| w.timer().clear_bit());
    while pac.RESETS.reset_done.read().timer().bit_is_clear() {}
```
BTW, is there some tutorial on how to build an uf2 file for loading to RAM? I finally got it working, but not without some wild edits to the linker script. (Moving everything to RAM was easy, but then I had to move the vector table because elf2uf2 expects the entry point to be the first thing. `Error: "A RAM binary should have an entry point at the beginning: 0x20000001 (not 0x200000a9)\n"`)

I'm still not fully convinced that it's correct to add  this to `Timer::new`, but I'd be much more comfortable with just clearing the reset bit, than with an unconditional reset.

Lichtso:
Then I will remove `timer.reset_bring_down(resets);` and only leave `timer.reset_bring_up(resets);` in there.

> BTW, is there some tutorial on how to build an uf2 file for loading to RAM?

I made [this](https://github.com/Lichtso/rp2040_template) for quick and easy development. It requires only one command to build, upload and run: `cargo run`
And yes, it requires different linker scripts depending on the deployment.

Lichtso:
So any opinions, merge or reject?

jonathanpallant:
I think it's unreasonable for our drivers to rely on any state left by the bootloader because, as we've seen, the bootloader doesn't always run. Also, newer versions of the silicon might have a different ROM bootloader.

So I'm happy to merge.

jonathanpallant (approved):

PR #142, opened 2021-09-28
Add an example using the Feather RP2040's onboard NeoPixel

jwillikers:
This uses PIO to drive the NeoPixel just like the Pro Micro example does.


9names (approved):
Looks good to me, thanks!

PR #143, opened 2021-09-28
Rework the PIO API

### LINKED ISSUE #141: PIO sideset does not work (wrong pindir)
I tried to write a PIO program that used sideset, but sideset assignments did not have any effect. It turns out that apparently it is necessary to set the pindir of the sideset pins before using them. The documentation calls the function `pio_sm_set_pindirs_with_mask()` for that which takes a 32-bit bitmask and sets the pindir for all currently selected `set` pins according to the mask. The following code is equivalent to that function, to be placed in StateMachine:

    pub fn set_pindirs_with_mask(&self, mut pins: u32, pindir: u32) {
        let mut pin = 0;
        while pins != 0 {
            if (pins & 1) != 0 {
                // Select a single "set" pin.
                self.sm().sm_pinctrl.write(|w| {
                    unsafe {
                        w.set_count().bits(1);
                        w.set_base().bits(pin as u8);
                    }
                    w
                });
                // Execute "set pindir, <dir>", where "dir" is 1 if the pin is in the mask.
                self.sm().sm_instr.write(|w| {
                    unsafe {
                        w.sm0_instr().bits(0xe080 | ((pindir >> pin) & 0x1) as u16);
                    }
                    w
                });
            }
            pin += 1;
            pins = pins >> 1;
        }
    }

This function has the nice sideeffect that it often saves a `set pindir, ...` instruction at the beginning of the program.

I do not know how to properly integrate such a function into the PIO abstraction - the Pico SDK implementation just saves and restores selected pins in that function. That, however, does not work if the state machine is currently active - although, if I understand it correctly, the current PIO abstraction does not try to provide any safeguards against such misuse? Other HALs use a lot more typestates for that.

mgottschlag:
# Rationale

The current PIO API has a number of limitations:
- As the state machines are contained in `PIO`, they cannot be moved around individually. Sometimes, a single PIO block implements multiple functions that are used in completely different parts of the code base.
- `StateMachine` is not Send, although it could be.
- Multiple state machines should be able to share code to reduce instruction space usage, yet currently they are not because PIOBuilder always allocates space. Sometimes, multiple state machines implement the same functionality (e.g., multiple SPI or I2C buses).
- Some functions must only be called in specific configurations. For example, pin directions have to be set while the state machine is stopped, as PINCTRL is modified. The current API does not reflect such restrictions, which makes it harder to use than necessary.

Also, during my work, I found a bug where used instruction space was not marked properly.

This PR also fixes #141 by providing a function to set pin directions.

# Design

This PR uses different types to represent state machines in different states - `UninitStateMachine<P>` is not associated with any program, whereas `StateMachine<P, Stopped>` and `StateMachine<P, Running>` are. Program installation in instruction memory is separated from state machine initialization via `InstalledProgram` and `PIO::install()`. Access from state machines to shared registers is performed via atomic operations to enable `Send`.

Old usage example:
```
let pio = rp2040_hal::pio::PIO::new(pac.PIO0, &mut pac.RESETS);
let sm = &pio.state_machines()[0];
let div = 0f32; // as slow as possible (0 is interpreted as 65536)
rp2040_hal::pio::PIOBuilder::default()
    .with_program(&program)
    .set_pins(led_pin_id, 1)
    .clock_divisor(div)
    .build(&pio, sm)
    .unwrap();
```
New usage example:
```
let (mut pio, sm0, _, _, _) = pac.PIO0.split(&mut pac.RESETS);
let installed = pio.install(&program).unwrap();
let div = 0f32; // as slow as possible (0 is interpreted as 65536)
let sm = rp2040_hal::pio::PIOBuilder::from_program(installed)
    .set_pins(led_pin_id, 1)
    .clock_divisor(div)
    .build(sm0);
sm.start();
```

# Testing

I tested that the blink examples work, and I tested with my own code that, however, does not use much more PIO functionality yet (except for sideset pins). Those programs work fine.

### DIFF
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -11,6 +11,7 @@ pub trait PIOExt:
     core::ops::Deref<Target = rp2040_pac::pio0::RegisterBlock> + SubsystemReset + Sized
 {
     /// Create a new PIO wrapper and split the state machines into individual objects.
+    #[allow(clippy::type_complexity)] // Required for symmetry with PIO::free().
     fn split(
         self,
         resets: &mut pac::RESETS,
@@ -188,8 +189,7 @@ impl<P: PIOExt> PIO<P> {
             {
                 self.pio.instr_mem[i + offset].write(|w| unsafe { w.bits(instr as u32) })
             }
-            self.used_instruction_space =
-                self.used_instruction_space | (((1 << p.code.len()) - 1) << offset);
+            self.used_instruction_space |= ((1 << p.code.len()) - 1) << offset;
             Ok(InstalledProgram {
                 offset: offset as u8,
                 length: p.code.len() as u8,
@@ -205,7 +205,7 @@ impl<P: PIOExt> PIO<P> {
     /// Removes the specified program from instruction memory, freeing the allocated space.
     pub fn uninstall(&mut self, p: InstalledProgram<P>) {
         let instr_mask = ((1 << p.length as u32) - 1) << p.offset as u32;
-        self.used_instruction_space = self.used_instruction_space & !instr_mask;
+        self.used_instruction_space &= !instr_mask;
     }
 }
 
@@ -220,12 +220,22 @@ impl<P: PIOExt> PIO<P> {
 ///
 /// # Examples
 ///
-/// ```
-/// let (mut pio, sm0, _, _, _) = pac.PIO0.split(&mut pac.RESETS);
+/// ```no_run
+/// use rp2040_hal::{pac, pio::PIOBuilder, pio::PIOExt};
+/// let mut peripherals = pac::Peripherals::take().unwrap();
+/// let (mut pio, sm0, _, _, _) = peripherals.PIO0.split(&mut peripherals.RESETS);
 /// // Install a program in instruction memory.
+/// let program = pio_proc::pio!(
+///     32,
+///     ".wrap_target
+///     set pins, 1 [31]
+///     set pins, 0 [31]
+/// .wrap
+///     "
+/// ).program;
 /// let installed = pio.install(&program).unwrap();
 /// // Configure a state machine to use the program.
-/// let (sm, rx, tx) = rp2040_hal::pio::PIOBuilder::from_program(installed).build(sm0);
+/// let (sm, rx, tx) = PIOBuilder::from_program(installed).build(sm0);
 /// // Uninitialize the state machine again, freeing the program.
 /// let (sm, installed) = sm.uninit(rx, tx);
 /// // Uninstall the program to free instruction memory.
@@ -239,7 +249,7 @@ impl<P: PIOExt> PIO<P> {
 /// the program anymore. The user must therefore make sure that `uninstall()` is only called on the
 /// PIO object which was used to install the program.
 ///
-/// ```
+/// ```ignore
 /// let (mut pio, sm0, sm1, sm2, sm3) = pac.PIO0.split(&mut pac.RESETS);
 /// // Install a program in instruction memory.
 /// let installed = pio.install(&program).unwrap();
@@ -530,7 +540,7 @@ impl<SM: ValidStateMachine> StateMachine<SM, Stopped> {
                 });
             }
             pin += 1;
-            pins = pins >> 1;
+            pins >>= 1;
         }
         // We modified PINCTRL, yet the program assumes a certain configuration, so restore the
         // previous value.
@@ -554,7 +564,7 @@ impl<P: PIOExt, SM: StateMachineIndex> StateMachine<(P, SM), Stopped> {
     ///
     /// # Example
     ///
-    /// ```
+    /// ```ignore
     /// sm0.synchronize_with(sm1).and_with(sm2);
     /// ```
     pub fn synchronize_with<'sm, SM2: StateMachineIndex>(
@@ -1151,6 +1161,7 @@ impl<P: PIOExt> PIOBuilder<P> {
     }
 
     /// Build the config and deploy it to a StateMachine.
+    #[allow(clippy::type_complexity)] // The return type cannot really be simplified.
     pub fn build<SM: StateMachineIndex>(
         self,
         mut sm: UninitStateMachine<(P, SM)>,
@@ -1268,7 +1279,7 @@ impl<P: PIOExt> PIOBuilder<P> {
         };
         (
             StateMachine {
-                sm: sm,
+                sm,
                 program: self.program,
                 _phantom: core::marker::PhantomData,
             },
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -96,6 +96,8 @@ unsafe impl<P: PIOExt + Send> Send for PIO<P> {}
 
 impl<P: PIOExt> PIO<P> {
     /// Free this instance.
+    ///
+    /// All output pins are left in their current state.
     pub fn free(
         self,
         _sm0: UninitStateMachine<P>,
@@ -103,7 +105,7 @@ impl<P: PIOExt> PIO<P> {
         _sm2: UninitStateMachine<P>,
         _sm3: UninitStateMachine<P>,
     ) -> P {
-        // TODO: Disable the PIO block.
+        // All state machines have already been stopped.
         self.pio
     }
 
@@ -242,7 +244,19 @@ impl<P: PIOExt> PIO<P> {
 /// `PIO::uninstall(program)` can be used to free the space occupied by the program once it is no
 /// longer used.
 ///
-/// TODO: Write an example?
+/// # Examples
+///
+/// ```
+/// let (mut pio, sm0, _, _, _) = pac.PIO0.split(&mut pac.RESETS);
+/// // Install a program in instruction memory.
+/// let installed = pio.install(&program).unwrap();
+/// // Configure a state machine to use the program.
+/// let sm = rp2040_hal::pio::PIOBuilder::from_program(installed).build(sm0);
+/// // Uninitialize the state machine again, freeing the program.
+/// let (sm, installed) = sm.uninit();
+/// // Uninstall the program to free instruction memory.
+/// pio.uninstall(installed);
+/// ```
 ///
 /// # Safety
 ///
@@ -251,7 +265,16 @@ impl<P: PIOExt> PIO<P> {
 /// the program anymore. The user must therefore make sure that `uninstall()` is only called on the
 /// PIO object which was used to install the program.
 ///
-/// TODO: Write an example?
+/// ```
+/// let (mut pio, sm0, sm1, sm2, sm3) = pac.PIO0.split(&mut pac.RESETS);
+/// // Install a program in instruction memory.
+/// let installed = pio.install(&program).unwrap();
+/// // Reinitialize PIO.
+/// let pio0 = pio.free(sm0, sm1, sm2, sm3);
+/// let (mut pio, _, _, _, _) = pio0.split(&mut pac.RESETS);
+/// // Do not do the following, the program is not in instruction memory anymore!
+/// pio.uninstall(installed);
+/// ```
 #[derive(Debug)]
 pub struct InstalledProgram<P> {
     offset: u8,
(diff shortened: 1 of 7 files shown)


mgottschlag:
The new example in this PR currently depends on rp-rs/pio-rs#9, so the two PRs should be merged together.

rp2040-hal/src/pio.rs
```diff
@@ -233,72 +371,180 @@ impl<P: Instance> StateMachine<P> {
...
+                self.sm.sm().sm_pinctrl.write(|w| {
+                    unsafe {
+                        w.set_count().bits(1);
+                        w.set_base().bits(pin as u8);
+                    }
+                    w
+                });
+                self.sm.sm().sm_instr.write(|w| {
+                    unsafe {
+                        const SET_PINDIRS: u16 = 0xe080;
```
henkkuli, line 501: We should be able to use instruction encoder here. See https://github.com/rp-rs/rp-hal/blob/833b69819e0d91b708c1890c25d0f7ecab0964a1/rp2040-hal/src/pio.rs#L168

rp2040-hal/src/pio.rs
```diff
@@ -165,57 +185,175 @@ impl<P: Instance> PIO<P> {
...
+            _phantom: core::marker::PhantomData,
         }
     }
 }
 
-/// PIO State Machine.
+/// PIO State Machine (uninitialized, without a program).
 #[derive(Debug)]
-pub struct StateMachine<P: Instance> {
+pub struct UninitStateMachine<P: PIOExt> {
```
henkkuli, line 311: I'd like to see this be `StateMachine<P, Uninit>`, but when I tried to draft an example of that I realized we probably need GATs for that because `InstalledProgram` is generic over `P`. Another option we could have today is
```rust
StateMachine<Uninit<PIO0>>;
StateMachine<StoppedPIO0>>;
StateMachine<Starged<PIO0>>;
```
but I don't like that either.

henkkuli (commented):
I like these improvements to the usability of the API. There are a couple of bugs I spotted, and probably some more I missed.

I didn't know that atomic access to peripherals was even possible. My only concern is that I don't exactly know whether using pointers like that is sound, and I also don't like manual pointer arithmetic in HAL crate. In general I think the PAC should provide us with atomic access. A quick search through `svd2rust` found that it is probably possible as apparently MSP430 devices have similar functionality: https://github.com/rust-embedded/svd2rust/blob/master/src/generate/generic_msp430_atomic.rs

henkkuli:
Apparently there are already issues for [`rp2040-pac`](https://github.com/rp-rs/rp2040-pac/issues/25) and [`svd2rust`](https://github.com/rust-embedded/svd2rust/issues/535).

mgottschlag:
Thanks a lot for the review. I will fix those bugs this evening.

After submitting the PR, I started to think about integrating DMA. While doing so, I noticed that the API needs some further restructuring. In particular, TX and RX FIFOs need to be managed by different objects than the state machine itself, so that the user is able to use TX and RX DMA in parallel. For example, an API such as the following may be possible:
```
let (sm, rx, tx) = rp2040_hal::pio::PIOBuilder::from_program(installed)
    .set_pins(led_pin_id, 1)
    .clock_divisor(div)
    .build(sm0);
let rx_dma = rx.read_with_dma(buffer);
sm.start();
// ...
let rx = rx_dma.wait().
let sm0 = sm.uninit(rx, tx);
```

However, in the last line, the API should ensure that only the right rx and tx objects can be used. This could, for example, be done by adding another generic parameter, where there would be the following types and `sm.id` would be removed:
```
UninitStateMachine<PIO, SM0>
StateMachine<PIO0, SM0, Running>
PIOTx<PIO0, SM0>
```
etc.

That, however, means that any function/type that receives a state machine object needs to be parametrized for the specific state machine index. I do not think that this would be a large problem, as we already have similar behavior for objects/functions operating on GPIO pins, where they need to be written as generics which take a specific GPIO pin type.

What do you think?

rp2040-hal/src/pio.rs
mgottschlag: I tried the latter, but did not like that it made it more difficult to implement functions that can operate on `Stopped` and `Running`, but not on `Uninit`. There are quite some such functions that just do not make sense on an UninitStateMachine.

I do not see this as a substantial usability concern, but I can look further into it if you insist.

mgottschlag:
Oh, and with regards to MSP430 atomics: The scenario is somewhat different as MSP430, if I understand it correctly, uses atomic instructions to operate at the same addresses, which is probably much easier to implement in svd2rust. This kind of atomic register access (register aliases) would probably have to be implemented in rp2040-pac, but I have no idea how that would be done.

henkkuli:
I think having everything be generic over the state machine won't be a problem. As you said, similar API is used elsewhere. Maybe one way to make it little easier to use would be to have types `PIO0SM0`, `PIO0SM1`, etc. and be generic over those instead. This way a function taking a state machine could have signature
```rust
fn foo<SM: ValidStateMachine>(_: StateMachine<SM, Running>) {}
```
instead of
```rust
fn foo<PIO: ValidPIO, SM: ValidStateMachine>(_: StateMachine<PIO, SM, Running>) {}
```

Regarding the atomics, I think I managed to hack something together already, though it is currently still a hack. Basically I just copied https://github.com/rust-embedded/svd2rust/blob/master/src/generate/generic_msp430_atomic.rs, changed the writes to something like
```rust
self.register
    .as_ptr()
    .add(0x2000 / core::mem::size_of::<REG::Ux>())
    .write_volatile(bits);
```
and force-added the file to every compiled crate. This has a couple of problems I haven't resolved yet: The most obvious one is that not every PAC crate should have this added, so a new flag or target needs to be added. More subtle problem is that the RP2040 documentation states that SIO device doesn't support atomic access (as it already is atomic by design), so this API shouldn't be exposed for SIO device.

rp2040-hal/src/pio.rs
```diff
@@ -6,23 +6,82 @@ use pio::{Program, SideSet, Wrap};
...
+        let sm0 = UninitStateMachine {
+            id: 0,
+            block: self.deref(),
+            sm: &self.deref().sm[0],
+            _phantom: core::marker::PhantomData,
+        };
+        let sm1 = UninitStateMachine {
+            id: 0,
+            block: self.deref(),
+            sm: &self.deref().sm[0],
```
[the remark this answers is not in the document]
mgottschlag: Fixed

rp2040-hal/src/pio.rs
```diff
@@ -165,57 +185,175 @@ impl<P: Instance> PIO<P> {
...
+            })
         } else {
-            None
+            Err(InstallError::NoSpace)
+        }
+    }
+
+    /// Removes the specified program from instruction memory, freeing the allocated space.
+    pub fn uninstall(&mut self, p: InstalledProgram<P>) {
+        let instr_mask = (1 << p.length as u32) << p.offset as u32;
```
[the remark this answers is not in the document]
mgottschlag: Fixed.

rp2040-hal/src/pio.rs
```diff
@@ -165,57 +185,175 @@ impl<P: Instance> PIO<P> {
...
+            sm_mask |= ((1 << sm.sm.id) as u32) << 8;
+        }
+        const ATOMIC_SET_OFFSET: usize = 0x2000;
+        // Safety: We only use the atomic alias of the register.
+        unsafe {
+            *(*sm_set[0].sm.block)
+                .ctrl
+                .as_ptr()
+                .add(ATOMIC_SET_OFFSET / 4) = sm_mask;
+        }
```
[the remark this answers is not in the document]
mgottschlag: I added write_volatile in all three places. This is _probably_ still undefined behavior - the documentation talks about the bounds of allocated memory, so I am not sure. Even if it is, I do not see a problem on this hardware platform.

rp2040-hal/src/pio.rs
mgottschlag: The code got slightly longer - is this what you mean?

mgottschlag:
I added parametrization for the state machines. I have to admit that my type design skills are not really great, so someone might want to check whether that's really the most elegant way to specify state machine indices.

Currently, the state machines still carry references around to the registers:
```
block: *const rp2040_pac::pio0::RegisterBlock,
sm: *const rp2040_pac::pio0::SM,
```
These fields have basically become unnecessary now, I will remove them in the next commit.

9names:
If you need atomics, what you'll want is the atomic- polyfill crate. It's not multi-core safe, but we are planning on making a fork that is.

9names:
> and force-added the file to every compiled crate. This has a couple of problems I haven't resolved yet: The most obvious one is that not every PAC crate should have this added, so a new flag or target needs to be added. More subtle problem is that the RP2040 documentation states that SIO device doesn't support atomic access (as it already is atomic by design), so this API shouldn't be exposed for SIO device.

Definitely sounds like something that should be opt-in.
I was thinking more along the lines of using svdtools to copy the peripheral, add a suffix _or, and change the peripheral offset. That would work with current svd2rust, won't break any existing uses but might make the interface a little less nice to deal with.

mgottschlag:
> How many registers do you need to add this for? Adding a few isn't a big deal.

Currently, the code only needs atomic access to the CTRL register.

rp2040-hal/src/pio.rs
```diff
@@ -650,6 +616,54 @@ impl<SM: ValidStateMachine> StateMachine<SM, Running> {
...
+pub struct Rx<SM: ValidStateMachine> {
+    block: *const rp2040_pac::pio0::RegisterBlock,
+    _phantom: core::marker::PhantomData<SM>,
+}
+
+impl<SM: ValidStateMachine> Rx<SM> {
+    /// Get the next element from RX FIFO.
+    ///
+    /// Returns `None` if the FIFO is empty.
+    pub fn read_rx(&mut self) -> Option<u32> {
```
henkkuli, line 629: Could this be just `read`? And `write` below?

henkkuli (commented):
I'm thinking whether having a transceiver type would be helpful or not, the idea being that it would provide convenient methods `read` and `write`, and `split` to create separate `Rx` and `Tx` channels.

henkkuli:
> Definitely sounds like something that should be opt-in.

Exactly. My implementation was just a quick POC to just check whether adding atomic methods to PAC is possible or not.

> I was thinking more along the lines of using svdtools to copy the peripheral, add a suffix _or, and change the peripheral offset. That would work with current svd2rust, won't break any existing uses but might make the interface a little less nice to deal with.

I don't think copying the peripheral should be the first solution. Especially as then the `Peripherals` struct would have both `PIO0` and `PIO0_or` fields, and both would need to be given to HAL, degrading the UX of the HAL greatly. Otherwise the user could use `PIO0_or` and `PIO0_nand` to modify the registers even if they had given access to `PIO0` away.

Instead I suggest that the SVD file should contain an attribute for every register telling whether the register has atomic counterparts or not. This would then instruct the codegen to implement `AtomicAccess` trait on the register, allowing us to gate the access to the atomic methods by this trait. I'm just not familiar enough with SVD to know whether this is possible, or how difficult that would be to implement.

rp2040-hal/src/pio.rs
henkkuli: Yes, this is what I meant.

rp2040-hal/src/pio.rs
henkkuli: I see. I guess one could still have something like
```rust
trait Initialized {...}
impl Initialized for Stopped {...}
impl Initialized for Running {...}
impl<State: Initialized> StateMachine<State> {...}
```
if one wanted to implement something only for `Stopped` and `Running`. But I don't insist on changing this, I'm just trying to brainstorm some ideas on how to make the API more self-consistent.

9names:
@mgottschlag Could you add an entry to the changelog for this?
I want CI to run with the latest pio-rs now that https://github.com/rp-rs/pio-rs/pull/9 is merged, after that I think this PR is good merge.

rp2040-hal/src/pio.rs
mgottschlag: Changed.

mgottschlag:
I added a changelog entry.

mgottschlag:
I will have a look at the failures this evening.

mgottschlag:
I fixed all clippy warnings, except for two warnings about very complex types that I somehow can't see in the CI output. I also marked most doc comment code examples as `ignore`, except for one which extended with `use` statements and a simple PIO program and marked as `no-run`.

I told clippy to ignore the return types of split() and build(), as those tuples are too complex for clippy. We could also package those into structs, but I do not think that it would help make the code more readable. GPIOs are placed into a struct instead of a tuple, for example, but there is no free() method for GPIOs. Here, there is, and that method simply expects the variables as returned by split() at the moment, and I think this symmetry is good.

9names (approved):
LGTM

PR #144, opened 2021-09-29
Implement QT Py BSP with example

onnenon:
This PR implements a BSP for the Adafruit QT Py RP2040 found [here](https://www.adafruit.com/product/4900).

Neopixel example has been tested and is working on actual hardware.

Pin mappings were taken from Adafruit's CircuitPython repo and can be found [here](https://github.com/adafruit/circuitpython/blob/main/ports/raspberrypi/boards/adafruit_qtpy_rp2040/pins.c).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 members = [
     "rp2040-hal",
     "boards/feather_rp2040",
+    "boards/qt_py_rp2040",
     "boards/pico",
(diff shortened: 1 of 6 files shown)


onnenon:
Amended my previous commit because I forgot to update the root readme.

9names (approved):
Looks good to me, thanks for the PR!

PR #146, opened 2021-09-30
Fix i2c write read

jannic:
Fixed i2c write_read function. Without that fix, reading an mpu6050 (using https://crates.io/crates/mpu6050) resulted in erratic values.


9names (approved):
Nice catch. LGTM.

PR #147, opened 2021-10-01
GPIO interrupt

Lichtso:
Both #4 and #24 left interrupts for GPIO as todos.

Adds the following methods for each PIN:
- `clear_interrupt(interrupt: Interrupt)`
- `interrupt_status(interrupt: Interrupt) -> bool`
- `is_interrupt_enabled(interrupt: Interrupt) -> bool`
- `set_interrupt_enabled(interrupt: Interrupt, enabled: bool)`
- `is_interrupt_forced(interrupt: Interrupt) -> bool`
- `set_interrupt_forced(interrupt: Interrupt, forced: bool)`

### DIFF
--- a/rp2040-hal/src/gpio/reg.rs
+++ b/rp2040-hal/src/gpio/reg.rs
@@ -1,8 +1,8 @@
 // Based heavily on and in some places copied from `atsamd-hal` gpio::v2
 use super::dynpin::{DynGroup, DynPinId};
 use super::{
-    InputOverride, InterruptOverride, OutputDriveStrength, OutputEnableOverride, OutputOverride,
-    OutputSlewRate,
+    InputOverride, Interrupt, InterruptOverride, OutputDriveStrength, OutputEnableOverride,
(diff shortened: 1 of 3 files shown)


rp2040-hal/src/gpio/reg.rs
```diff
@@ -255,6 +255,101 @@ pub(super) unsafe trait RegisterInterface {
...
+    /// Interrupt status.
+    #[inline]
+    fn interrupt_status(&self, interrupt: Interrupt) -> bool {
+        let num = self.id().num as usize;
+        unsafe {
+            let cpuid = *(pac::SIO::ptr() as *const u32);
+            let io = &(*pac::IO_BANK0::ptr());
+            let reg = (&io.proc0_ints0)
+                .as_ptr()
+                .add(num / 8 + cpuid as usize * 12);
```
9names, line 279: These pointers are likely to be unsound (due to pointing outside of the original object).
I'm not going to block merging based on this, but I'll raise an issue so we can fix up the PAC to make these safe

PR #149, opened 2021-10-02
Fix qt_py_rainbow.rs

jannic:
Add RESET argument to Timer::new call


jannic:
Also fixed by https://github.com/rp-rs/rp-hal/pull/150 so this PR s no longer needed.

PR #151, opened 2021-10-02
Add simple interrupt_disable+spinlock critical-section impl

9names:
Implement [critical section's custom-impl trait](https://github.com/embassy-rs/critical-section/blob/a9d071fbb55720b8017b8dafa8f8fad4667715bb/src/lib.rs#L36) based (heavily) on examples and feedback from dirbaio and jannic.

Interrupts are disabled before acquiring the spinlock to ensure that an interrupt on the same core does not deadlock.
The interrupt-enable on failure to acquire the lock ensures that interrupts still have a chance to fire on the same core if main() is blocked on the critical-section.
Performance for this implementation is not expected to be great.

This assumes that spinlock31 is reserved exclusively for this purpose.

This should not be merged before https://github.com/rp-rs/rp2040-pac/pull/39 to avoid churn.

Definitely squash when you do perform the merge, we don't need the history from this branch.


rp2040-hal/src/critical_section_impl.rs
```diff
@@ -0,0 +1,75 @@
...
+            LOCK_ALREADY_OWNED
+        } else {
+            // Spin until we get the lock
+            loop {
+                // Need to disable interrupts to ensure that we will not deadlock
+                // if an interrupt enters critical_section::Impl after we acquire the lock
+                cortex_m::interrupt::disable();
+                // Read the spinlock reserved for `critical_section`
+                if (*pac::SIO::ptr()).spinlock31.read().bits() != 0 {
+                    // We just acquired the lock.
```
jannic, line 43: I think a `core::sync::atomic::compiler_fence(Ordering::SeqCst)` needs to be added here and just before the spinlock is released, below.
Otherwise, the compiler is allowed to re-order instructions in an invalid way.
Then, the ordering of the accesses to LOCK_OWNER could be changed to `Ordering::Relaxed`, as proper synchronization between the cores is guaranteed by the spinlock. (There can still be concurrent reads and writes due to the read at the beginning of the function, which is not covered by the spinlock. So we need the atomic access. But in that case, the actual value read doesn't matter, so `Ordering::Relaxed` will be sufficient.)

rp2040-hal/src/critical_section_impl.rs
```diff
@@ -0,0 +1,79 @@
...
+const LOCK_ALREADY_OWNED: u8 = 2;
+
+unsafe impl critical_section::Impl for RpSpinlockCs {
+    unsafe fn acquire() -> u8 {
+        // Store the initial interrupt state and current core id in stack variables
+        let interrupts_active = cortex_m::register::primask::read().is_active();
+        // We reserved 0 as our `LOCK_UNOWNED` value, so add 1 to core_id so we get 1 for core0, 2 for core1.
+        let core = (*pac::SIO::ptr()).cpuid.read().bits() as u8 + 1_u8;
+        // Do we already own the spinlock?
+        if LOCK_OWNER.load(Ordering::Acquire) == core {
```
jannic, line 31: This ordering could be Relaxed.

jannic:
LGTM, let's merge this.

Some possible future improvements:
- The current implementation will wait for the spinlock with interrupts disabled. Interrupt latency may be improved if we could avoid that. However without atomic `compare_exchange`, it's probably not possible to guarantee that we never wait for the spinlock with interrupts disabled.
- The critical section, as implemented, does (formally) not guarantee any ordering between the cores. It only guarantees that only one thread of execution can enter the critical section at any time. A Mutex based on this critical section might need to add such an ordering. (In theory. I don't think it actually matters with a cortex-m0+ CPU.)
- We can decide later if we want to use sio::Spinlock

jannic:
One additional remark:

> The critical section, as implemented, does (formally) not guarantee any ordering between the cores. It only guarantees that only one thread of execution can enter the critical section at any time. A Mutex based on this critical section might need to add such an ordering. (In theory. I don't think it actually matters with a cortex-m0+ CPU.)

This actually matches the description of critical sections in https://docs.rs/critical-section/0.2.5/critical_section/struct.CriticalSection.html:

> Moreover, the caller must use adequate fences to prevent the compiler from moving the instructions inside the critical section to the outside of it. Sequentially consistent fences are suggested immediately after entry and immediately before exit from the critical section.

9names:
> LGTM, let's merge this.

Sounds good to me!

PR #158, opened 2021-10-05
Each BSP supplies its own memory.x file.

jonathanpallant:
There's also a default in RP2040-HAL to make the examples build.


jannic:
What happens if a binary crate (which depends on rp-hal) brings it's own memory.x? Who gets priority?
Example: https://github.com/rp-rs/rp2040-project-template

jonathanpallant:
This is a good question - it depends on linker and the order in which the linker search path arguments are supplied on the command line.

Mainly though this issue comes up with rp2040-hal including a memory.x (which it must to ensure the examples compile), as well as the BSPs. If you want to include your own memory.x, you probably don't want to use a BSP.

9names (approved):
LGTM. I'll create an issue about looking into the scenario Jannic described to ensure we cover that use-case

PR #159, opened 2021-10-05
Make pio-proc a dev-dependency

jannic:
It's only used in an example, so it doesn't need to be a regular
dependency.


9names (approved):
LGTM

PR #160, opened 2021-10-05
WIP: Add picotool compatible 'binary info' metadata

jonathanpallant:
This is a WIP proof of concept. I don't suggest you merge it.

TODO:
* [x] Move 'binary_info' stuff into a module in the HAL (or maybe a crate of its own?)
* [ ] Make some macros so its easier to put the metadata in your program
* [ ] Instead of making the user create a table of entry references, get the macros to stuff a reference to every entry created into a separate section - that section is then what gets the start/end stuffed into the header.
* [ ] Support pin descriptions

I'd like to see this get to:

```rust
binary_info! {
    program_name: "My silly program",
    program_version: env!("CRATE_VERSION"),
};
```

How you would get pin configuration info from the BSP and into your application is TBD.


thejpster:
TIL the paste macro exists. Might be useful here.

thejpster:
Closed in favour of https://github.com/rp-rs/rp-hal/pull/824

PR #162, opened 2021-10-08
Implement peripheral support for i2c and an advanced example

ithinuel:
The PR features:

* an implementation for asynchronous API for I2C as bus controller
* an implementation for asynchronous API for I2C as bus peripheral
* an advanced example that shows the use of read/write for both controller & peripheral.

Open questions:

* Is `nostd_async` an appropriate executor? (it only works on stable toolchain)
* Shall the peripheral side of the demo validate the state transition (as implemented) or simply serve the requests?
* Shall we implement the `embassy-trait` for i2c? The proposed API & the embassy APIs are basically async adaptation of the eh's blocking hal.

### DIFF
--- a/boards/pico/Cargo.toml
+++ b/boards/pico/Cargo.toml
@@ -25,7 +25,7 @@ cortex-m-rtic = "0.6.0-alpha.5"
 rp2040-boot2 = "0.2"
 nb = "1.0"
 i2c-pio = { git = "https://github.com/ithinuel/i2c-pio-rs", rev = "fb6167d02b7fbc46a83f344f5242823bcd16e271" }
-nostd_async = { git = "https://github.com/sammhicks/nostd_async", features = ["cortex_m"] }
+nostd_async = { git = "https://github.com/sammhicks/nostd_async", rev = "6f170891e7f4d4f6772974bf4c02007a23b9ef26", features = ["cortex_m"] }
 
 [features]
 default = ["rt"]
--- a/boards/pico/Cargo.toml
+++ b/boards/pico/Cargo.toml
@@ -25,6 +25,7 @@ cortex-m-rtic = "0.6.0-alpha.5"
 rp2040-boot2 = "0.2"
 nb = "1.0"
 i2c-pio = { git = "https://github.com/ithinuel/i2c-pio-rs", rev = "fb6167d02b7fbc46a83f344f5242823bcd16e271" }
+nostd_async = { git = "https://github.com/sammhicks/nostd_async", features = ["cortex_m"] }
 
 [features]
 default = ["rt"]
--- a/boards/pico/Cargo.toml
+++ b/boards/pico/Cargo.toml
@@ -18,8 +18,19 @@ usb-device= "0.2.8"
 usbd-serial = "0.1.1"
 usbd-hid = "0.5.1"
 futures = { version = "0.3", default-features = false, optional = true }
-embassy = { git = "https://github.com/embassy-rs/embassy", optional = true }
-embassy_traits = { git = "https://github.com/embassy-rs/embassy", package = "embassy-traits", optional = true }
+
+[dependencies.embassy]
+git = "https://github.com/embassy-rs/embassy"
+rev = "a8797f84f69b7668a3f89b6cba3e39bce5649079"
+optional = true
+
+# namespaced features will let use use "dep:embassy-traits" in the features rather than using this
+# trick of renaming the crate.
+[dependencies.embassy_traits]
+git = "https://github.com/embassy-rs/embassy"
+rev = "a8797f84f69b7668a3f89b6cba3e39bce5649079"
+package = "embassy-traits"
+optional = true
 
 [dev-dependencies]
 panic-halt= "0.2.0"
--- a/boards/pico/Cargo.toml
+++ b/boards/pico/Cargo.toml
@@ -17,6 +17,9 @@ embedded-time = "0.12.0"
 usb-device= "0.2.8"
 usbd-serial = "0.1.1"
 usbd-hid = "0.5.1"
+futures = { version = "0.3", default-features = false, optional = true }
+embassy = { git = "https://github.com/embassy-rs/embassy", optional = true }
+embassy_traits = { git = "https://github.com/embassy-rs/embassy", package = "embassy-traits", optional = true }
 
 [dev-dependencies]
 panic-halt= "0.2.0"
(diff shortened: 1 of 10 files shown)


rp2040-hal/src/i2c/peripheral.rs
```diff
@@ -0,0 +1,211 @@
...
+                i2c,
+                pins: (sda_pin, scl_pin),
+                mode: PhantomData,
+            },
+            state: State::Idle,
+        }
+    }
+}
+
+#[derive(Debug)]
```
jonathanpallant, line 93: Could we move these `enum` and `struct` types up to the top of the file? I think it makes them easier to find.

rp2040-hal/Cargo.toml
```diff
@@ -24,6 +24,7 @@ usb-device = "0.2.8"
 vcell = "0.1"
 void = { version = "1.0.2", default-features = false }
 rand_core = "0.6.3"
+futures = { version = "*", default-features = false }
```
jonathanpallant, line 27: This should probably have a version number - I think they're at "0.3".

boards/pico/Cargo.toml
```diff
@@ -24,6 +24,7 @@ embedded-hal ="0.2.5"
 cortex-m-rtic = "0.6.0-alpha.5"
 rp2040-boot2 = "0.1.2"
 nb = "1.0"
+nostd_async = { git = "https://github.com/sammhicks/nostd_async", features = ["cortex_m"] }
```
jonathanpallant, line 27: Can we pin this to a specific revision?

boards/pico/examples/pico_i2c_controller_peripheral/controller.rs
```diff
@@ -0,0 +1,38 @@
+use super::ADDRESS;
```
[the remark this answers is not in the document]
jonathanpallant: Uh, I now see these are modules for one large example. But I think the same comment applies!

jonathanpallant (dismissed):

rp2040-hal/src/i2c/peripheral.rs
ithinuel: I usually try to keep the structure of the file as 
- impl-for-extern/parent-types
- repeated for each local type t:
  - type t
  - impl for type t

but I'm ok to move them at the top too.

boards/pico/examples/pico_i2c_controller_peripheral/main.rs
```diff
@@ -0,0 +1,109 @@
+//! Send data from an I2C controller to an I2C peripheral.
+//!
+//! This will transmit data between I2C0 on pins GP0, GP1 acting as a controller to I2C1 on pins
+//! GP2, GP3 acting as a peripheral.
+//!
+//! A pull up is required on SCL & SDA lines in order to reach the expected 1MHz. Although it
+//! depends on the hardware context (wire length, impedance & capacitance), a typical value of 2KOhm
+//! should generally work fine.
```
9names, line 8: The comment block on top of controller and peripheral are pretty good, but I think the main.rs is a little brief.
You asked for help with comments, but I think it might be better if I just write an alternative file comment so you can copy what you like from it.

```suggestion
//! I2C controller and I2C peripheral async demo.
//! 
//! This example demonstrates use of both I2C peripherals (I2C0 and I2C1) at the same time on a single Pico using [Embassy](https://github.com/embassy-rs/embassy), an async executor.
//! 
//! Each peripheral is passed to an async task, which allows them to operate independently of each other:
//!   - The controller task (ctrl_demo) uses I2C0. It calls the demo controller code in `controller.rs`
//!   - The peripheral task (prph_demo) uses I2C1. It calls the demo peripheral code in `peripheral.rs`
//! 
//! ### Wiring notes:
//! 
//! I2C0 uses pin `GP0` for `SDA`, and `GP1` for `SCL`.
//! 
//! I2C1 uses `GP2` for `SDA`, and `GP3` for `SCL`.
//! 
//! For this demo to function you must connect the `SDA` signals (`GP0` and `GP2`) to each other using wires.
//! You must also connect the `SCL` signals (`GP1` and `GP3`) to each other.
//! 
//! A pull up resistor (to 3.3V, which is available on pin `36`) is required on SCL & SDA lines in order to reach the expected 1MHz. Although it
//! depends on the hardware context (wire length, impedance & capacitance), a typical value of 2KOhm
//! should generally work fine.
//! 
//! If you do not connect the resistor and instead use the internal pull-ups on the I2C pins, you may need to lower the I2C frequency to avoid transmission errors.
```

rp2040-hal/src/i2c.rs
```diff
@@ -116,22 +121,100 @@ impl SclPin<I2C0> for Gpio21 {}
...
     i2c: I2C,
     pins: Pins,
+    mode: PhantomData<Mode>,
 }
 
 const TX_FIFO_SIZE: u8 = 16;
+const RX_FIFO_SIZE: u8 = 16;
 
-fn i2c_reserved_addr(addr: u8) -> bool {
+fn i2c_reserved_addr(addr: u16) -> bool {
```
9names, line 152: Is the type change here so we can support 10bit addressing in the future?

rp2040-hal/src/i2c/controller.rs
```diff
@@ -0,0 +1,302 @@
...
+                .write(|w| w.ic_fs_scl_lcnt().bits(lcnt as u16));
+            i2c.ic_fs_spklen.write(|w| {
+                w.ic_fs_spklen()
+                    .bits(if lcnt < 16 { 1 } else { (lcnt / 16) as u8 })
+            });
+            i2c.ic_sda_hold
+                .modify(|_r, w| w.ic_sda_tx_hold().bits(sda_tx_hold_count as u16));
+        }
+
+        // Enable IP
```
9names, line 108: I'd prefer `I2C block` or `I2C peripheral` over `IP` here for clarity

rp2040-hal/src/i2c/peripheral.rs
```diff
@@ -0,0 +1,199 @@
...
+            w.rx_fifo_full_hld_ctrl().enabled();
+            w.ic_restart_en().enabled();
+            w.tx_empty_ctrl().enabled()
+        });
+
+        // Clear FIFO threshold
+        i2c.ic_tx_tl.write(|w| unsafe { w.tx_tl().bits(0) });
+        i2c.ic_rx_tl.write(|w| unsafe { w.rx_tl().bits(0) });
+
+        // Enable IP
```
9names, line 86: I'd prefer `I2C block` or `I2C peripheral` over `IP` here for clarity

rp2040-hal/src/i2c/peripheral.rs
```diff
@@ -0,0 +1,199 @@
...
+                self.state = if stat.rd_req().bit_is_set() {
+                    State::Read
+                } else {
+                    State::Write
+                };
+                Some(I2CEvent::Start)
+            }
+            State::Read if stat.rd_req().bit_is_set() => {
+                // Bit is cleared by a call to write
+                // self.0.i2c.ic_clr_rd_req.read();
```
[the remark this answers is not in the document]
ithinuel: No this is due to how the hardware is made. On a read request, clearing this bit indicates the I2C Block that data has been pushed to the FIFO and it can stop clock-stretching.
I'll update the comment to make that more obvious.

rp2040-hal/src/i2c.rs
ithinuel: Indeed. For now this is mostly to save us from unnecessary conversion between u8 & u16, since the register takes u16 either way, we may as well just use u16 internally.

rp2040-hal/src/i2c/peripheral.rs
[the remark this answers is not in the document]
ithinuel: Updated, let me know what you think about this new version.

9names (approved):
LGTM

rp2040-hal/src/i2c/peripheral.rs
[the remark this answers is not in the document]
9names: Yep much clearer, thanks!

PR #167, opened 2021-10-12
Implements methods to allow presetting the pin state & direction

ithinuel:
Enabling those methods allows to save a few valuable instructions in the PIO's memory.

### DIFF
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -372,6 +372,50 @@ pub struct UninitStateMachine<SM: ValidStateMachine> {
 unsafe impl<SM: ValidStateMachine + Send> Send for UninitStateMachine<SM> {}
 
 impl<SM: ValidStateMachine> UninitStateMachine<SM> {
+    /// Sets unmasked pin directions
+    ///
+    /// The iterator's item are pairs of `(pin_number, pin_high)`.
+    pub fn set_pins_with_iter(&mut self, pins: impl Iterator<Item = (u8, bool)>) {
+        let saved_ctrl = self.sm().sm_pinctrl.read();
+        for (pinnum, pinstate) in pins {
+            self.sm()
+                .sm_pinctrl
+                .write(|w| unsafe { w.set_base().bits(pinnum).set_count().bits(1) });
+            self.set_instruction(
+                pio::InstructionOperands::SET {
+                    destination: pio::SetDestination::PINS,
+                    data: if pinstate { 1 } else { 0 },
+                }
+                .encode(),
+            );
+        }
+        self.sm()
+            .sm_pinctrl
+            .write(|w| unsafe { w.bits(saved_ctrl.bits()) });
+    }
+
+    /// Set pin directions.
+    ///
+    /// The iterator's item are pairs of `(pin_number, as_output)`.
+    pub fn set_pindirs_with_iter(&mut self, pindirs: impl Iterator<Item = (u8, bool)>) {
+        let saved_ctrl = self.sm().sm_pinctrl.read();
+        for (pinnum, as_output) in pindirs {
+            self.sm()
+                .sm_pinctrl
+                .write(|w| unsafe { w.set_base().bits(pinnum).set_count().bits(1) });
+            self.set_instruction(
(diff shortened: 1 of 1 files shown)


rp2040-hal/src/pio.rs
```diff
@@ -372,6 +372,50 @@ pub struct UninitStateMachine<SM: ValidStateMachine> {
 unsafe impl<SM: ValidStateMachine + Send> Send for UninitStateMachine<SM> {}
 
 impl<SM: ValidStateMachine> UninitStateMachine<SM> {
+    /// Sets unmasked pin directions
+    ///
+    /// The iterator's item are pairs of `(pin_number, pin_high)`.
+    pub fn set_pins_with_iter(&mut self, pins: impl Iterator<Item = (u8, bool)>) {
```
henkkuli, line 378: I think `IntoIterator` would be more flexible. Then one could use this as
```rust
sm.set_pins_with_iter([(1, false), (5, true)]);
```

ithinuel: Good point, that indeed makes it much nicer to use!

ithinuel:
@henkkuli 
> I'm thinking whether the type `(u8, bool)` is good to use in both places, or whether we should define enums like `PinState { Low, High }` and `PinDir { Input, Output }` for this API. Those could then implement `From<bool>` if someone wants to use `true` and `false` instead.

I would lean towards that too but I didn't want to deviate too much from the c-sdk that uses a mask. Writing this I also realize we already have a `set_pindirs_with_mask` that requires the state machine to be initialized but stopped.

For a bit of context I'm implementing the i2c support and the [c-sdk](https://github.com/raspberrypi/pico-examples/blob/master/pio/i2c/i2c.pio#L97-L109) has some pre-init code to avoid generating glitches on the bus when the PIO takes over.

rp2040-hal/src/pio.rs
```diff
@@ -360,6 +360,24 @@ impl<P: PIOExt, SM: StateMachineIndex> ValidStateMachine for (P, SM) {
     }
 }
 
+/// Pin State
+///
+/// Note this may not be the same as its electrical state (e.g. if invert is enabled on the pin).
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[allow(missing_docs)]
```
henkkuli, line 367: Why missing docs? Could you just give short doc for enum variants, say "Pin is in high state".

rp2040-hal/src/pio.rs
```diff
@@ -360,6 +360,24 @@ impl<P: PIOExt, SM: StateMachineIndex> ValidStateMachine for (P, SM) {
...
+pub enum PinState {
+    Active,
```
henkkuli, line 369: Would `High` and `Low` be more idiomatic?

rp2040-hal/src/pio.rs
ithinuel: It seems redundant to document `PinState::Active` with
> Pin in active state.

But I can of course add those line.

rp2040-hal/src/pio.rs
```diff
@@ -500,56 +522,58 @@ impl<SM: ValidStateMachine> StateMachine<SM, Stopped> {
...
-                });
-                let set_pindirs = pio::Instruction {
-                    operands: pio::InstructionOperands::SET {
-                        destination: pio::SetDestination::PINDIRS,
-                        data: ((pindir >> pin) & 0x1) as u8,
-                    },
-                    delay: 0,
-                    side_set: None,
+    /// The iterator's item are pairs of `(pin_number, pin_state)`.
+    pub fn set_pins_with_iter(&mut self, pins: impl IntoIterator<Item = (u8, PinState)>) {
```
jannic, line 531: I wondered why the method includes the suffix `_with_iter` - perhaps to distinguish it from `PIOBuilder::set_pins`?
In that case I'd expect that both methods do basically the same, just with different parameters. Which is not the case, they do fundamentally different things.
If it is because you expect that there may be different variants of this method in the future, the suffix is fine.
(I'd prefer to have a stronger distinction from `PIOBuilder::set_pins`, but have no idea for a better name, either.)

PR #169, opened 2021-10-17
Fix irq reg accesses

### LINKED ISSUE #161: GPIO interrupt register accesses may be unsound
The pointers used to access GPIO interrupt registers are likely to be unsound (due to pointing outside of the original object).
Check if the original object in the PAC is an array, and if it isn't, convert it to an array then update src/gpio/reg.rs functions to index into it.

_Originally posted by @9names in https://github.com/rp-rs/rp-hal/pull/147#discussion_r723732755_

9names:
The code from PR #147 used raw pointers to get/set interrupt registers.
The registers in the PAC needed to be converted to arrays to make it simpler to implement this, so there is a separate PR for the PAC that this PR depends on: https://github.com/rp-rs/rp2040-pac/pull/43
Closes #161

### DIFF
--- a/rp2040-hal/src/gpio/reg.rs
+++ b/rp2040-hal/src/gpio/reg.rs
@@ -272,7 +272,7 @@ pub(super) unsafe trait RegisterInterface {
     fn interrupt_status(&self, interrupt: Interrupt) -> bool {
         let num = self.id().num as usize;
         unsafe {
-            let cpuid = *(pac::SIO::ptr() as *const u32);
+            let cpuid = (*pac::SIO::ptr()).cpuid.read().bits();
             let io = &(*pac::IO_BANK0::ptr());
             let reg = (&io.proc0_ints0)
                 .as_ptr()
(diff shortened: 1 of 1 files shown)


9names:
Note that this is a port of the existing code. I'm not certain, for example, that the register clearing logic is correct.
The Datasheet is not very clear about how to clear the interrupts. It states:
```

9names:
Oops, didn't mean to close this.

jannic:
> It is not stated what you should write to the register. Other parts of the datasheet mention writing a bitmask, again, not clear if that's a bitmask with only the bit you care about set. The code currently does a load + bitwise OR the interrupt + write. This will probably work, but may also clear other interrupts that weren't intended to be cleared. I cannot find in the C SDK where they handle this.

The bits are listed in the register table to have a type of "WC", which is explained in Appendix A (Register Field Types): "The bit is cleared by writing a 1, using either a normal write or the clear alias."

PR #170, opened 2021-10-17
Use volatile register access for gpio interrupts

jannic:
This addresses the concerns raised in https://github.com/rp-rs/rp-hal/issues/161#issuecomment-945074409, however still doing direct register access and not using the PAC read/write abstraction.
I didn't find a good (readable, performant) way to use the PAC. Additionally, as cortex-m0+ doesn't provide atomic read-modify-write operations, the writing methods need to use the atomic access aliases to avoid data races.

### DIFF
--- a/rp2040-hal/src/gpio/reg.rs
+++ b/rp2040-hal/src/gpio/reg.rs
@@ -6,6 +6,7 @@ use super::{
 };
 use crate::gpio::dynpin::{DynDisabled, DynFunction, DynInput, DynOutput, DynPinMode};
 use crate::pac;
+use core::ptr::{read_volatile, write_volatile};
 
 //==============================================================================
 //  ModeFields
@@ -278,7 +279,7 @@ pub(super) unsafe trait RegisterInterface {
                 .as_ptr()
                 .add(num / 8 + cpuid as usize * 12);
             let bit_in_reg = num % 8 * 4 + interrupt as usize;
-            (*reg & (1 << bit_in_reg)) != 0
+            (read_volatile(reg) & (1 << bit_in_reg)) != 0
         }
     }
 
@@ -293,7 +294,7 @@ pub(super) unsafe trait RegisterInterface {
                 .as_ptr()
                 .add(num / 8 + cpuid as usize * 12);
             let bit_in_reg = num % 8 * 4 + interrupt as usize;
-            (*reg & (1 << bit_in_reg)) != 0
+            (read_volatile(reg) & (1 << bit_in_reg)) != 0
         }
     }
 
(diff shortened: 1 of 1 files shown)


rp2040-hal/src/atomic_register_access.rs
```diff
@@ -0,0 +1,56 @@
...
+///
+/// [section_2_1_2]: https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf#atomic-rwtype
+///
+/// # Safety
+///
+/// In addition to the requirements of [core::ptr::write_volatile],
+/// `register` must point to a register providing atomic aliases.
+#[inline]
+pub unsafe fn write_xor(register: *mut u32, bits: u32) {
+    let alias = (register as usize + 0x1000) as *mut u32;
```
jonathanpallant, line 22: `alias.offset(0x400)` ?

rp2040-hal/src/atomic_register_access.rs
```diff
@@ -0,0 +1,56 @@
...
+///
+/// [section_2_1_2]: https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf#atomic-rwtype
+///
+/// # Safety
+///
+/// In addition to the requirements of [core::ptr::write_volatile],
+/// `register` must point to a register providing atomic aliases.
+#[inline]
+pub unsafe fn write_bitmask_set(register: *mut u32, bits: u32) {
+    let alias = (register as usize + 0x2000) as *mut u32;
```
jonathanpallant, line 38: `alias.offset(0x800)`?

rp2040-hal/src/atomic_register_access.rs
```diff
@@ -0,0 +1,56 @@
...
+///
+/// [section_2_1_2]: https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf#atomic-rwtype
+///
+/// # Safety
+///
+/// In addition to the requirements of [core::ptr::write_volatile],
+/// `register` must point to a register providing atomic aliases.
+#[inline]
+pub unsafe fn write_bitmask_clear(register: *mut u32, bits: u32) {
+    let alias = (register as usize + 0x3000) as *mut u32;
```
jonathanpallant, line 54: `alias.offset(0xC00)`?

jonathanpallant (commented):
Because `.offset()` works in U32 sized units, not sure if it's clearer than the usize-cast-add-pointer-cast, which gets to use the offsets given in the datasheet.

rp2040-hal/src/atomic_register_access.rs
9names: alias.offset(0x400).write_volatile(bits)?

jannic: It may be useful to allow for 8 or 16 bit writes, yes. But for now, we don't need it. As I changed the methods to `pub(crate)` visibility, they are not part of the API and therefore can be easily updated when the actual need for such writes arises.

PR #175, opened 2021-10-19
API to enable/disable interrupts for each PWM channel.

contradict:
I'm not sure this is the correct way to do this, but I would like to be able to configure the interrupt registers of the PWM peripheral.

It seems like the Interrupt struct should take ownership of some data to enforce its uniqueness, but I'm not sure what that should be.

### DIFF
--- a/rp2040-hal/src/pwm/mod.rs
+++ b/rp2040-hal/src/pwm/mod.rs
@@ -402,7 +402,7 @@ where
 
     /// Did this slice trigger an overflow interrupt?
     #[inline]
-    pub fn is_overflow(&self) -> bool {
+    pub fn has_overflown(&self) -> bool {
         let mask = self.bitmask();
         unsafe { (*pac::PWM::ptr()).ints.read().bits() & mask == mask }
     }


rp2040-hal/src/pwm/mod.rs
```diff
@@ -373,6 +373,147 @@ where
...
+
+impl Interrupt {
+    /// Enable the interrupt for the indicated channel.
+    pub fn enable_interrupt<S: SliceId, M: SliceMode + ValidSliceMode<S>>(
+        &mut self,
+        _: &Slice<S, M>,
+    ) {
+        match S::DYN {
+            DynSliceId { num } if num == 0 => unsafe {
+                (*pac::PWM::ptr()).inte.write(|w| w.ch0().set_bit())
```
henkkuli, line 387: This will clear all other bits in the register. The correct method would be
```suggestion
                (*pac::PWM::ptr()).inte.modify(|_, w| w.ch0().set_bit())
```
In practice I'd like to see this using the atomic register access instead. This way parts of the interrupt could be sent between threads.

rp2040-hal/src/pwm/mod.rs
```diff
@@ -373,6 +373,147 @@ where
...
+    }
+
+    /// Clear the overflow bit associated with this slice.
+    pub fn clear_overflow<S: SliceId, M: SliceMode + ValidSliceMode<S>>(
+        &mut self,
+        _: &Slice<S, M>,
+    ) {
+        match S::DYN {
+            DynSliceId { num } if num == 0 => unsafe {
+                (*pac::PWM::ptr()).intr.write(|w| w.ch0().clear_bit())
```
henkkuli, line 489: The `INTR` register has been marked with type `WC` meaning that in order to clear the interrupt bit, 1 should be written. I.e. this should be
```suggestion
                (*pac::PWM::ptr()).intr.write(|w| w.ch0().set_bit())
```
to write only the bit we want to clear.

rp2040-hal/src/pwm/mod.rs
```diff
@@ -373,6 +373,147 @@ where
...
+    }
+
+    /// Disable the interrupt for the indicated slice.
+    pub fn disable_interrupt<S: SliceId, M: SliceMode + ValidSliceMode<S>>(
+        &mut self,
+        _: &Slice<S, M>,
+    ) {
+        match S::DYN {
+            DynSliceId { num } if num == 0 => unsafe {
+                (*pac::PWM::ptr()).inte.write(|w| w.ch0().clear_bit())
```
henkkuli, line 421: Same as above

henkkuli:
Looks good, though this can't be merged without atomic access. This is because the `Slice` implements `Send` and hence can be sent to a different core or to an interrupt handler, but the interrupt enable and disable methods modify a shared register in a non-atomic way. Hope that #170 gets merged soon.

rp2040-hal/src/pwm/mod.rs
```diff
@@ -371,6 +371,38 @@ where
...
+        unsafe {
+            (*pac::PWM::ptr())
+                .inte
+                .modify(|r, w| w.bits(r.bits() & !(1 << I::DYN.num)))
+        };
+    }
+
+    /// Did this slice trigger an overflow interrupt?
+    #[inline]
+    pub fn is_overflow(&mut self) -> bool {
```
henkkuli, line 397: No need for unique access:
```suggestion
    pub fn is_overflow(&self) -> bool {
```

contradict: Good call, fixed.

PR #177, opened 2021-10-20
Mark PIO StateMachine, Rx and Tx as Send

contradict:


rp2040-hal/src/pio.rs
```diff
@@ -487,6 +487,9 @@ impl<SM: ValidStateMachine, State> StateMachine<SM, State> {
     }
 }
 
+/// Safety: All shared register access is atomic.
```
ithinuel, line 490:
```suggestion
/// Safety: All shared register accesses are atomic.
```

As far as I can see, this is true for now but I think this requires an extensive explanation as to why those accesses are atomic and/or a reminder at each access about their requirement to be atomic.

ithinuel (changes requested):

PR #179, opened 2021-10-20
Add PIO features required for PIO as I2C support

ithinuel:
This PR implements extra features on the PIO peripheral to allow for the implementation of the PIO as I2C support.

This PR also provides a fix for the inconsistency between type-state on startup after soft/warm resets (eg debugger, watchdog, bootloader). Note this fix is currently only applied on GPIO & PIO but is likely to be required on other peripherals.

Preceding PR:

- https://github.com/rp-rs/rp-hal/pull/167

### DIFF
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -542,7 +542,7 @@ impl<SM: ValidStateMachine> StateMachine<SM, Stopped> {
     /// other state machines of the same PIO block.
     ///
     /// The iterator's item are pairs of `(pin_number, pin_state)`.
-    pub fn set_pins_with_iter(&mut self, pins: impl IntoIterator<Item = (u8, PinState)>) {
+    pub fn set_pins(&mut self, pins: impl IntoIterator<Item = (u8, PinState)>) {
         let saved_ctrl = self.sm.sm().sm_pinctrl.read();
         for (pin_num, pin_state) in pins {
             self.sm
@@ -569,7 +569,7 @@ impl<SM: ValidStateMachine> StateMachine<SM, Stopped> {
     /// other state machines of the same PIO block.
     ///
     /// The iterator's item are pairs of `(pin_number, pin_dir)`.
-    pub fn set_pindirs_with_iter(&mut self, pindirs: impl IntoIterator<Item = (u8, PinDir)>) {
+    pub fn set_pindirs(&mut self, pindirs: impl IntoIterator<Item = (u8, PinDir)>) {
         let saved_ctrl = self.sm.sm().sm_pinctrl.read();
         for (pinnum, pin_dir) in pindirs {
             self.sm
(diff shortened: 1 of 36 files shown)


rp2040-hal/src/gpio/pin.rs
```diff
@@ -959,9 +959,13 @@ macro_rules! gpio {
                 impl Pins {
                     /// Take ownership of the PAC peripherals and SIO slice and split it into discrete [`Pin`]s
                     pub fn new(io : [<IO_ $Group:upper>], pads: [<PADS_ $Group:upper>], sio: [<SioGpio $Group>], reset : &mut pac::RESETS) -> Self {
+                        pads.reset_bring_down(reset);
+                        io.reset_bring_down(reset);
```
jannic, line 963: I still think there should be a way to initialize HAL devices without a full peripheral reset, especially for GPIO where reconfigurations could introduce unnecessary glitches.
However, as a default, those resets are correct.
Special, potentially unsafe, alternatives to create those objects without resets can be added later, if actually needed.

ithinuel: I do think so too. As discussed over matrix, a solution could be to introduce an extra state for the peripherals and let the user choose between the safe (reset) way or the unsafe shortcut to the state they want. (In some cases we could just use things like DynPin an allow to move into a specific state).

But in the mean time i think we should address that rather important inconsistency as more using will be confused by the result they get from resetting/reflashing their development target.

rp2040-hal/src/pio.rs
```diff
@@ -656,22 +726,89 @@ pub struct Tx<SM: ValidStateMachine> {
...
+            let reg_ptr = self.block().txf[SM::id()].as_ptr() as *mut T;
+            core::ptr::write_volatile(reg_ptr, value);
+        }
 
         true
     }
+
+    /// Checks if the state machine has stalled on empty TX FIFO during a blocking PULL, or an OUT
+    /// with autopull enabled.
+    pub fn has_stalled(&self) -> bool {
```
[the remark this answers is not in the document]
jannic: > I'd prefer devsnek's suggestion over just "stalled". it's also more accurate here - has_stalled could mean that it may or may not currently be stalled, but has stalled some time since you last asked/cleared a flag.
Isn't that an argument in favor of `has_stalled`?
The flag is documented as "State machine has stalled on empty TX FIFO during ablocking PULL, or an OUT with autopull enabled. Write 1 to clear."
So its meaning is, as you wrote, "that it may or may not currently be stalled, but has stalled some time since you last asked/cleared a flag".

rp2040-hal/src/pio.rs
```diff
@@ -633,19 +689,33 @@ pub struct Rx<SM: ValidStateMachine> {
 }
 
 impl<SM: ValidStateMachine> Rx<SM> {
+    fn block(&self) -> &pac::pio0::RegisterBlock {
```
jannic, line 692: I know we already have a `fn block` in pio.rs, but I don't like the name. It sounds too much like a command. (Block for something, like in `nb::block`.)
I'd prefer, for example, `fn register_block`.

PR #187, opened 2021-10-29
Fix UART rx

jannic:


9names:
Context for this PR from rp-rs channel on matrix chat:

> bzee: 
I'm curious to know if anyone else has problems receiving UART on the Pico. I've tried a sensor and USB to serial, but I've ran out of debugging ideas...
I am either not getting the right bytes, or I am getting errors, like Break or Framing errors... I am at the point of thinking something is wrong with the UART driver.

> jannic:
The old code read the data register repeatedly, even when the FIFO was already empty. I don't know (and don't find in the datasheet) what the chip does in that situation. Perhaps that could cause spurious errors? Otherwise, the additional reads 'only' cause that 4 out of 5 bytes received get dropped.

9names (approved):
LGTM

PR #188, opened 2021-10-29
Fix UART init code

jannic:
Set uartlcr_h before enabling the UART.
Writing uartlcr_h while the UART is enabled is forbidden by the datasheet.

Details can be found here:
https://developer.arm.com/documentation/ddi0183/g/programmers-model/register-descriptions/line-control-register--uartlcr-h?lang=en
(As referenced from the Datasheet)


9names (approved):
LGTM

Issue #186, opened 2021-10-28
Safety requirement must be verbosely advertised to not be overlooked in future changes


ithinuel:
This issue is a follow-up of https://github.com/rp-rs/rp-hal/pull/177

Because of that change ALL further changes in PIO must guaranty that they are atomic and won't break that contract.

The current code isn't verbose enough about that requirement and it will likely be overlooked in coming PRs especially considering that not all peripheral are marked Send and reviewers may not be aware of that requirement.

There need to be a detailed explanation about what that marker implies in terms of safety requirement on `PIO` as well as a `// SAFETY: ` tag on every access to the `PIO`'s register block.

jannic:
@ithinuel what do you think, is this issue sufficiently handled by #190 or are there still important facts missing from the comments?

ithinuel:
That seems right yes. Thank you, I'm closing the issue.

PR #190, opened 2021-10-31
Document properties needed for PIO Send safety

contradict:
Start to address #186.

#177 added the Send trait to StateMachine, Tx and Rx, That documentation was expanded as suggested. 

PIO, UninitStateMachine, Interrupt were marked Send before #177, added some documentation to those as well.


9names (approved):
LGTM, thanks!

PR #191, opened 2021-11-01
Rename set_instruction to exec_instruction

jannic:
This is meant to make the action caused by the function call more obvious. (Better suggestions are welcome!)

set_instruction sounds a little bit like it would somehow change the instruction memory of the state machine, which is not the case. All it does is forcing the state machine to immediately execute the specified instruction.

Also, update the comment regarding calling this function while the state machine is running with information from datasheet section 3.5.7.

@ithinuel, if this is merged, i2c-pio-rs needs to be updated as well, as it calls set_instruction.

### DIFF
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -525,13 +525,13 @@ impl<SM: ValidStateMachine, State> StateMachine<SM, State> {
         self.exec_instruction(instruction);
     }
 
-    ///Execute the instruction immediately.
+    /// Execute the instruction immediately.
+    /// 
+    /// While this is allowed even when the state machine is running, the datasheet says:
+    /// > If EXEC instructions are used, instructions written to INSTR must not stall.
+    /// It's unclear what happens if this is violated.
     pub fn exec_instruction(&mut self, instruction: u16) {
-        // This is allowed even if the state machine is running.
-        //
-        // However, the datasheet says:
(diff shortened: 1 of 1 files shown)


thejpster (approved):
Seems OK to me.

rp2040-hal/src/pio.rs
```diff
@@ -519,10 +519,20 @@ impl<SM: ValidStateMachine, State> StateMachine<SM, State> {
...
+    }
+
+    ///Execute the instruction immediately.
+    pub fn exec_instruction(&mut self, instruction: u16) {
+        // This is allowed even if the state machine is running.
+        //
+        // However, the datasheet says:
+        // "If EXEC instructions are used, instructions written to INSTR must not stall"
+        // It's unclear what happens if this is violated.
+        self.sm.exec_instruction(instruction);
```
ithinuel, line 535: I think this warning should be in the function's documentation.
```suggestion
    /// Execute the instruction immediately.
    /// 
    /// While this is allowed even when the state machine is running, the datasheet says:
    /// > If EXEC instructions are used, instructions written to INSTR must not stall.
    /// It's unclear what happens if this is violated.
    pub fn exec_instruction(&mut self, instruction: u16) {
        // TODO: clarify what happens if the instruction stalls.
        self.sm.exec_instruction(instruction);
```

ithinuel (commented):
LGTM, I left a comment but I don't have strong opinion about this, feel free to merge if that's not important.

PR #192, opened 2021-11-06
Adafruit ItsyBitsy RP2040 Board Support Package

DrewTChrist:
Here's the [pinout](https://cdn-blog.adafruit.com/uploads/2021/05/itsybitsy-rp2040-pins-1536x984.png)

With the help of ithinuel on the matrix server I was able to get the blinky and rainbow neopixel examples to work. I believe the lib.rs matches the pinout I linked above.


DrewTChrist:
I think maybe I should include the original author of the docs and examples in the Cargo.toml. I didn't actually have any hand in writing those I simply changed them to work with the board.

ithinuel (approved):
The pin mapping LGTM :)

9names (changes requested):
Could you please rename blinky.rs to itsy_bitsy_blinky.rs?
Keeping the example names unique allows us to build any example in the hal or bsp's from the root of the repo, which is really convenient.
Other than that, this just needs to be run through cargo fmt to make CI happy, and I think this is good to go.

DrewTChrist:
I've updated the name of blinky.rs and ran the code through cargo fmt

9names (approved):

9names:
Looks good to me, thanks for the PR!

PR #193, opened 2021-11-06
Use the PIOn peripherals in pin functions rather than creating an enum.

ithinuel:
This allows the `FuncionPio?` to be inferred from the `pac::PIO?` used.
This change shall be transparent for anyone using the `FunctionPio?` aliases.


9names (approved):
LGTM

PR #173, opened 2021-10-18
README: Add option --locked to cargo install elf2uf2-rs

jannic:
Without the --locked option, cargo currently tries to compile it using
clap v3.0.0-beta.5, which fails:

```
[...]
   Compiling clap v3.0.0-beta.5
   Compiling elf2uf2-rs v1.3.5
error[E0432]: unresolved import `clap::Clap`
 -->
/home/jan/.cargo/registry/src/github.com-1ecc6299db9ec823/elf2uf2-rs-1.3.5/src/main.rs:3:5
  |
3 | use clap::Clap;
  |     ^^^^^^^^^^ no `Clap` in the root
[...]
```

With --locked, it uses clap v3.0.0-beta.4, which works.

PR #196, opened 2021-11-10
Add --locked to second `cargo install elf2uf2-rs`

chinoto:
I skipped ahead in the instructions, running the other `cargo install elf2uf2-rs` that wasn't fixed in #173, and encountered the same issue.


9names (approved):
Lgtm

PR #197, opened 2021-11-10
pio: Fix unsound Send clippy warning

chinoto:
Clippy complains about the following line being unsound: [rp2040-hal/src/pio.rs:535](https://github.com/rp-rs/rp-hal/blob/f9b3d8341d9e8e83d6be89ae3692f85646bde97a/rp2040-hal/src/pio.rs#L535)

```
this implementation is unsound, as some fields in `StateMachine<SM, State>` are `!Send`
`#[warn(clippy::non_send_fields_in_send_ty)]` on by default
add bounds on type parameter `SM` that satisfy `InstalledProgram<SM::PIO>: Send`
for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#non_send_fields_in_send_ty
```

The warning can be resolved by making PIOExt, StateMachineIndex, and ValidStateMachine be supertraits of Send. This also makes adding the Send bound redundant in a few places.

### DIFF
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -11,7 +11,7 @@ const PIO_INSTRUCTION_COUNT: usize = 32;
 
 /// PIO Instance
 pub trait PIOExt:
-    core::ops::Deref<Target = rp2040_pac::pio0::RegisterBlock> + SubsystemReset + Sized
+    core::ops::Deref<Target = rp2040_pac::pio0::RegisterBlock> + SubsystemReset + Sized + Send
 {
     /// Create a new PIO wrapper and split the state machines into individual objects.
     #[allow(clippy::type_complexity)] // Required for symmetry with PIO::free().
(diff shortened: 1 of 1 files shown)


chinoto:
Is there ever a case where these traits should not require being Send? Maybe it made more sense to make the Send requirement only apply to usage with StateMachine.

thejpster:
Paging @ithinuel

ithinuel (changes requested):
Thanks @thejpster for paging in. I reproduced the lint with nightly.
Thank you @chinoto for looking into to that.

From the tests I did locally, only the first hunk (adding `Send` as a super trait to `PIOExt`) is needed to resolve this lint. This PR can be simplified.

chinoto:
If I remove the Send requirement from StateMachineIndex, I get: 
> `SM` cannot be sent between threads safely\nrequired because it appears within the type `(P, SM)`
But you're right about `ValidStateMachine` not needing it. The other changes in this branch are to get rid Send bounds made redundant by making PIOExt require Send.
(The extra pushes were because I realized I should put back the Send bound where ValidStateMachine is used)

thejpster:
Are we good to go on this one? Any other changes @all?

ithinuel (approved):

PR #198, opened 2021-11-14
Added timer alarms

### LINKED ISSUE #100: Timer module support

VictorKoenders:
This PR adds an API for the 4 build-in alarms.

To schedule an alarm:

```rust
let mut timer = Timer::new(cx.device.TIMER, &mut cx.device.RESETS);
let mut alarm0 = timer.alarm_0().unwrap();
alarm0.enable_interrupt();
alarm0.schedule(300_000.microseconds()).unwrap();

// interrupt
#[no_mangle]
pub fn TIMER_IRQ_0() {
     alarm0.clear_interrupt();
     led.toggle().unwrap();
     alarm0.schedule(300_000.microseconds()).unwrap();
}
```

Closes #100

### DIFF
--- a/rp2040-hal/src/timer.rs
+++ b/rp2040-hal/src/timer.rs
@@ -13,9 +13,6 @@ pub struct Timer {
     alarms: [bool; 4],
 }
 
-// Safety: All access is read-only.
-unsafe impl Sync for Timer {}
-
 impl Timer {
     /// Create a new [`Timer`]
     pub fn new(timer: TIMER, resets: &mut RESETS) -> Self {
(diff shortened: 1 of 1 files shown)


VictorKoenders:
I've tried calling `.schedule(0.microseconds())` and that seems to be random on if it actually triggers or not. I'm guessing this is because the time gets set to `timelr` but the clock ticks over at that point. 

Checking the datasheet nothing is mentioned on this behavior. Other chips (like the NRF52xxx) recommend setting the time a couple of ticks in the future. I could add an error if `time <= 1.microseconds()` just to be safe, what do you think?

9names:
There's 3 options from my point of view:
1: error for time < 1.microsecond as you suggest. Seems reasonable.
2: do nothing, document that user must set time sufficiently greater than timer::now(), and assume they will read this.
3: do a quick calculation of the minimum effective sleep that you could expect will survive the read+modify+write+check with maybe an interrupt handler or critical_section thrown in and still be greater than now(), and use something like
```rust
const MIN_DURATION:u32 = 100;
duration = max(duration, MIN_DURATION);
```
Most folks don't want the "hope the user does it correctly" option in the HAL, and I've seen more preference for error instead of fixing up the user's assumptions, so I think the order of preference would be 1,3,2

9names:
This PR closes #100

VictorKoenders:
Note to self: I should also add a `.disable_interrupt()` function

VictorKoenders:
I've added a check on 10 microseconds, that seems very reliable on the board I'm using.
I've also added a `.disable_interrupt()` function

rp2040-hal/src/timer.rs
```diff
@@ -3,12 +3,14 @@
...
 use crate::resets::SubsystemReset;
+use core::marker::PhantomData;
 
 /// Timer peripheral
 pub struct Timer {
     timer: TIMER,
+    alarms: [bool; 4],
 }
 
 // Safety: All access is read-only.
```
9names, line 16: This is no longer true

rp2040-hal/src/timer.rs
```diff
@@ -47,6 +52,46 @@ impl Timer {
             next_end: None,
         }
     }
+
+    /// Retrieve a reference to alarm 0. Will only return a value the first time this is called
+    pub fn alarm_0(&mut self) -> Option<Alarm0> {
+        if self.alarms[0] {
```
9names, line 58: There could be a data race on obtaining alarms in the current setup.
Acquisition with an interrupt::free would avoid races for single core, which should be enough for now

VictorKoenders: The `alarm_x` methods takes a mutable reference to `Timer`, does that not mean it should be safe for interrupts? The programmer will have to get a mutable reference and do locking themselves

VictorKoenders:
I made `AlarmX` take a mutable reference to `Timer` whenever it was writing to addresses that only gets some bits updated (and thus could conflict with other alarms)

rp2040-hal/src/timer.rs
VictorKoenders: Removed safety comment, this is a breaking change though

VictorKoenders:
Rebased onto main

jonathanpallant (approved):

jonathanpallant:
LGTM!

PR #199, opened 2021-11-15
Bump to latest ws2812 driver's revision.

ithinuel:
This change makes the ws2812 driver take ownership of the pin rather than its id.

This strengthen the relation between the LED driver and the pin used as well as simplifying it integration by letting the compiler resolve the pin id from the pin instance.

### DIFF
--- a/boards/feather_rp2040/Cargo.toml
+++ b/boards/feather_rp2040/Cargo.toml
@@ -22,7 +22,7 @@ rp2040-boot2 = "0.2"
 nb = "1.0.0"
 smart-leds = "0.3.0"
 pio = { git = "https://github.com/rp-rs/pio-rs.git", branch = "main" }
-ws2812-pio = { git = "https://github.com/ithinuel/ws2812-pio-rs", rev = "e1cdd475de20697aaa0bc58803512927094f0326" }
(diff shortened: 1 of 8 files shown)


thejpster (approved):
LGTM - but will there be a ws2812 release we can point at instead?

ithinuel:
That is possible indeed. The documentation level of the crate maybe a bit low but for a 0.x I guess it can be acceptable :)

jannic:
There is a circular dependency: ws2812-pio depends on the main branch of rp-hal, not on a release.
So in order to get this resolved, the following steps would have to happen first:

- [ ] publish pio to crates.io
- [ ] update rp-hal/rp2040-hal to depend on the released version of pio
- [ ] publish updated version of rp2040-hal
- [ ] update ws2812-pio to depend on published versions of pio and rp2040-hal
- [ ] publish updated version of ws2812-pio

ithinuel:
> So in order to get this resolved, the following steps would have to happen first:

Do you mean this PR or the release of the ws2812-pio driver?

it'd be great if we could merge in this PR and instead track the publication requirement in an issue.

jannic:
This PR is fine. It's just not sufficient to get the new versions published to crates.io.  But I didn't mean to delay this PR.

PR #200, opened 2021-11-16
PIO helper functions

thejpster:


rp2040-hal/src/pio.rs
```diff
@@ -754,6 +754,66 @@ impl<SM: ValidStateMachine> Tx<SM> {
...
+    /// | 31   | DREQ_PWM_WRAP7  |
+    /// | 32   | DREQ_I2C0_TX    |
+    /// | 33   | DREQ_I2C0_RX    |
+    /// | 34   | DREQ_I2C1_TX    |
+    /// | 35   | DREQ_I2C1_RX    |
+    /// | 36   | DREQ_ADC        |
+    /// | 37   | DREQ_XIP_STREAM |
+    /// | 38   | DREQ_XIP_SSITX  |
+    /// | 39   | DREQ_XIP_SSIRX  |
+    pub fn dreq_value(&self) -> u8 {
```
jannic, line 809: Same function could be added to Tx, as well. Only difference would be an additional `+ 4`.

rp2040-hal/src/pio.rs
```diff
@@ -754,6 +754,66 @@ impl<SM: ValidStateMachine> Tx<SM> {
         unsafe { &*self.block }
     }
 
+    /// Gets the FIFO's DMA address
+    pub fn dma_address(&self) -> u32 {
```
jannic, line 758: Why call it `dma_address`? DMA is just your use case, not a property of the returned value. I agree that the function is probably mainly useful for DMA, but it can still be confusing if somebody expects that this address is somehow special.
What about `pub fn fifo_ptr(&self)`?
And I'd return a `*mut u32`, even if that means that in the DMA use case, the caller has to do the `as u32` conversion.

jannic (approved):

PR #201, opened 2021-11-16
GPIO23 actually controls the 'power save' function on the SMPS.

thejpster:
If GPIO23 is low, the power supply is in low-power mode. But it
generates some noise. If GPIO23 is high, the power supply is less
efficient, but it generates less noise too.


boards/pico/src/lib.rs
```diff
@@ -33,7 +33,7 @@ hal::bsp_pins!(
     Gpio20 { name: gpio20 },
     Gpio21 { name: gpio21 },
     Gpio22 { name: gpio22 },
-    Gpio23 { name: bootsel },
+    Gpio23 { name: b_power_save },
```
jannic, line 36: Just curious: What does the prefix `b_` mean?
Or is it a typo and should be `n_power_save` because of negative logic, low level means power save is active?

boards/pico/examples/pico_gpio_in_out.rs
```diff
@@ -66,7 +66,7 @@ fn main() -> ! {
     let mut led_pin = pins.led.into_push_pull_output();
 
     // Our button input
-    let button_pin = pins.bootsel.into_pull_down_input();
+    let button_pin = pins.gpio15.into_pull_down_input();
```
jannic, line 69: Still doesn't fit the comment at the top of the file: "The button should cause the line to be grounded, as the input pin is pulled high internally by this example."

boards/pico/src/lib.rs
thejpster: b for bar. Same as n.

jannic (approved):

PR #204, opened 2021-11-19
Add SIO FIFO

thejpster:


rp2040-hal/src/sio.rs
```diff
@@ -60,15 +67,92 @@ impl Sio {
...
+        while self.read().is_some() {
+            // Spin until FIFO empty
+        }
+    }
+
+    /// Push to the FIFO, spinning if there's no space.
+    pub fn write_blocking(&mut self, value: u32) {
+        // We wait for the FIFO to have some space
+        while !self.is_write_ready() {
+            cortex_m::asm::wfe();
```
jannic, line 134: With this, I'd assume there would be a corresponding sev on the read side of the FIFO. Or is it implicitly asserted by the hardware, somehow?

rp2040-hal/src/sio.rs
```diff
@@ -60,15 +67,92 @@ impl Sio {
...
+    }
+
+    /// Pop from the FIFO, spinning if there's currently no data.
+    pub fn read_blocking(&mut self) -> u32 {
+        // Spin until FIFO has data
+        loop {
+            if let Some(data) = self.read() {
+                return data;
+            }
+        }
```
jannic, line 152: Why not use wfe here, like in write_blocking?

rp2040-hal/src/sio.rs
jonathanpallant: I assume bootloader code is doing a `sev`. This is lifted straight from the Pico-SDK multicore example.

rp2040-hal/src/sio.rs
jonathanpallant: It's copied from pico-sdk. It seems to be that we `sev` on write and `wfe` on read. I imagine the wait for write space is very short, because the other core should pop almost instantly.

rp2040-hal/src/sio.rs
jonathanpallant: Wait, I might have made an error.

jonathanpallant:
I think I cleared by my wfe/sev mix-ups. Is this OK now @jannic ?

jannic:
Yes, looks good.

PR #207, opened 2021-11-20
Added RtcClock DateTime and alarms

VictorKoenders:
Extended `RtcClock` to be able to set and load the `DateTime` on the chip, as per chapter 4.8 of the datasheet. Also added the ability to set an alarm. 

This code is mostly ported from the C sdk, except that I split `DateTime` into `DateTime` and `DateTimeFilter`, so we don't have to deal with the -1 logic that the C sdk has.

I haven't tested this on real hardware yet, so this PR is a draft for now.

### DIFF
--- a/rp2040-hal/src/rtc.rs
+++ b/rp2040-hal/src/rtc.rs
@@ -1,3 +1,606 @@
 //! Real Time Clock (RTC)
 // See [Chapter 4 Section 8](https://datasheets.raspberrypi.org/rp2040/rp2040_datasheet.pdf) for more details
-// TODO
+
+use crate::clocks::StoppableClock;
+use rp2040_pac::RTC;
+
+pub use crate::clocks::RtcClock;
+
+/// Structure containing date and time information
+pub struct DateTime {
+    /// 0..4095
+    pub year: u16,
+    /// 1..12, 1 is January
+    pub month: u8,
+    /// 1..28,29,30,31 depending on month
+    pub day: u8,
+    ///
+    pub day_of_week: DayOfWeek,
+    /// 0..23
+    pub hour: u8,
+    /// 0..59
+    pub minute: u8,
+    /// 0..59
+    pub second: u8,
+}
+
+impl DateTime {
+    /// Validate the datetime, making sure it's a valid value.
+    ///
+    /// Note that the day of month is not checked currently. This means that you can set e.g. February 31th, and this function will still pass correctly.
+    pub fn validate(&self) -> Result<(), DateTimeError> {
+        if self.year > 4095 {
+            Err(DateTimeError::InvalidYear)
+        } else if self.month < 1 || self.month > 12 {
+            Err(DateTimeError::InvalidMonth)
+        } else if self.day < 1 || self.day > 31 {
+            Err(DateTimeError::InvalidDay)
+        } else if self.hour > 23 {
+            Err(DateTimeError::InvalidHour)
+        } else if self.minute > 59 {
+            Err(DateTimeError::InvalidMinute)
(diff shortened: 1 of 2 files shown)


VictorKoenders:
Tested with RTIC:

```rs
// init

let mut rtc = clocks.rtc_clock;
rtc.set_datetime(&DateTime {
    year: 0,
    month: 1,
    day: 1,
    day_of_week: DayOfWeek::Monday,
    hour: 0,
    minute: 0,
    second: 0,
})
.unwrap();
rtc.schedule_alarm(DateTimeFilter::default().second(1));
```

```rs
#[task(binds = RTC_IRQ, local = [rtc, led])]
fn real_time_clock(ctx: real_time_clock::Context) {
    ctx.local.led.toggle().unwrap();
    let rtc = ctx.local.rtc;
    rtc.clear_interrupt();
    let now = rtc.now().unwrap();

    let seconds = if now.second == 59 { 0 } else { now.second + 1 };
    rtc.schedule_alarm(DateTimeFilter::default().second(seconds));
}
```

Happily blinks away at 1hz

rp2040-hal/src/rtc.rs
```diff
@@ -1,3 +1,620 @@
...
-// TODO
+
+pub use crate::clocks::RtcClock;
+
+use crate::clocks::Clock;
+use embedded_time::fixed_point::FixedPoint;
+use rp2040_pac::{RESETS, RTC};
+
+/// Structure containing date and time information
+pub struct DateTime {
```
jonathanpallant, line 11: It might be worth using [chrono](https://docs.rs/chrono/0.4.19/chrono/) instead of re-implementing calendar functionality here. We could put it behind a feature gate to avoid the dependency if required - but ensure chrono is added with `no-default-features` to turn off the `std` feature.

rp2040-hal/src/rtc.rs
```diff
@@ -1,3 +1,620 @@
...
+            // and this file is the only file that messes with these addresses.
+            let rtc = unsafe { &*RTC::ptr() };
+
+            rtc.ctrl.modify(|_, w| w.rtc_enable().clear_bit());
+            // TODO: Add a timeout to this?
+            while rtc.ctrl.read().rtc_active().bit_is_set() {
+                core::hint::spin_loop();
+            }
+
+            let setup_0 = ((t.year as u32) << consts::RTC_SETUP_0_YEAR_LSB)
```
jonathanpallant, line 189: I think there are suitable PAC accessor methods available, so you shouldn't need the consts. See https://docs.rs/rp2040-pac/0.2.0/rp2040_pac/rtc/setup_0/index.html.

rp2040-hal/src/rtc.rs
```diff
@@ -1,3 +1,620 @@
 //! Real Time Clock (RTC)
 // See [Chapter 4 Section 8](https://datasheets.raspberrypi.org/rp2040/rp2040_datasheet.pdf) for more details
-// TODO
+
+pub use crate::clocks::RtcClock;
```
jonathanpallant, line 4: `RtcClock` in this context means the oscillator which drives the RTC IP block on the chip. This module should define its own `Rtc` object, which owns a `pac::RTC` structure - see `Spi::new()` for an example.

```
let r = hal::rtc::Rtc::new(p.RTC);
// or
let r = p.RTC.split();
```

I prefer the former.

jonathanpallant (changes requested):
I love the idea of improving our Rtc support, but this needs a little more work as noted above.

VictorKoenders:
Decided to rework this from scratch with the given feedback.

New PR here: #213

PR #208, opened 2021-11-24
Added spinlocks

VictorKoenders:
Implements spinlocks, a simple wrapper around the sio::spinlockX address.

As per documentation:

```
2.3.1.3. Hardware Spinlocks

The  SIO  provides  32  hardware  spinlocks,  which  can  be  used  to  manage  mutually-exclusive  access  to  shared  software
resources. Each spinlock is a one-bit flag, mapped to a different address (from SPINLOCK0 to SPINLOCK31). Software
interacts with each spinlock with one of the following operations:
• Read: attempt to claim the lock. Read value is nonzero if the lock was successfully claimed, or zero if the lock had
already been claimed by a previous read.
• Write (any value): release the lock. The next attempt to claim the lock will be successful.

If both cores try to claim the same lock on the same clock cycle, core 0 succeeds.

Generally  software  will  acquire  a  lock  by  repeatedly  polling  the  lock  bit  ("spinning"  on  the  lock)  until  it  is  successfully
claimed. This is inefficient if the lock is held for long periods, so generally the spinlocks should be used to protect the
short critical sections of higher-level primitives such as mutexes, semaphores and queues.

For debugging purposes, the current state of all 32 spinlocks can be observed via SPINLOCK_ST.
```

For usability I've added a `Spinlock` trait that all `SpinlockX` implement, so we can make an API like:

```rs
pub struct Mutex<T, S: Spinlock> {
    ...
}

let mutex = Mutex::<u32, SpinLock5>::new(32);
*mutex.lock() = 1234;
```

### DIFF
--- a/rp2040-hal/src/sio.rs
+++ b/rp2040-hal/src/sio.rs
@@ -150,6 +150,7 @@ macro_rules! impl_spinlock {
             /// **warning**: These spinlocks are not re-entrant, meaning that the following code will cause a deadlock:
             ///
             /// ```no_run
+            /// use rp2040_hal::sio::{Spinlock0, Spinlock};
             /// let lock_1 = Spinlock0::claim();
             /// let lock_2 = Spinlock0::claim(); // deadlock here
             /// ```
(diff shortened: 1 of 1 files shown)


9names:
Would this be easier or harder if the spinlocks were in a single array as per
https://github.com/rp-rs/rp2040-pac/pull/39

VictorKoenders:
Should be the same, the macro would have to take an index instead of a register. We can wait on that PR if you want

jonathanpallant (approved):

jonathanpallant:
Thank you for the PR! :)

PR #209, opened 2021-11-24
Draft: DMA

mgottschlag:
*Since I probably do not have much time for this during the next weeks, I'll put it here as a draft PR. I'll pick it up eventually, but if anybody needs DMA right now, feel free to test the code and/or fix the shortcomings listed below.*

# Introduction

This PR adds support for the DMA engine of the RP2040. The DMA engine has shown to be very different to other DMA engines I have previously worked with and has some interesting "features" which inevitably affect the abstraction layer:

* Individual DMA channels do not have any method to chain buffers. Instead, linked lists of buffers are implemented either by chaining multiple DMA channels together (one per buffer) or via an additional auxiliary channel. This implementation so far only supports the former, and only for double-buffered transfers - as soon as one buffer has finished, the user has to use the now-unused DMA channel to chain another buffer to the channel.
* The DMA engine only supports ring buffers when they are aligned to power-of-two addresses and sizes. The code currently does not support endless transfers to such buffers - instead, the user has to manually set up a double-buffered transfer and has to manually add buffers as described above.
* The DMA engine does not support any interrupt at the mid-point of a buffer, which makes ringbuffers rather difficult to use anyways unless the HAL chains two ringbuffers together.

As these points show, some simple DMA transfers can be implemented using one channel, whereas others require a tuple of two channels, which is reflected in the API.

Note that the code explicitly supports transfers directly between different peripherals - peripherals and buffers implement the same traits. SPI is a special case in that it always receives as much data as it sends - the DMA code therefore has a special type which implements transfers for such peripherals using twice the number of DMA channels. Other peripherals such as UART should provide a method to split the peripheral into a RX and a TX part as done by many other HALs.

# Example

See `rp2040-hal/examples/pio_dma.rs` for a complete usage example. Some relevant parts from that file:

```
    // Transfer a single message via DMA.
    let tx_buf = singleton!(: [u32; 4] = message).unwrap();
    let rx_buf = singleton!(: [u32; 4] = [0; 4]).unwrap();
    let tx_transfer = SingleBufferingConfig::new(dma.ch0, tx_buf, tx).start();
    let rx_transfer = SingleBufferingConfig::new(dma.ch1, rx, rx_buf).start();
    let (ch0, tx_buf, tx) = tx_transfer.wait();
    let (ch1, rx, rx_buf) = rx_transfer.wait();
   // Here, the channels as well as the buffers can be reused.
```

Chained transfers require a tuple of two channels, as described above:

```
let tx_buf2 = singleton!(: [u32; 4] = message).unwrap();
    let rx_buf2 = singleton!(: [u32; 4] = [0; 4]).unwrap();
    let tx_transfer = DoubleBufferingConfig::new((ch0, ch1), tx_buf, tx).start();
    let mut tx_transfer = tx_transfer.read_next(tx_buf2);
    let rx_transfer = DoubleBufferingConfig::new((dma.ch2, dma.ch3), rx, rx_buf).start();
    let mut rx_transfer = rx_transfer.write_next(rx_buf2);
    loop {
        // We simply immediately enqueue the buffers again.
        if tx_transfer.is_done() {
            let (tx_buf, next_tx_transfer) = tx_transfer.wait();
            tx_transfer = next_tx_transfer.read_next(tx_buf);
        }
        if rx_transfer.is_done() {
            let (rx_buf, next_rx_transfer) = rx_transfer.wait();
            // Do something with rx_buf here!
            rx_transfer = next_rx_transfer.write_next(rx_buf);
        }
    }
```

# Memory Safety

The code uses the `embedded-dma` traits and assumes that it gets `static` references to memory. Implementations of `Drop` should make sure that a buffer is not used anymore once the corresponding DMA object is destructed (see below, this is currently missing). The API is unsafe in the presence of leaks where `drop()` is not called, e.g., `mem::forget()`.

# Known Defects

- The code still contains many notes I wrote down while designing the API, those can just be thrown out.
- The DMA types are all still missing implementations of `Drop`, which is absolutely essential for memory safety.
- Other types of transfers (ring buffers?) are not implemented yet.
- Currently, only PIO and SPI are supported. Some other peripherals require refactoring to be usable with DMA - for example, the UART code should be refactored to provide a method to split the UART into a RX and a TX object to allow for parallel operations on both.

### DIFF
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -17,7 +17,8 @@ embedded-time = "0.12.0"
 embedded-dma = "0.1.2"
 itertools = { version = "0.10.1", default-features = false }
 nb = "1.0"
-rp2040-pac = "0.1.5"
+#rp2040-pac = "0.1.5"
+rp2040-pac = { path = "../../rp2040-pac" }
 paste = "1.0"
 pio = { git = "https://github.com/rp-rs/pio-rs.git", branch = "main" }
 pio-proc = { git = "https://github.com/rp-rs/pio-rs.git", branch = "main" }
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -14,6 +14,7 @@ cortex-m = "0.7.2"
 embedded-hal = { version = "0.2.5", features = ["unproven"] }
 eh1_0_alpha = { version = "=1.0.0-alpha.5", package="embedded-hal", optional=true }
 embedded-time = "0.12.0"
+embedded-dma = "0.1.2"
 itertools = { version = "0.10.1", default-features = false }
 nb = "1.0"
 rp2040-pac = "0.1.5"
--- a/rp2040-hal/examples/pio_dma.rs
+++ b/rp2040-hal/examples/pio_dma.rs
@@ -1,0 +1,146 @@
+//! This example shows how to read from and write to PIO using DMA.
+//!
+//! If a LED is connected to that pin, like on a Pico board, it will continously output "HELLO
+//! WORLD" in morse code. The example also tries to read the data back. If reading the data fails,
+//! the message will only be shown once, and then the LED remains dark.
+#![no_std]
+#![no_main]
+
+use cortex_m::singleton;
+use cortex_m_rt::entry;
+use hal::dma::{DMAExt, TransferConfig};
+use hal::gpio::{FunctionPio0, Pin};
+use hal::pac;
+use hal::pio::PIOExt;
+use hal::sio::Sio;
+use panic_halt as _;
+use rp2040_hal as hal;
+
+#[link_section = ".boot2"]
+#[used]
+pub static BOOT2: [u8; 256] = rp2040_boot2::BOOT_LOADER;
+
+#[entry]
+fn main() -> ! {
+    let mut pac = pac::Peripherals::take().unwrap();
+
+    let sio = Sio::new(pac.SIO);
+    let pins = hal::gpio::Pins::new(
+        pac.IO_BANK0,
+        pac.PADS_BANK0,
+        sio.gpio_bank0,
+        &mut pac.RESETS,
+    );
+
+    // configure LED pin for Pio0.
+    let _led: Pin<_, FunctionPio0> = pins.gpio25.into_mode();
+    // PIN id for use inside of PIO
+    let led_pin_id = 25;
+
+    // HELLO WORLD in morse code:
+    // .... . .-.. .-.. --- / .-- --- .-. .-.. -..
+    let message = [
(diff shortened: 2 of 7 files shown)


mgottschlag:
As the problem just came up in the Matrix room:

- [ ] The documentation should describe how to implement DMA transfers of variable length, i.e., it should describe how to implement `StaticReadBuffer` so that the type returns only the part of an array that should be transferred.

korken89:
Hi, I'm thinking on continuing this.
Anything against? :)

mgottschlag:
Yes, please!

ogoffart:
Is anyone still working on this?
I have rebased this branch and tried it out in https://github.com/slint-ui/rp-hal/tree/dma

I don't know if this is the place for feedback, but i'm going to do it anyway :-)

I'm trying to use the DMA to flush a line of pixels to a screen connected to the Raspberry pi pico with the SPI. 
My attempt at using the branch is there: https://github.com/slint-ui/slint/pull/1344

Here is what i find difficult.

I have two buffer of pixels (which is a `&'static mut [Rgb565]` (rg565 is basically a u16)) one buffer is being rendered, and the idea is that the other buffer can be processed and send to the SPI.
So I want to pass the buffer to the DMA then get it back when needed. But it needs a ReadBuffer, and `&'static mut` is only a WriteBuffer. If i coerce it to a `&'static` by removing the mut, then i can't get the mut back when i'm finish to prepare the next frame.
also i only need to send part of this buffer. So that's why my patch has a `PartialReadBuffer` that wrap a `&'static mut` and a Range, but that's some use of unsafe code that i'd like to avoid.

The other problem I had is that the SPI is used by several part of the code. The touch screen and the display both need access to the SPI.  I'm using the shared-bus crate to share with both the touchscreen and the display, but then i still can't get the SPI handle to give it to the DMA when needed.  So i've had to use unsafe to steal the SPI.
(Even if it was not the case, the display driver also had a hold of it and of the CS pin, so that's another problem, but that's because of the display driver)
(Another complication is that the touch device work with  much lower frequency that the display, so more unsafe code is needed to change the frequency)

Something i found complicated was that after the DMA, the touch driver would not work anymore. After hours of debugging, i found out it was because the read FIFO was not cleared, so i cleared it like so:
```
while !to.read().is_err() {}
```
but  that's not really intuitive. I wonder if the wait() call should not clear the fifo or something.

Altaflux:
I have been using this PR successfully for DMA with PIO and works great!
One caveat is that DMA thru PIO supports data-sources of not only u32, but u16 and i think u8 too.
How it is currently implemented I can only use it with u32 arrays although it would be efficient to use smaller sized data types.

The code already supports configuring DMA to switch between the different data-types, the only problem appears to be that it is currently hardcoded to only accept u32 type sources.

9names:
Closed via #457

PR #217, opened 2021-12-01
BSP for the Adafruit KB2040

DrewTChrist:
I was able to run the kb2040_rainbow example in the pull request, but I haven't tried much more with it yet. I linked the pinout below, so please let me know if you spot any mistakes. I used the naming convention that the other Adafruit BSPs use because it appears the pro micro BSP has a different way of naming the pins.

I also added the itsy_bitsy_rp2040 BSP description to the readme.

[KB2040](https://www.adafruit.com/product/5302)
[KB2040 pinout](https://cdn-blog.adafruit.com/uploads/2021/11/pinlabels.png)


boards/kb2040/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+#[cfg(feature = "rt")]
+extern crate cortex_m_rt;
+#[cfg(feature = "rt")]
+pub use cortex_m_rt::entry;
+
+pub use hal::pac;
+
+hal::bsp_pins!(
+    Gpio0 { name: d0 },
+    Gpio1 { name: d1 },
```
9names, line 13: Shouldn't these be Tx/Rx to match the UART default function on these pins?

DrewTChrist: Oh yes you're right. I will fix that when I get home today. For future reference, how does one figure out the default function for each pin? The silkscreen says Tx/Rx which I overlooked. I took a peak at section 1.4 of the datasheet, but I don't see anything about default pin functions.

9names: I meant default to mean the recommended use as per the board markings.
I usually pick 0 and 1 for UART TX/RX because they're easy to locate on a pico.

To work out the pin default, look at the reset value of funcsel and the "function select" page of the datasheet.
I believe their default function is to not have a special function.

9names (approved):
CI failures are unrelated to this PR.
LGTM

PR #223, opened 2021-12-03
Add safety docs from atsamd-rs to GPIO traits to make CI happy

9names:
Rust 1.57 clippy triggers warnings that were not present on 1.56, breaking CI
```
warning: docs for unsafe trait missing `# Safety` section
```

Since we borrowed these traits from atsamd-rs's pin abstraction, I've copied the safety docs from there.
https://github.com/atsamd-rs/atsamd/blob/16ff6e0afd3e5e01084149ad033ac04f5eb9fd82/hal/src/gpio/v2/reg.rs#L213

PR #224, opened 2021-12-03
Generate sev in sio.fifo.write()

### LINKED ISSUE #222: Multicore: sio.fifo.write() doesn't wake other core
sio.fifo.read_blocking() puts the current core to sleep.
sio.fifo.write_blocking() pushes a character, then sends a signal to wake the other core.
sio.fifo.write() does not send a wake signal, so the reading core will stay asleep forever.

Do we want to keep this functionality?
It's definitely a foot gun, the first example I wrote was broken by the assumption that write would wake the other core if it was blocking.

9names:
Generate a send event signal via `sev` in sio.fifo.write() so that sio.fifo.blocking_read() will wake up.
Closes #222


jannic:
LGTM. Clippy errors are the ones fixed by https://github.com/rp-rs/rp-hal/pull/223 and not related to this PR.

PR #226, opened 2021-12-03
Add basic multicore FIFO example

### LINKED ISSUE #219: Add docs and an example for multicore
Current impl does not have an example, needs a file-level comment

_Originally posted by @9names in https://github.com/rp-rs/rp-hal/pull/89#discussion_r760060812_

9names:
Resolves #219

### DIFF
--- a/rp2040-hal/src/multicore.rs
+++ b/rp2040-hal/src/multicore.rs
@@ -11,7 +11,7 @@
 //! # Usage
 //!
 //! ```no_run
-//! static mut CORE1_STACK: Stack<4096> = Stack { mem: [0; 4096] };
+//! static mut CORE1_STACK: Stack<4096> = Stack::new();
 //! fn core1_task() -> ! {
 //!     loop{}
 //! }
@@ -98,6 +98,13 @@ pub struct Multicore<'p> {
 pub struct Stack<const SIZE: usize> {
     /// Memory to be used for the stack
     pub mem: [usize; SIZE],
+}
+
+impl<const SIZE: usize> Stack<SIZE> {
+    /// Construct a stack of length SIZE, initialized to 0
+    pub const fn new() -> Stack<SIZE> {
+        Stack { mem: [0; SIZE] }
+    }
 }
 
 impl<'p> Multicore<'p> {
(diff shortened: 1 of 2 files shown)


9names:
One thing I discovered while writing this: we probably need to Sio::split, since trying to use pins while also using `Multicore` is painful due to both taking a `&mut Sio`

9names:
Now has some docs for the multicore module as well as the example.

9names:
As noticed by wct on matrix chat:
"160 mhz is not the correct system bus frequency"
I've updated the example to send that value over the FIFO instead.

rp2040-hal/src/multicore.rs
```diff
@@ -1,5 +1,41 @@
...
+//! The options for an entrypoint for core1 are
+//! - a function that never returns - eg
+//! `fn core1_task() -> ! { loop{} }; `
+//! - a lambda (note: This requires a global allocator which requires a nightly compiler. Not recommended for beginners)
+//!
+//! # Usage
+//!
+//! ```no_run
+//! #[repr(C, align(32))]
+//! pub struct Stack<const SIZE: usize> {
```
thejpster, line 15: Maybe `Stack` should be in `multicore` - seeing as pretty much everyone is going to want it.

9names: I'll bring it into multicore.rs

Also still worth considering whether we switch multicore from taking a **&mut slice** to taking a **struct Stack** if we expect users to always use the type to avoid pulling out the inner slice, as per @jannic's suggestion:

https://github.com/rp-rs/rp-hal/compare/main...jannic:user-stack-aligned

thejpster: Excellent. Maybe a `::new()` as well? ;)

9names: Okay, you twisted my arm :D

jannic: Just curious: Do you happen to know if the rust compiler reliably inlines `new` and avoids copying the created object?
(I guess in  this case it wouldn't matter much as long as you don't try to give half of your own stack to the other core.)

9names: In this case it's a const fn so it's always going to be.
It has to be if it's going to initialise a static object in Rust, since there's no life before main.
In the general case I think it depends on what you're doing in new().

thejpster (approved):

PR #235, opened 2021-12-08
Update Pico RTIC example to use timer alarm HAL

camrbuss:
Example now uses pico bsp, RTIC multilock, and rp-hal Timer Alarm HAL.

### DIFF
--- a/boards/pico/examples/pico_rtic.rs
+++ b/boards/pico/examples/pico_rtic.rs
@@ -8,8 +8,9 @@ use rp2040_hal as hal;
 mod app {
 
     use embedded_hal::digital::v2::OutputPin;
+    use embedded_time::duration::Extensions;
     use pico::{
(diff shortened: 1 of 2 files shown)


boards/pico/examples/pico_rtic.rs
```diff
@@ -8,16 +8,18 @@ use rp2040_hal as hal;
 mod app {
 
     use embedded_hal::digital::v2::OutputPin;
+    use embedded_time::duration::Extensions;
     use pico::{
-        hal::{self, clocks::init_clocks_and_plls, pac, watchdog::Watchdog, Sio},
+        hal::{self, clocks::init_clocks_and_plls, watchdog::Watchdog, Sio},
```
9names, line 13: Since you're getting hal through the BSP now (which is nice), we should remove `use rp-hal as hal` from line 5 and switch the rtic::app line to use `pico::hal::pac` instead.

9names (approved):
LGTM

jonathanpallant (approved):

PR #240, opened 2021-12-12
WIP: alternative to passing PeripheralClock freq to I2C/SPI as generic Into<Hertz<u32>>

9names:
After spending time troubleshooting why my I2C peripheral wasn't working (and determining that I fed the wrong clock frequency into the constructor for I2C), I thought it would be worth floating the idea of stripping out the generic constructor with one that takes the Clocks::PeripheralClock struct directly.

Pros:
- Can't feed in the wrong number
- You were probably going to give it the data from this struct anyway, so the conversion goes into the peripheral driver

Cons:
- Any changes to the Clock structures will require changes to the SPI and I2C peripherals drivers
- How to resolve lifetimes?

### DIFF
--- a/rp2040-hal/src/i2c.rs
+++ b/rp2040-hal/src/i2c.rs
@@ -9,6 +9,19 @@
 //! let mut peripherals = pac::Peripherals::take().unwrap();
 //! let sio = Sio::new(peripherals.SIO);
 //! let pins = Pins::new(peripherals.IO_BANK0, peripherals.PADS_BANK0, sio.gpio_bank0, &mut peripherals.RESETS);
+//! let mut watchdog = hal::Watchdog::new(pac.WATCHDOG);
+//! const XTAL_FREQ_HZ:u32 = 12_000_000;
+//! let clocks = hal::clocks::init_clocks_and_plls(
+//!         XTAL_FREQ_HZ,
+//!         pac.XOSC,
+//!         pac.CLOCKS,
+//!         pac.PLL_SYS,
+//!         pac.PLL_USB,
+//!         &mut pac.RESETS,
+//!         &mut watchdog,
+//!     )
+//!     .ok()
+//!     .unwrap();
 //!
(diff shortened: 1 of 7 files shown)


rp2040-hal/src/i2c.rs
```diff
@@ -241,18 +241,17 @@ macro_rules! hal {
...
                 /// Configures the I2C peripheral to work in master mode
-                pub fn $i2cX<F, SystemF>(
+                pub fn $i2cX<F>(
                     i2c: $I2CX,
                     sda_pin: Pin<Sda, FunctionI2C>,
                     scl_pin: Pin<Scl, FunctionI2C>,
                     freq: F,
                     resets: &mut RESETS,
-                    system_clock: SystemF) -> Self
+                    system_clock: super::clocks::PeripheralClock) -> Self
```
jannic, line 250: I don't think ownership of `PeripheralClock` can be passed to the I2C controller: That would fail as soon as you try to initialize a second peripheral which also needs to now the current clock frequency.

Perhaps pass a reference, and to avoid lifetime issues, don't store the reference in the struct, but extract the actual frequency in the constructor, and only store that value?
(BTW: The parameter probably shouldn't be called `system_clock`. I guess that's the reason why you made the mistake of passing the system clock frequency.)

9names: > extract the actual frequency in the constructor, and only store that value?

We already don't ever use it after the constructor. If we did need it again, we'd need to look it up.
Maybe that's a better answer, but would need the infra to do that lookup - I'm keen to do it but wanted to get other parts done first.
But for now, taking a ref should be good, i'll do that.

> The parameter probably shouldn't be called system_clock

I'll need to check if it's supposed to take the system clock or peripheral clock, though I suspect you are correct.

9names: Going to move this back to draft until I can work out lifetimes :(

PR #242, opened 2021-12-18
Comment out embassy traits.

thejpster:
The git dep prevents us from publishing, rp2040-hal, and embassy
aren't doing a release any time soon.


9names (approved):
Lgtm

PR #243, opened 2021-12-19
Set up for 0.3.0 release.

thejpster:
Somehow we already had a bunch of commits talking about 0.3.0, but it
was never released. Hence why this maybe doesn't change as many version
numbers as you might expect.

### DIFF
--- a/rp2040-hal/CHANGELOG.md
+++ b/rp2040-hal/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - SPI example
 - Watchdog example
 - ADC documentation
+- Lots of bug fixes :)
 
 ### Changed
 
--- a/rp2040-hal/CHANGELOG.md
+++ b/rp2040-hal/CHANGELOG.md
@@ -18,12 +18,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.3.0] - 2021-12-19
 
 ### Added
+
 - A README!
 - Implementation of the `critical-section` API
-- DMA support
-- Embedded HAL 1.0 Alpha support
+- Embedded HAL 1.0-alpha6 support
 - I²C Controller and Peripheral support
-- Multi-core support
+- Multi-core support, including spin-locks and SIO FIFO
+- RTC support
+- USB Device support
+- Timer support
 - PIO support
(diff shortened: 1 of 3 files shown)


rp2040-hal/CHANGELOG.md
```diff
@@ -8,24 +8,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ### Added
-- peripheral drivers: Timer(counter) with EH traits, USB, PIO
-- examples: Watchdog, GPIO, ADC, i2c, rtic blinky, usb (serial echo, mouse)
-- docs: Watchdog
-- bsps: adafruit feather
-- bsp examples: sparkfun pro micro PIO RGB rainbow LED
+- A README!
+- Implementation of the `critical-section` API
+- DMA support
```
9names, line 23: We don't have DMA support yet, probably best to leave that for another release

rp2040-hal/CHANGELOG.md
```diff
@@ -8,24 +8,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Embedded HAL 1.0 Alpha support
```
9names, line 24: Each alpha is a breaking release, should specify alpha.6 here so users know which version to expect

rp2040-hal/CHANGELOG.md
```diff
@@ -8,24 +8,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- PIO support
+- Implementation of `rng_core::RngCore` for `RingOscillator`
+- ADC example
+- GPIO Interrupt support
+- Multi-core FIFO example
+- PIO LED Blinky example
+- ROM Functions example
+- SPI example
+- Watchdog example
+- ADC documentation
```
9names, line 37: Missing:
- RTC
- Timers (unless you group that under RTC
- Spinlocks (unless you group that under multicore)
- USB

And "many bug fixes"!

9names (changes requested):

rp2040-hal/CHANGELOG.md
9names: - hardware divide/modulo

thejpster:
I think that's all the extra items ticked off the list.

9names (approved):

Issue #244, opened 2021-12-20
Move the BSPs to their own repos


thejpster:
Now that we have a bunch of BSPs and they are relatively stable, I'd like to propose kicking them out of the rp-hal repository and giving each of them their own repository.

Advantages:

* It's much easier to tag and manage BSP releases, because the entire repo gets published as one package on crates.io
* Examples can have nicer names because they won't clash with each other when you build a workspace

Disadvantages:

* A change to every BSP now requires 9 pull-requests instead of one.

thejpster:
Most of the BSP crates are publishable, except for `pico` which is using the unpublished embassy-traits package. We'll need to remove that.

```console
➜  boards git:(main) for i in *; do pushd $i; cargo publish --dry-run; popd; done 

~/Documents/rp-rs/rp-hal/boards/adafruit_macropad
    Updating crates.io index
   Packaging adafruit_macropad v0.1.0 (/home/jonathan/Documents/rp-rs/rp-hal/boards/adafruit_macropad)
   Verifying adafruit_macropad v0.1.0 (/home/jonathan/Documents/rp-rs/rp-hal/boards/adafruit_macropad)
  Downloaded rp2040-hal v0.3.0
  Downloaded 1 crate (109.8 KB) in 1.25s
    Finished dev [unoptimized + debuginfo] target(s) in 16.10s
   Uploading adafruit_macropad v0.1.0 (/home/jonathan/Documents/rp-rs/rp-hal/boards/adafruit_macropad)
warning: aborting upload due to dry run
~/Documents/rp-rs/rp-hal/boards

~/Documents/rp-rs/rp-hal/boards/feather_rp2040
    Updating crates.io index
   Packaging feather_rp2040 v0.1.0 (/home/jonathan/Documents/rp-rs/rp-hal/boards/feather_rp2040)
   Verifying feather_rp2040 v0.1.0 (/home/jonathan/Documents/rp-rs/rp-hal/boards/feather_rp2040)
    Finished dev [unoptimized + debuginfo] target(s) in 14.70s
   Uploading feather_rp2040 v0.1.0 (/home/jonathan/Documents/rp-rs/rp-hal/boards/feather_rp2040)
warning: aborting upload due to dry run
~/Documents/rp-rs/rp-hal/boards

~/Documents/rp-rs/rp-hal/boards/itsy_bitsy_rp2040
    Updating crates.io index
[...]

PR #245, opened 2021-12-20
Clean up BSPs

thejpster:
Gets them all to a 0.1.0 release.

You might want to do https://github.com/rp-rs/rp-hal/issues/244 instead (but at least they're in a state where they can be separated out fairly cleanly).


9names (approved):
Looks good, happy to merge this even if we decided to go through with the #244 split.

If we're prefixing all boards with a vendor name should we also rename pico to raspberry-pi-pico to be consistent?
Or is it obvious enough that Pico is from Raspberry Pi and the prefix adds no value?

anall:
If we're publishing it, `pico` seems like something we should avoid claiming? -- too generic of a name.

9names:
I hadn't considered that, but also since it's generic and short it is also already taken.
https://crates.io/crates/pico

thejpster:
I did wonder if I should rename pico. We probably should. `raspberrypi-pico` or `raspberry-pi-pico`?

thejpster:
Oh one more thing I thought about but haven't yet done - should the crate owners all be reset to "The rp-rs Developers"?

9names:
I like `raspberrypi-pico` more, don't really need another dash to disambiguate.

> should the crate owners all be reset to "The rp-rs Developers"

Owners in which respect? If you mean the authors field in Cargo.toml that's fine by me, would be good to get quorum before doing that though.

jannic:
What about `rp-pico`?

thejpster:
> What about `rp-pico`?

OK, that does have a certain consistency to it.

thejpster:
> Oh one more thing I thought about but haven't yet done - should the crate owners all be reset to "The rp-rs Developers"?

@ithinuel
@hmvp 
@eolder
@onnenon 
@anall 

Are you happy I reset your BSP crates to have an author of "The rp-rs Developers" in the relevant `Cargo.toml` files?

boards/rp-pico/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "pico"
+name = "rp-pico"
 version = "0.1.0"
```
jannic: Version needs to be higher for `rp-pico`, as there is already an older BSP with version `0.1.3` with that name.
So bump it to `0.2.0`, I guess?

thejpster: https://github.com/rp-rs/rp-hal/pull/245/commits/6271e8670fe61cafb48f636ceaf5e001949a1442

dylanede:
I noticed that a few of the BSPs have dependencies they don't use. For example adafruit-macropad depends on `cortex-m` even though it does not import anything from it. Is that intentional? If not maybe cleaning that up should be included in this PR.

thejpster:
This PR has been merged but I'd be happy to see one that trims dependencies down the minimum set required.

PR #210, opened 2021-11-25
UART owned pins + reader/writer split

VictorKoenders:
This PR makes the UartPeripheral take ownership of the pins that it needs.

```rust
const XOSC_CRYSTAL_FREQ: u32 = 12_000_000; // Typically found in BSP crates

let mut peripherals = pac::Peripherals::take().unwrap();
let sio = Sio::new(peripherals.SIO);
let pins = Pins::new(peripherals.IO_BANK0, peripherals.PADS_BANK0, sio.gpio_bank0, &mut peripherals.RESETS);
let mut watchdog = Watchdog::new(peripherals.WATCHDOG);
let mut clocks = init_clocks_and_plls(XOSC_CRYSTAL_FREQ, peripherals.XOSC, peripherals.CLOCKS, peripherals.PLL_SYS, peripherals.PLL_USB, &mut peripherals.RESETS, &mut watchdog).ok().unwrap();

// This is new:
// Set up UART on GP0 and GP1 (Pico pins 1 and 2)
let pins = (
    pins.gpio0.into_mode::<FunctionUart>(),
    pins.gpio1.into_mode::<FunctionUart>(),
);

let uart = UartPeripheral::new(peripherals.UART0, pins, &mut peripherals.RESETS)
    .enable(
        uart::common_configs::_9600_8_N_1,
        clocks.peripheral_clock.into(),
    )
    .unwrap();

uart.write_full_blocking(b"Hello World!\r\n");

// this is also new
let (reader, writer) = uart.split();

writer.write_full_blocking(b"Hello World!\r\n");
```

`UartPeripheral` takes any type that implements `trait ValidUartPinout<UART: UartDevice>`. This is implemented for:
- tuple `(TX, RX)`
- tuple `(TX, RX, CTS, RTS)`
- the new `Pins` struct, in which developers can hand-pick the pins they want to use.

There are 4 traits that constrain the pins to the 4 functions they can have: `Tx<UART>`, `Rx<UART>`, `Cts<UART>` and `Rts<UART>`.

These traits are also implemented for `()`, so each one of the fields is optional. This is desired because some boards/devices have a single-wire UART connection that can only transfer or receive, so developers can use e.g. `(Gpio1, ())`

PR #247, opened 2021-12-23
UART IRQ examples

thejpster:
Demonstrates some methods for getting your HAL peripherals somewhere an interrupt routine can use them.

### DIFF
--- a/boards/adafruit-feather-rp2040/src/lib.rs
+++ b/boards/adafruit-feather-rp2040/src/lib.rs
@@ -7,8 +7,8 @@ extern crate cortex_m_rt;
 #[cfg(feature = "rt")]
 pub use cortex_m_rt::entry;
 
-//// The linker will place this boot block at the start of our program image. We
-//// need this to help the ROM bootloader get our code up and running.
+/// The linker will place this boot block at the start of our program image. We
+/// need this to help the ROM bootloader get our code up and running.
 #[cfg(feature = "boot2")]
 #[link_section = ".boot2"]
 #[no_mangle]
--- a/boards/adafruit-itsy-bitsy-rp2040/src/lib.rs
+++ b/boards/adafruit-itsy-bitsy-rp2040/src/lib.rs
@@ -7,8 +7,8 @@ extern crate cortex_m_rt;
 #[cfg(feature = "rt")]
 pub use cortex_m_rt::entry;
 
-//// The linker will place this boot block at the start of our program image. We
-//// need this to help the ROM bootloader get our code up and running.
+/// The linker will place this boot block at the start of our program image. We
+/// need this to help the ROM bootloader get our code up and running.
 #[cfg(feature = "boot2")]
 #[link_section = ".boot2"]
 #[no_mangle]
--- a/boards/adafruit-kb2040/examples/adafruit_kb2040_rainbow.rs
+++ b/boards/adafruit-kb2040/examples/adafruit_kb2040_rainbow.rs
@@ -28,7 +28,7 @@ use adafruit_kb2040::{
 use smart_leds::{brightness, SmartLedsWrite, RGB8};
 use ws2812_pio::Ws2812;
 
-//// The linker will place this boot block at the start of our program image.
+/// The linker will place this boot block at the start of our program image.
 /// We need this to help the ROM bootloader get our code up and running.
 #[link_section = ".boot2"]
 #[used]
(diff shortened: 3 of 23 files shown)


jannic:
Very nice examples. They show the complexities involved in setting up an interrupt and passing data to and from interrupt context, without hiding it behind some opaque abstraction. At the same time, they are easy enough to just read the code and understand what's going on. I like it.
(But note that I don't have enough experience with interrupt code to judge if it's a good approach or if the code could be even simpler.)

rp2040-hal/examples/uart_irq_echo.rs
```diff
@@ -0,0 +1,193 @@
...
+/// This how we transfer the UART into the Interrupt Handler
+static GLOBAL_UART: Mutex<RefCell<Option<Uart>>> = Mutex::new(RefCell::new(None));
+
+/// Entry point to our bare-metal application.
+///
+/// The `#[entry]` macro ensures the Cortex-M start-up code calls this function
+/// as soon as all global variables are initialised.
+///
+/// The function configures the RP2040 peripherals, then writes to the UART in
+/// an inifinite loop.
```
9names, line 75:
```suggestion
/// an infinite loop.
```

rp2040-hal/examples/uart_irq_tx_buffer.rs
```diff
@@ -0,0 +1,285 @@
+//! # UART IRQ TX BUffer Example
```
9names, line 1:
```suggestion
//! # UART IRQ TX Buffer Example
```

rp2040-hal/examples/uart_irq_tx_buffer.rs
```diff
@@ -0,0 +1,285 @@
...
+type Uart = hal::uart::UartPeripheral<hal::uart::Enabled, pac::UART0>;
+
+/// This describes the queue we use for outbound UART data
+struct UartQueue {
+    mutex_cell_queue: Mutex<RefCell<Queue<u8, 64>>>,
+    interrupt: pac::Interrupt,
+}
+
+/// The linker will place this boot block at the start of our program image. We
+// need this to help the ROM bootloader get our code up and running.
```
9names, line 66:
```suggestion
/// The linker will place this boot block at the start of our program image. We
/// need this to help the ROM bootloader get our code up and running.
```

rp2040-hal/examples/uart_irq_tx_buffer.rs
```diff
@@ -0,0 +1,285 @@
...
+    interrupt: hal::pac::Interrupt::UART0_IRQ,
+};
+
+/// Entry point to our bare-metal application.
+///
+/// The `#[entry]` macro ensures the Cortex-M start-up code calls this function
+/// as soon as all global variables are initialised.
+///
+/// The function configures the RP2040 peripherals, then writes to the UART in
+/// an inifinite loop.
```
9names, line 91:
```suggestion
/// This function configures the RP2040 peripherals, then writes to the UART in
/// an infinite loop.
```

rp2040-hal/examples/uart_irq_tx_buffer.rs
```diff
@@ -0,0 +1,285 @@
...
+    let pins = hal::gpio::Pins::new(
+        pac.IO_BANK0,
+        pac.PADS_BANK0,
+        sio.gpio_bank0,
+        &mut pac.RESETS,
+    );
+
+    // UART TX (characters sent from RP2040) on pin 1 (GPIO0)
+    let _tx_pin = pins.gpio0.into_mode::<hal::gpio::FunctionUart>();
+    // UART RX (characters reveived by RP2040) on pin 2 (GPIO1)
```
9names, line 130:
```suggestion
    // UART RX (characters received by RP2040) on pin 2 (GPIO1)
```

rp2040-hal/examples/uart_irq_tx_buffer.rs
```diff
@@ -0,0 +1,285 @@
...
+    cortex_m::interrupt::free(|cs| {
+        GLOBAL_UART.borrow(cs).replace(Some(uart));
+    });
+
+    // But we can blink an LED.
+    let mut led_pin = pins.gpio25.into_push_pull_output();
+
+    loop {
+        // Light the LED whilst the main thread is in the transmit routine. It
+        // shouldn't be on very long, but it will be on whilst we get enough
```
9names, line 155:
```suggestion
        // shouldn't be on very long, but it will be on until we get enough
```

9names (commented):
Code + documentation was good, left a few suggestions for fixing typos I found while reading through.
I assume we're going to wait until #210 lands before merging this?

thejpster:
Makes sense

9names (approved):

PR #248, opened 2021-12-26
Bump versions of HAL and BSPs

jannic:
The change to UartPeripheral in #210 was a breaking change:
Bump the version of the HAL and all dependent BSP crates.


thejpster:
Hmm, these kinds of change are going to be quite a bit more work if we do #244.

thejpster (approved):

thejpster:
Er, GHA appears to have hung. Or the cargo check is taking over an hour to complete...

thejpster:
Builds fine on my machine. I guess we'll have to let it time out and then try again.

PR #252, opened 2022-01-01
boards/pico: Remove mention of boot select button

david-sawatzke:
The boot selector pin is attached to SPI_SS and not easily read (an
example
https://github.com/raspberrypi/pico-examples/blob/master/picoboard/button/button.c),
since the code reading needs to reside in SRAM.

Simply fix the documentation by only mentioning the pin


9names:
Good change, thanks for the PR!

Issue #250, opened 2021-12-30
Example pico_uart_irq_echo doesn't build


kolbma:
```
$ cargo build --release --example pico_uart_irq_echo
   Compiling rp-pico v0.3.0 ($HOME/temp/rp-hal/boards/rp-pico)
error: linking with `rust-lld` failed: exit status: 1
  |
[...]

PR #254, opened 2022-01-03
Fix ci masking link errors

9names:
Fix CI BSP examples not generating failure on link errors

#250 exists because CI was not generating errors on failure to link.
This was due to [RUSTFLAGS: "-D warnings" ](https://github.com/9names/rp-hal/blob/c02984b422d3cc12e14a3e17373f7bb3b57e5e58/.github/workflows/build_and_test.yml#L8) suppressing the link errors.

In this PR:
- I remove "-D warnings" from RUSTFLAGS, to ensure we don't mask our warnings like linker error
- switch from `cargo check` to `cargo build` so we test linking, and also add release builds to the workflow in case they get build errors.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -26,6 +26,14 @@ jobs:
           args: --workspace
       - uses: actions-rs/cargo@v1
         with:
+          command: build
+          args: --release --workspace --examples
+      - uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --release --workspace --examples --features eh1_0_alpha
+      - uses: actions-rs/cargo@v1
+        with:
           command: test
           args: --doc --target x86_64-unknown-linux-gnu
       - uses: actions-rs/cargo@v1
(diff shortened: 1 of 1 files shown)


9names:
Note this is expected to have a build error - the examples need to be fixed!

thejpster (approved):

thejpster:
Hmm. I wonder if this builds the source branch contents, or what the destination branch will look like post merge. They might be different if the source branch is not up to date.

We should build both really, to be sure.

anall:
> or what the destination branch will look like post merge.

I believe that's the idea behind Bors (https://bors.tech/) and Homu (https://github.com/barosl/homu)

thejpster:
Ah. Well then can someone update the PR now the fix has gone into main?

thejpster:
Oh, it's in @9names's fork, so I don't think I can do it. At least not on my phone.

9names:
> Oh, it's in @9names's fork, so I don't think I can do it. At least not on my phone.

You should be allowed - when you make a PR through github you usually (always?) give write permission on your branch for anyone with write permissions on the target repo.
I'm happy to rebase it though.

thejpster:
Yeah but there's no button in the mobile UI to update the branch.

I can see the CI is passing, but I can't see the examples in the logs. Can anyone verify that these commands do actually build the examples?

thejpster (approved):

thejpster:
Sweet!

PR #258, opened 2022-01-08
Add an example for accessing an SD/MMC card via SPI

WeirdConstructor:
Hi!

I've connected an SD card to my Raspberry Pi Pico now and successfully used the `embedded_sdmmc` crate with it.

I cleaned up my example code and added diagnostics LED signals to get feedback even without a debugging probe.

I hope it's okay to add defmt to the dev-dependencies, even though the examples use the UF2 upload of the Pico.

### DIFF
--- a/boards/rp-pico/Cargo.toml
+++ b/boards/rp-pico/Cargo.toml
@@ -28,6 +28,10 @@ cortex-m-rtic = "0.6.0-rc.4"
 nb = "1.0"
 i2c-pio = { git = "https://github.com/ithinuel/i2c-pio-rs", rev = "df06e4ac94a5b2c985d6a9426dc4cc9be0d535c0" }
 heapless = "0.7.9"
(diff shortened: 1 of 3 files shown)


WeirdConstructor:
If this is merged, I can also make a PR for the ws2812 example I got in another branch (which is based on this branch currently unfortunately).

General question: Would it make sense to port these examples to the more generic rp-hal crate too? Or is that out of scope? I just want to make starting out for others after me a bit easier (these examples also serve as reference for myself).

9names (approved):
Looks good to me.
Love the ASCII art diagram!

9names:
> Would it make sense to port these examples to the more generic rp-hal crate too?

My preference is to keep the HAL examples simple, preferably involving as few moving parts as possible while showing off the different modes for the peripherals.
For the moment the more fully-worked examples are going into BSPs.
I think at some stage we might want to make a new section in the workspace that has projects rather than examples.
For example: showing how to use all the various Pico expansion boards would be good to have, but maybe we don't want to always keep them up to date with the BSP(s).

PR #259, opened 2022-01-09
Added pin function documentation from the datasheet to rp-pico

WeirdConstructor:
I don't know if this is useful or if it is out of scope of the board specific documentation, but I quickly (half automatically via script) converted the rp2040 datasheet section 1.4.3 to board specific pin documentation.

Like I already wrote on Matrix, I would love to do this for `rp2040_hal::gpio::pin::bank0` too, but the macro does not support documentation meta information and I don't know how to add it.

### DIFF
--- a/boards/rp-pico/src/lib.rs
+++ b/boards/rp-pico/src/lib.rs
@@ -19,23 +19,23 @@ pub use hal::pac;
 
 hal::bsp_pins!(
     /// GPIO 0 supports following functions:
-    /// - `SPI0 RX  `
-    /// - `UART0 TX `
+    /// - `SPI0 RX`
+    /// - `UART0 TX`
     /// - `I2C0 SDA`
     /// - `PWM0 A`
     /// - `SIO`, `PIO0`, `PIO1`
     Gpio0 { name: gpio0 },
 
     /// GPIO 1 supports following functions:
-    /// - `SPI0 CSn `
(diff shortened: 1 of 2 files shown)


thejpster:
Huh! I did not know the `bsp_pins` macro supported aliases!

I like this idea in principle (subject to the Markdown tables being formatted correctly for the benefit of those reading the source code directly). Do we have any examples of using the aliases, or would be simplify any of the existing examples?

I also created #262, but that's not a blocker for this PR.

WeirdConstructor:
I've properly formatted the tables in the comments.

WeirdConstructor:
> Huh! I did not know the `bsp_pins` macro supported aliases!

Yes, neither did I see that, it was a suggestion by @9names!

> I like this idea in principle (subject to the Markdown tables being formatted correctly for the benefit of those reading the source code directly). Do we have any examples of using the aliases, or would be simplify any of the existing examples?

Unfortunately I did not yet see an application in my examples yet. But the `pico_uart_irq_echo.rs` example could benefit from this probably. PWM seems to work with `hal::pwm::Slices` and not use the pins directly.

thejpster (approved):

thejpster:
@9names, does this look good to you?

9names:
Yep, looks good to me!

PR #260, opened 2022-01-09
Add Raspberry Pi Pico WS2812 LED example

WeirdConstructor:
Here is the example for WS2812 LED, like the other example tested with hardware of course, with comments and ASCII diagram :-)

### DIFF
--- a/boards/rp-pico/examples/pico_ws2812_led.rs
+++ b/boards/rp-pico/examples/pico_ws2812_led.rs
@@ -1,0 +1,224 @@
+//! # Pico WS2812 RGB LED Example
+//!
+//! Drives 3 WS2812 LEDs connected directly to the Raspberry Pi Pico.
+//! This assumes you drive the Raspberry Pi Pico via USB power, so that VBUS
+//! delivers the 5V and at least enough amperes to drive the LEDs.
+//!
+//! For a more large scale and longer strips you should use an extra power
+//! supply for the LED strip (or know what you are doing ;-) ).
+//!
+//! The example also comes with an utility function to calculate the colors
+//! from HSV color space. It also limits the brightness a bit to save a
+//! few amperes.
+//!
+//! The example assumes you connected the data input to pin 6 of the
+//! Raspberry Pi Pico, which is GPIO4 of the rp2040. Here is a circuit
+//! diagram that shows the assumed setup:
+//!
+//! ```text
+//!  _______________      /----------------------\
+//! |+5V  /---\  +5V|----/         _|USB|_       |
+//! |DO <-|LED|<- DI|-\           |1  R 40|-VBUS-/
+//! |GND  \---/  GND|--+---\      |2  P 39|
+//!  """""""""""""""   |    \-GND-|3    38|
+//!                    |          |4  P 37|
+//!                    |          |5  I 36|
+//!                    \------GP4-|6  C   |
+//!                               |7  O   |
+//!                               |       |
+//!                               .........
+//!                               |20   21|
+//!                                """""""
+//! Symbols:
+//!     - (+) crossing lines, not connected
+//!     - (o) connected lines
+//! ```
+//!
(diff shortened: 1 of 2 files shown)


boards/rp-pico/examples/pico_ws2812_led.rs
```diff
@@ -0,0 +1,224 @@
...
+    );
+
+    // Setup a delay for the LED blink signals:
+    let mut frame_delay =
+        cortex_m::delay::Delay::new(core.SYST, clocks.system_clock.freq().integer());
+
+    // Import the `sin` function for a smooth hue animation from the
+    // Pico rp2040 ROM:
+    let sin = unsafe {
+        core::mem::transmute::<_, fn(f32) -> f32>(rp_pico::hal::rom_data::float_funcs::fsin())
```
thejpster, line 122: Why does this need a transmute?

boards/rp-pico/examples/pico_ws2812_led.rs
```diff
@@ -0,0 +1,224 @@
...
+        sm0,
+        clocks.peripheral_clock.freq(),
+        timer.count_down(),
+    );
+
+    let mut leds: [RGB8; STRIP_LEN] = [(0, 0, 0).into(); STRIP_LEN];
+    let mut t = 0.0;
+
+    // Bring down the overall brightness of the strip to save
+    // a few amperes (every LED draws ~60mA, RGB means 3 LEDs per
```
thejpster, line 147: Shouldn't this be 'a few milliamps'? An Ampere = An Amp = a enough current to blow the board up! Also applies to the comment at the top of the file.

boards/rp-pico/examples/pico_ws2812_led.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+}
+
+pub fn hsv2rgb(hue: f32, sat: f32, val: f32) -> (f32, f32, f32) {
+    let c = val * sat;
+    let v = (hue / 60.0) % 2.0 - 1.0;
+    let v = if v < 0.0 { -v } else { v };
+    let x = c * (1.0 - v);
+    let m = val - c;
+    let (r_, g_, b_) = if (0.0..60.0).contains(&hue) {
```
thejpster, line 199: This might be clearer as `if hue < 60.0 { ... ` (same for each branch).
Also, I wasn't sure why it was `r_` and not just `r`? Underscores make me think of marking variables as unused, although this is at the wrong end for that.

thejpster (changes requested):
Thank you for this PR! I'm sure many people will be happy to see this example added.

Just a couple of nits I noticed whilst reading though - alternations or clarifiying comments would be fine.

WeirdConstructor:
I've cleaned up the hsv2rgb coding style. I've no idea where I got that transmute from, it seems to work fine without! And I also changed the comments to express my intent a bit better.

boards/rp-pico/examples/pico_ws2812_led.rs
WeirdConstructor: I've really no idea where I got that snippet from. Fixed that, even the unsafe was useless.

boards/rp-pico/examples/pico_ws2812_led.rs
WeirdConstructor: Yea, my intent with the comments was more about warning the user that they can get quickly into ampere range if they scale up the number of LEDs.

boards/rp-pico/examples/pico_ws2812_led.rs
WeirdConstructor: Absolutely, I kinda went from `hue >= 0.0 && hue < 60.0` to something that clippy showed me and did not ask further. Obviously the if cascade demands a much simpler logic X-)

thejpster (approved):

Issue #246, opened 2021-12-23
Add extra humans to our crates, and fix team name


thejpster:
Only named humans can 'own' a crate, and hence have permissions to add or remove other owners.

'Team' membership merely grants publish and yank rights.

I propose we ensure that we always have at least two active humans listed against any crate, so they can add or remove teams as required in the future.

We also want to switch from `github:rp-rs:rp-rs` to `github:rp-rs:admins`.

These are the crates:

| Name | Human Owner | Team | Fixed |
|-------|---------------|-----|------|
| rp2040-pac | @eolder | rp-rs | ❌ | 
| rp2040-hal | @eolder | rp-rs | ❌ | 
| pio | @thejpster @Luminarys | admins | ✅ |
| rp2040-boot2 | @thejpster @jonathanpallant | admins | ❌ |
| rp-pico | @thejpster @jannic  | admins |  ✅ |

thejpster:
Pinging @eolder for comment

eolder:
Updated!

thejpster:
I removed jonathanpallant from boot2 and invited eolder. Nearly done!

thejpster:
Although I suppose there are the BSPs to consider

9names:
pio-proc and pio-parser are presently owned exclusively by @thejpster.
the admin team should get added to the owners on crates.io, too.

jonathanpallant:
Ah, good catch

thejpster:
Team added. I need a human to add too.

jannic:
There are still a few crates without a second human owner.

Also, I just noticed that I was the only owner for some BSP crates which where first uploaded by my when during the recent HAL+BSP releases. ([pimoroni-servo2040](https://crates.io/crates/pimoroni-servo2040), [waveshare-rp2040-lcd-0-96](https://crates.io/crates/waveshare-rp2040-lcd-0-96), [waveshare-rp2040-zero](https://crates.io/crates/waveshare-rp2040-zero))

I added the team to those crates. A second human is still missing.

PR #261, opened 2022-01-09
Add Solder Party RP2040 Stamp board

jkristell:
Add support for the [RP2040 Stamp board](https://www.solder.party/docs/rp2040-stamp/) by Solder Party.

This board has a single "device" - a neopixel. The example I included for this is copied and updated from the feather board example.

### DIFF
--- a/boards/solderparty-rp2040-stamp/Cargo.toml
+++ b/boards/solderparty-rp2040-stamp/Cargo.toml
@@ -1,0 +1,31 @@
+[package]
+name = "solderparty-rp2040-stamp"
+version = "0.1.0"
+authors = ["The rp-rs Developers"]
+edition = "2018"
+homepage = "https://github.com/rp-rs/rp-hal/tree/main/boards/solderparty-rp2040-stamp"
+description = "Board Support Package for the Solder Party RP2040 Stamp"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/rp-rs/rp-hal.git"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+cortex-m = "0.7.2"
+rp2040-boot2 = { version = "0.2.0", optional = true }
+rp2040-hal = { path = "../../rp2040-hal", version = "0.4.0"}
+cortex-m-rt = { version = "0.7", optional = true }
+
+[features]
+default = ["rt", "boot2"]
+boot2 = ["rp2040-boot2"]
+rt = ["cortex-m-rt","rp2040-hal/rt"]
+
+[dev-dependencies]
+panic-halt= "0.2.0"
+embedded-hal ="0.2.5"
+nb = "1.0.0"
+smart-leds = "0.3.0"
+pio = "0.1.0"
+ws2812-pio = { git = "https://github.com/ithinuel/ws2812-pio-rs", rev = "4f0d81e594ea9934f9c4c38ed9824ad0cce4ebb5" }
+embedded-time = "0.12.0"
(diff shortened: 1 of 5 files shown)


boards/solderparty-rp2040-stamp/README.md
```diff
@@ -0,0 +1,94 @@
...
+[solderparty-rp2040-stamp]: https://github.com/rp-rs/rp-hal/tree/main/boards/solderparty-rp2040-stamp
+[rp2040-hal]: https://github.com/rp-rs/rp-hal/tree/main/rp2040-hal
+[Raspberry Silicon RP2040]: https://www.raspberrypi.org/products/rp2040/
+
+## Using
+
+To use this crate, your `Cargo.toml` file should contain:
+
+```toml
+solderparty-rp2040-stamp = "0.1.0"
```
thejpster, line 19: This will need to be a git url up until 0.1.0 is published.

boards/solderparty-rp2040-stamp/README.md
```diff
@@ -0,0 +1,94 @@
+# [solderparty-rp2040-stamp] - Board Support for the [Solder Party RP2040 Stamp]
+
+You should include this crate if you are writing code that you want to run on
+an [Solder Party RP2040 Stamp]
```
thejpster, line 4: /a/ [Solder Party RP2040 Stamp]

boards/solderparty-rp2040-stamp/examples/solderparty_stamp_neopixel_rainbow.rs
```diff
@@ -0,0 +1,96 @@
...
+        &mut pac.RESETS,
+    );
+
+    let timer = Timer::new(pac.TIMER, &mut pac.RESETS);
+    let mut delay = timer.count_down();
+
+    // Configure the addressable LED
+    let (mut pio, sm0, _, _, _) = pac.PIO0.split(&mut pac.RESETS);
+    let mut ws = Ws2812::new(
+        // The onboard NeoPixel is attached to GPIO pin #16 on the Feather RP2040.
```
thejpster, line 60: Should probably change this comment to whatever the Stamp uses

thejpster (changes requested):
Looks good - just a couple of minor nits.

Have you tested this on a Stamp, or was this done simply by referring to the datasheet?

thejpster:
Paging @ithinuel who I think had some Solder Party hardware incoming.

boards/solderparty-rp2040-stamp/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "solderparty-rp2040-stamp"
+version = "0.1.0"
+authors = ["Johan Kristell <johan.kristell@gmail.com>"]
```
thejpster, line 4: Are you OK to change this to "The rp-rs Developers", as per #246?

jkristell:
Thanks for the review. I updated the branch with your suggestions.

Should have mention that it was tested on a Stamp using the carrier board from Solder Party.

jkristell:
After thinking some more about this, maybe it makes more sense to support the carrier board?: https://www.solder.party/docs/rp2040-stamp/carrier/

The only difference, as far as I can tell, is the naming of the pins, and maybe some pins missing on the carrier board. 

Alternatively, support both, with this crate, with the help of feature flags and default to the carrier feature on.

What do you think?

ithinuel:
Actually, without knowing I was about to do something with it some people talked to me about this board how they were excited about the stamp form factor and all the pins it exposed. So I'd say that we should definitely expose the pins for the stamp.

Using a feature to reduce the scope/pin naming to the carrier board works for me too 😃

thejpster (approved):

thejpster:
I feel like I need to go and rationalise all these colour wheel examples now to use a common colour wheel crate!

PR #263, opened 2022-01-12
Fix alarm schedule race condition

WeirdConstructor:
Hi!

Mathias on the rp-rs Matrix channel remarked that there is a race condition in the schedule() function for Timer alarms: 

image
(Please excuse my humor :-) )

I wrapped the critical section in a `cortex_m::interrupt::free(|_| { })`. Mathias said that it would be better for hard real time to add a check instead, whether the alarm time was already reached. But then figured that it would be quite hard to actually trigger the (right) alarm interrupt from there afterwards.

So using `interrupt::free` seems to be the best solution for now and gets rid of the race condition.


thejpster (approved):

thejpster:
Thanks for the PR!

PR #264, opened 2022-01-12
Pico i2c oled text display ssd1306 example

WeirdConstructor:
Hi,

here is another example to get newcomers started with Rust on the Raspberry Pi Pico. It's a for a more or less regular 128x64 OLED Display you can find either on AliExpress or Amazon.

I hope it's okay adding lots of examples, I know that these help a great deal when trying to get familiar with the embedded Rust ecosystem.

### DIFF
--- a/boards/rp-pico/examples/pico_i2c_oled_display_ssd1306.rs
+++ b/boards/rp-pico/examples/pico_i2c_oled_display_ssd1306.rs
@@ -1,6 +1,6 @@
-//! # Pico (monochome) OLED Display with SSD1306 Driver Example
+//! # Raspberry Pi Pico (monochome) 128x64 OLED Display with SSD1306 Driver Example
 //!
-//! This example assumes you got an OLED Display with an SSD1306 driver
+//! This example assumes you got an 128x64 OLED Display with an SSD1306 driver
 //! connected to your Raspberry Pi Pico. The +3.3V voltage source of the
 //! Raspberry Pi Pico will be used, and the output pins 21 and 22 of the board
 //! (on the lower right).
@@ -19,13 +19,13 @@
 //!   _|USB|_         |    /-----\  |  |  /--------+--\
 //!  |1  R 40|        |   /    __|__|__|__|___     |  |
 //!  |2  P 39|        |  /    | ____________  |    |  |
-//!  |3    38|- GND --|-/     | |Hello worl|  |    |  |
+//!  |3    38|- GND --+-/     | |Hello worl|  |    |  |
 //!  |4  P 37|        |       | |Hello Rust|  |    |  |
 //!  |5  I 36|-+3.3V -/       | |counter: 1|  |    |  |
 //!  |6  C   |                | |          |  |    |  |
 //!  |7  O   |                | """"""""""""  |    |  |
 //!  |       |                 """""""""""""""     |  |
-//!  |       |              (SSD1306 OLED Display) |  |
+//!  |       |       (SSD1306 128x64 OLED Display) |  |
 //!  .........                                     /  /
 //!  |       |                                    /  /
 //!  |     22|-GP17 I2C0 SCL---------------------/  /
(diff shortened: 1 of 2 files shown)


boards/rp-pico/examples/pico_i2c_oled_display_ssd1306.rs
```diff
@@ -0,0 +1,219 @@
...
+        pac.PADS_BANK0,
+        sio.gpio_bank0,
+        &mut pac.RESETS,
+    );
+
+    // Configure two pins as being I²C, not GPIO
+    let sda_pin = pins.gpio16.into_mode::<hal::gpio::FunctionI2C>();
+    let scl_pin = pins.gpio17.into_mode::<hal::gpio::FunctionI2C>();
+
+    // Create the I²C drive, using the two pre-configured pins. This will fail
```
thejpster, line 125: I²C driver?

boards/rp-pico/examples/pico_i2c_oled_display_ssd1306.rs
```diff
@@ -0,0 +1,219 @@
...
+// and the drawing routines:
+use embedded_graphics::{
+    mono_font::{ascii::FONT_9X18_BOLD, MonoTextStyleBuilder},
+    pixelcolor::BinaryColor,
+    prelude::*,
+    text::{Baseline, Text},
+};
+
+// The display driver:
+use ssd1306::{prelude::*, I2CDisplayInterface, Ssd1306};
```
thejpster, line 76: I don't think I'd import I2CDisplayInterface here as it's not clear where it comes from just by the name alone. We only refer to it once so calling it ssd1306::I2CDisplayInterface I think is actually more readable.

boards/rp-pico/examples/pico_i2c_oled_display_ssd1306.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    fn as_str(&self) -> &str {
+        core::str::from_utf8(&self.buf[0..self.ptr]).unwrap()
+    }
+}
+
+impl core::fmt::Write for FmtBuf {
+    fn write_str(&mut self, s: &str) -> core::fmt::Result {
+        let len = s.len();
+        self.buf[self.ptr..(self.ptr + len)].copy_from_slice(s.as_bytes());
```
thejpster, line 213: This panics on buffer overflow. We should either:

a) make it clear in the comments it will do this
b) truncate text that doesn't fit, or
c) use a heapless::String (which implements core::fmt::Write)

boards/rp-pico/examples/pico_i2c_oled_display_ssd1306.rs
```diff
@@ -0,0 +1,219 @@
...
+    let timer = hal::Timer::new(pac.TIMER, &mut pac.RESETS);
+    let mut delay = timer.count_down();
+
+    let mut count = 0;
+
+    let mut buf = FmtBuf::new();
+    loop {
+        buf.reset();
+        // Format some text into a static buffer:
+        core::fmt::write(&mut buf, format_args!("counter: {}", count)).unwrap();
```
thejpster, line 160: I'd use write! here

thejpster (changes requested):
Excellent! Always good to see an embedded-graphics example. Just some minor notes but otherwise very good.

boards/rp-pico/examples/pico_i2c_oled_display_ssd1306.rs
WeirdConstructor: Makes absolutely sense.

boards/rp-pico/examples/pico_i2c_oled_display_ssd1306.rs
WeirdConstructor: yes, thanks :-)

boards/rp-pico/examples/pico_i2c_oled_display_ssd1306.rs
WeirdConstructor: I'll truncate it. Don't want to drag in another dependency just for this. And it should not crash, because people are going to copy & paste it (I know myself).

thejpster (approved):

PR #265, opened 2022-01-12
embassy-traits: Fix unexpected restart on repeated read.

ithinuel:
I noticed that it was always generating a restart condition rather than only reading a byte.
This fixes that issue.


9names (approved):
Ah, you can even see where you intended to handle this in the existing code!
LGTM.

PR #268, opened 2022-01-18
Added `impl Default` for UartConfig

### LINKED ISSUE #253: UartConfig cannot be constructed
Commit bdfb4d82c9c9d1d2f9f23c45f4d03132e7f2b962 made the `UartConfig` `#[non_exhaustive]`. This means that users of this crate cannot actually construct an instance of this type.

There are a couple of solutions for this that I can see:
1. Remove the `#[non_exhaustive]`
1. Add `impl Default`, with the following defaults:
   ```rs
   impl Default for UartConfig {
       fn default() -> Self {
           Self {
               baudrate: Baud(0),
               data_bits: DataBits::Eight,
               stop_bits: StopBits::One,
               parity: None,
           }
       }
   }
   ```
1. Expect people to take a `uart::common_configs` instance and modify that. e.g:
    ```rs
    let mut config = common_configs::_115200_8_N_1;
    config.baudrate = Baud(0);
    ```
1. Introduce a builder pattern  (Probably not preferred, but still worth a mention)

VictorKoenders:
Closes #253

### DIFF
--- a/rp2040-hal/src/uart/utils.rs
+++ b/rp2040-hal/src/uart/utils.rs
@@ -61,7 +61,8 @@ pub enum Parity {
 /// A struct holding the configuration for an UART device.
 ///
 /// The `Default` implementation implements the following values:
-/// ```
+/// ```ignore
+/// # // can't actually create this with the non_exhaustive attribute
 /// UartConfig {
 ///    baudrate: Baud(0),
 ///    data_bits: DataBits::Eight,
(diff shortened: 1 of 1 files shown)


thejpster:
Can we set the default Baud rate to a valid value - like 115,200 bps?

thejpster (approved):

thejpster:
Thank you for the PR!

PR #270, opened 2022-01-19
WIP: port to new PAC spinlock syntax

9names:
Did the minimal work to get the spinlocks working with the new arrayified registers.
The change in macro style was only because that was the easiest thing for me to get working.

We will want to get a new PAC release out before merging this.

### DIFF
--- a/rp2040-hal/src/sio.rs
+++ b/rp2040-hal/src/sio.rs
@@ -231,7 +231,7 @@ pub trait Spinlock: typelevel::Sealed + Sized {
 }
 
 macro_rules! impl_spinlock {
-    ($($spinlock_name:ident,$spinlock_num:literal)*) => {
+    ($($spinlock_name:ident => $spinlock_num:literal,)*) => {
         $(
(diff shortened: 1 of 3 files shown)


thejpster (approved):

rp2040-hal/src/sio.rs
```diff
@@ -276,46 +277,46 @@ macro_rules! impl_spinlock {
...
+    Spinlock22 => 22,
+    Spinlock23 => 23,
+    Spinlock24 => 24,
+    Spinlock25 => 25,
+    Spinlock26 => 26,
+    Spinlock27 => 27,
+    Spinlock28 => 28,
+    Spinlock29 => 29,
+    Spinlock30 => 30,
+    Spinlock31 => 31,
```
VictorKoenders, line 319: I wonder if spinlock 31 should be removed here as it's used internally. I can't imagine people needing more than 30 spinlocks

9names:
Closed in favour of #281

PR #272, opened 2022-01-20
Set hw reset bits for watchdog

9names:
Initial work to get watchdog working properly.
Need to work out what is sufficient.
Currently using the same flags from the [pico bootrom](https://github.com/raspberrypi/pico-bootrom/blob/7a751f28436a8ee4849ac7799eb3aa1457e6f091/usb_device_tiny/runtime.c#L70)


9names:
The SDK [also uses the same flags](https://github.com/raspberrypi/pico-sdk/blob/2062372d203b372849d573f252cf7c6dc2800c0a/src/rp2_common/hardware_watchdog/watchdog.c#L40), and everyone else calls it through the SDK so I'm going to assume these are sufficient.

thejpster (approved):

PR #276, opened 2022-01-25
Fix i2c peripheral mode state machine

ithinuel:
The `start_det` flag is set as soon as a start condition occurs whereas `rd_req` flag is only set once the address byte is received.


9names (approved):
LGTM

PR #278, opened 2022-01-27
Implement intrinsics for the hardware divider

Sizurka:
This implements intrinsics for the hardware divider so normal operators use it instead of the default software implementations.  So with this applied `/` and `%` operations should make use of the hardware divider.  This also makes access to the divider interrupt safe by implementing the save/restore logic used by the Pico SDK. 

Unsurprisingly, there's quite a bit of room for optimization here, when you compare the generated assembly to that hand optimized version in the Pico SDK.  However, in my testing it still results in about a 60% speedup compared to the default software implementations.

To be honest, I'm not completely sure why the override works correctly: the functions defined by [compiler-builtins intrinsics macro](https://github.com/rust-lang/compiler-builtins/blob/ea0cb5b589cc498d629c545e9bae600301ba6aed/src/macros.rs#L61) don't look like they're defined as weak symbols.   However, the generated assembly does the correct thing and the new definitions are used instead.  So this is probably just my failure to understand the whole link chain being used or something about Rust's symbol resolution order, but it does make me a little nervous.

There is also a bit of an oddity in the signed modulus operators: the compiler generates code that uses `__aeabi_idivmod` instead of a more direct `__modsi3` while `__aeabi_idivmod` [isn't defined as a weak symbol](https://github.com/rust-lang/compiler-builtins/blob/ea0cb5b589cc498d629c545e9bae600301ba6aed/src/arm.rs#L81) so we can't override it, and even it if was it'd require a bit of nasty implementation due to the odd way it returns results (remainder in `r1`).  So the modulus operator ends up doing an implied multiply after the division, this isn't a huge deal though since we do have hardware multiplication.  The same applies to `__aeabi_uidivmod` but that at least calls `__udivmodsi4` directly.

### References

[Unsigned intrinsics](https://github.com/rust-lang/compiler-builtins/blob/ea0cb5b589cc498d629c545e9bae600301ba6aed/src/int/udiv.rs)
[Signed intrinsics](https://github.com/rust-lang/compiler-builtins/blob/ea0cb5b589cc498d629c545e9bae600301ba6aed/src/int/sdiv.rs)
[Pico SDK divider](https://github.com/raspberrypi/pico-sdk/blob/2062372d203b372849d573f252cf7c6dc2800c0a/src/rp2_common/pico_divider/divider.S)
[Pico SDK save/restore](https://github.com/raspberrypi/pico-sdk/blob/2062372d203b372849d573f252cf7c6dc2800c0a/src/rp2_common/hardware_divider/include/hardware/divider_helper.S)

### DIFF
--- a/rp2040-hal/src/sio.rs
+++ b/rp2040-hal/src/sio.rs
@@ -215,6 +215,21 @@ where
     }
 }
 
+// Don't use cortex_m::asm::delay(8) because that ends up delaying 15 cycles
+// on Cortex-M0.  Each iteration of the inner loop is 3 cycles and it adds
+// one extra iteration.
+#[inline(always)]
+fn divider_delay() {
+    cortex_m::asm::nop();
+    cortex_m::asm::nop();
+    cortex_m::asm::nop();
+    cortex_m::asm::nop();
+    cortex_m::asm::nop();
+    cortex_m::asm::nop();
+    cortex_m::asm::nop();
+    cortex_m::asm::nop();
+}
+
 impl HwDivider {
     /// Perform hardware unsigned divide/modulo operation
     pub fn unsigned(&self, dividend: u32, divisor: u32) -> DivResult<u32> {
@@ -222,7 +237,7 @@ impl HwDivider {
             sio.div_udividend.write(|w| unsafe { w.bits(dividend) });
             sio.div_udivisor.write(|w| unsafe { w.bits(divisor) });
 
-            cortex_m::asm::delay(8);
+            divider_delay();
 
             // Note: quotient must be read last
             let remainder = sio.div_remainder.read().bits();
@@ -243,7 +258,7 @@ impl HwDivider {
             sio.div_sdivisor
                 .write(|w| unsafe { w.bits(divisor as u32) });
 
-            cortex_m::asm::delay(8);
+            divider_delay();
 
             // Note: quotient must be read last
             let remainder = sio.div_remainder.read().bits() as i32;
--- a/rp2040-hal/src/sio.rs
+++ b/rp2040-hal/src/sio.rs
@@ -230,45 +230,141 @@ fn divider_delay() {
     cortex_m::asm::nop();
 }
 
+fn divider_unsigned(dividend: u32, divisor: u32) -> DivResult<u32> {
+    save_divider(|sio| {
+        sio.div_udividend.write(|w| unsafe { w.bits(dividend) });
+        sio.div_udivisor.write(|w| unsafe { w.bits(divisor) });
+
+        divider_delay();
+
+        // Note: quotient must be read last
+        let remainder = sio.div_remainder.read().bits();
+        let quotient = sio.div_quotient.read().bits();
+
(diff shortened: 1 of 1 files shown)


jonathanpallant:
Thank you for the PR! I've wanted this for a while :)

I've opened https://github.com/rust-lang/compiler-builtins/issues/451 to see what's the best way to do this from their point of view. Like you, I am nervous this only currently works due to specific linker order, or other funnies I don't fully understand.

bjorn3:
> To be honest, I'm not completely sure why the override works correctly: the functions defined by compiler-builtins intrinsics macro don't look like they're defined as weak symbols. However, the generated assembly does the correct thing and the new definitions are used instead. So this is probably just my failure to understand the whole link chain being used or something about Rust's symbol resolution order, but it does make me a little nervous.

This works because libcompiler_builtins.rlib is an archive with a single object file for each individual intrinisic. The linker by default only pulls in object files from an archive if any of the symbols in it is needed. If an earlier object file already defined all symbols needed from an object in an archive, this object is skipped without giving a symbol conflict. Only if at least one symbol is needed from the object in an archive will the linker give an error on conflicting definitions. On non-standalone targets many of these intrinsics are defined by both libgcc_s (or libcompiler_rt for clang I think) and libcompiler_builtins. I don't think rustc guarantees which one will win, but it has to guarantee that it works at all.

Sizurka:
Ah that makes sense.  I wasn't aware of there being special magic when generating the compiler-builtins object files.  Thanks for the explanation.  

It also explains I couldn't override `__aeabi_idivmod` without weak linkage: it's not wrapped into its own module so its part of an object that gets pulled in for the other symbols.

So that does make it dependent on link order, although a relatively safe assumption about it (toolchain libraries linked last).  It does make it relatively fragile if you try to break it: adding an explicit dependency on compiler-builtins gets link errors because it ends up getting linked before rp-hal.  Hopefully there's relatively easy answer for the proper way to add a platform specific implementation.

jonathanpallant:
OK, so `compiler-builtins` (https://github.com/rust-lang/compiler-builtins/issues/451#issuecomment-1026237516) says that the canonical approach is just to define the functions yourself. `rustc` will always link `compiler-builtins` last and as each function is in its own object file, they just get pushed out.

jonathanpallant:
Also, any functions in compiler-builtins which aren't in their own object file is probably a bug that they can fix.

thejpster:
Can I get a second pair of eyes on this? @9names ?

9names (commented):
Code looks good to me.
__udivsi3 and __divsi3 are getting invoked implicitly as you would expect.
It's passing a stress test I wrote that invokes this from both cores + interrupts (to test the dirty save / restore functionality).
I haven't exhaustively tested different inputs / outputs for correctness, but there's not really any edge cases to look for in integer div/mod - just division by zero? And that's still being caught before it hits the hardware, so I think we're good.

thejpster (approved):

thejpster:
Ok let's go! Thank you @Sizurka !

PR #280, opened 2022-01-30
Use atomic operations for altering Timer alarm interrupts

9names:
The existing Alarm implementation uses register.modify() calls to access the interrupt manipulation registers.
To make this a little safer, the original implementer made them all take &mut Timer to ensure there wasn't a data race when performing that RMW.
That's great for safety, but bad for ergonomics.

This PR uses the atomic_clear and atomic_set functions to avoid these data races.
Functionality is largely the same as the PWM interrupt interactions, example:
https://github.com/rp-rs/rp-hal/blob/5771f872f2d08af821ea405d09d12e3d404996b5/rp2040-hal/src/pwm/mod.rs#L385-L391


9names:
Example program using this newer syntax:
https://gist.github.com/9names/56d5fc0c9757122c27ca1cca64203dda

thejpster (approved):

9names:
Probably worth giving people some warning when this gets merged + released, since it is an API change.

PR #283, opened 2022-01-30
Improve ROM function access

### LINKED ISSUE #238: Provide faster access to ROM functions
Our current method of accessing ROM functions involves a ROM table walk to find the correct function first.
A few options we could pursue:
- Store a function-pointer table in flash for each variant of the ROM - check ROM version on each call then jump to correct table
- Static function pointer in each of the wrapper functions, initialized by ROM table walk on first use
- Reserve space in linker file for RAM table, initialize entire table in rt, [#entry], or first use

Sizurka:
First, this standardizes ROM function access between the floating point and normal functions.  It makes the direct call (e.x. `hal::rom_data::popcount32(0xFF)`) do the operation, while also adding a module with the same name as the function and `ptr()` function to get a function pointer for local caching.  So there are now two ways to access every ROM function:

```rust
let result = hal::rom_data::popcount32(0xFF);
let pc32 = hal::rom_data::popcount32::ptr();
let result = pc32(0xFF);
```

This represents an API change for the floating point functions, but a simple one: the previous API is just the `::ptr()` call.

Next, it implements a crate feature (`rom-cache`) to enable automatic caching via static variables of the ROM table lookups.  The actual savings varies based on the function call (this is because the table lookup function [is looping over the ROM table](https://github.com/raspberrypi/pico-bootrom/blob/ef22cd8ede5bc007f81d7f2416b48db90f313434/bootrom/bootrom_rt0.S#L64)).  However, for example, this saves around 22 cycles for subsequent calls to floating point functions, in exchange for a bit more flash, 2 bytes of RAM per used function, and about 10 extra cycles on the first access (plus the table walk).  Even for things at the start of the table (`popcount32`) it still saves 4 cycles.

Closes #238

### DIFF
--- a/boards/rp-pico/examples/pico_ws2812_led.rs
+++ b/boards/rp-pico/examples/pico_ws2812_led.rs
@@ -119,7 +119,7 @@ fn main() -> ! {
 
     // Import the `sin` function for a smooth hue animation from the
     // Pico rp2040 ROM:
-    let sin = rp_pico::hal::rom_data::float_funcs::fsin();
+    let sin = rp_pico::hal::rom_data::float_funcs::fsin::ptr();
 
     // Create a count down timer for the Ws2812 instance:
     let timer = Timer::new(pac.TIMER, &mut pac.RESETS);
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -53,3 +53,4 @@ rt = ["rp2040-pac/rt"]
 #
 # embassy-traits = ["embassy_traits", "futures"]
 alloc = []
+rom-func-cache = []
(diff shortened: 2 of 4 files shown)


thejpster (changes requested):
This looks great. Thank you!

My only suggestion is maybe changing the feature to rom-func-cache. And maybe using an Atomic U16 to avoid the unsafe static mut (and although it *should* do an atomic 16 bit write with a strh I'm not sure that's guaranteed and if it did two byte writes your function would be hazardous because after one byte it's non zero but also wrong).

Sizurka:
Thanks for the review.  I've updated to use atomics for the cached pointers and changed the feature name.

I also changed the documentation wording for the floating point functions, since it was somewhat wrong after the change, now that I look at it.  The base function does the operation now, so the documentation should reflect that; so this is just removing the "Returns a function that will..." from all of them.

thejpster:
Great, thanks!

One last observation - AtomicX types don't need to be mut :)

Sizurka:
Fixed; I'd just assumed they needed outer mutability without actually checking.

thejpster:
:tada:

PR #285, opened 2022-02-01
example of side setting using PIO

astraw:
Hi, thanks for the great project. I would like to suggest this example, a slight modification of `rp2040-hal/examples/pio_proc_blink.rs` which implements side setting on the PIO. I would have found this example useful because I didn't realize that there was a different way to enable the pins for this case https://github.com/rp-rs/pio-rs/issues/18.


thejpster (approved):

thejpster:
This looks great. Thank you for the PR!

It looks like rustfmt failed - would you mind checking that and doing a cargo fmt if required?

astraw:
Thanks. I now ran `cargo fmt` and pushed a new commit.

thejpster (approved):

thejpster:
Thank you!

PR #291, opened 2022-02-10
Avoid 64bit division

jannic:
While looking at the code generated by https://github.com/rp-rs/rp-hal/pull/288, I wondered why there were references to 64bit division code (`compiler_builtins::int::specialized_div_rem::u64_div_rem`) for a basically empty firmware.

Turned out that `init_clocks_and_plls()` contained 64bit integer divisions.

As the code to do those divisions takes about 1kB even in fully optimized builds, I changed the clock calculation code to use 32bit divisions, instead.

To keep the code small, I accepted some minimal additional rounding error: In case the remainder of the division is bigger than 2^24, the lower 8 bits will be thrown away, causing a relative rounding error of at most 2^-16.
(To be exact, instead of the expected rounding-down of normal integer division, the value might get rounded up, instead. The resulting difference from the true value might be even smaller than before.)

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -48,3 +48,7 @@ jobs:
         with:
           command: test
           args: --doc --target x86_64-unknown-linux-gnu --features chrono
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --tests --target x86_64-unknown-linux-gnu
--- a/rp2040-hal/CHANGELOG.md
+++ b/rp2040-hal/CHANGELOG.md
@@ -13,7 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Updated embedded-hal alpha support to version 1.0.0-alpha.7
+- Update embedded-hal alpha support to version 1.0.0-alpha.7
+- Avoid 64-bit division in clock calculations
 
 ## [0.3.0] - 2021-12-19
 
--- a/rp2040-hal/src/clocks/macros.rs
+++ b/rp2040-hal/src/clocks/macros.rs
@@ -180,19 +180,13 @@ macro_rules! clock {
 
                 #[doc = "Configure `"$name"`"]
                 fn configure_clock<S: ValidSrc<$name>>(&mut self, src: &S, freq: Hertz) -> Result<(), ClockError>{
-                    let src_freq: Hertz<u64> = src.get_freq().into();
+                    let src_freq: Hertz<u32> = src.get_freq().into();
 
                     if freq.gt(&src_freq){
                         return Err(ClockError::CantIncreaseFreq);
                     }
 
-                    // Div register is 24.8) int.frac divider so multiply by 2^8 (left shift by 8)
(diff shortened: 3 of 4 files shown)


9names:
I tested this after rebasing onto the new main, since the new intrinsics code saves a decent chunk of code space too.

On a trivial project I tested on it's clearly a saving (240 bytes dev, 944 bytes release).
With a more complex project (defmt enabled, lots of printing, approx 22KB release and 27KB dev), it saved 288 bytes on dev build but adds 272 bytes in release (because it's no longer using the u64_div_rem that was already included due to math done elsewhere).

It's a trade-off. 272 bytes isn't a big chunk out of the 2MB on the Pico but it is *something*.
On the other hand, 944 bytes is a huge amount for a small firmware - my small project went from 5KB to 4KB - that *might* be a big deal if you were running entirely from RAM.

jannic:
Those 272 are a real concern, as this change is all about saving a few bytes of flash.

How common is 64 bit division in embedded software? Having used ATmega before, even 32bit arithmetic sometimes feels like a luxury. :smile: If most real-world code ends up using it, the proposed optimization would be counterproductive.

In theory one could implement both approaches and let the user chose, but most users would not care at all, so having this choice would only be confusing.

thejpster (commented):
Some unit tests / doctests on fractional_div would be useful I think, to test all the corner cases.

thejpster:
I think I'm happy with this though. 64 bit maths is fairly uncommon so most people benefit.

I wonder though if we can use const fn to make this all go away though? I mean, the crystal frequency is always known and the sysclk is almost always known at compile time.

jannic:
> Some unit tests / doctests on fractional_div would be useful I think, to test all the corner cases.

I agree. I did some tests myself by pasting both the old and the new code into the same binary, throwing millions of random numbers at them, and comparing the results - and caught some subtle bugs in earlier versions of that patch.
Probably too much for a unit test, but having some tests is definitely a good idea.

> I wonder though if we can use const fn to make this all go away though?

I thought the same yesterday, when I saw https://github.com/embassy-rs/embassy/commit/640ddc948190fe9a24d6db616cb103d6a94f81f9.
But `init_clocks_and_plls()`/`configure_clock()` have side effects and so can't be a `const fn`. And inside those functions, the frequency is no longer a constant known in advance.
Some API like this could work, but would be ugly:
```
  const fn calculate_divider(freq: u32, source_freq: u32) -> u32 { ... }
  
  const DIVIDER = calculate_divider( ... );
  [...]
  some_clock.configure( DIVIDER );
```
For `init_clocks_and_plls` one would need many dividers, so that would become even more ugly.
Perhaps the BSPs (where the crystal frequency can be assumed as constant) could provide some kind of `init_clocks_and_plls_to_max_frequency()` method which doesn't take any clock frequency parameters at all, so everything could be precalculated?

thejpster:
Latest changes look good

jannic:
It can be merged. I don't have actionable ideas on how to improve this further, at the moment.

9names:
Would you mind resolving the merge conflict on the changelog?

jannic:
Ok, done

thejpster:
Looks like clippy has a sad

9names (approved):

Issue #292, opened 2022-02-10
Spinlocks not cleared by reset


jannic:
When flashing a new firmware with `probe-run`, the automatically executed reset doesn't clear the spinlocks. Therefore, if (for example) the spinlock used for critical sections was locked at the moment when `probe-run` interrupted the running program, the new firmware will deadlock as soon as it tries to acquire the spinlock. The same is true for a `sys_reset()`.

Possible workaround: Adding `(*pac::SIO::ptr()).spinlock[number_of_spinlock].write(|w| w.bits(0));` at the top of `main`.

Small program demonstrating the issue:
```
use bsp::hal::sio::Spinlock0;

#[entry]
fn main() -> ! {
    //unsafe { (*pac::SIO::ptr()).spinlock[0].write(|w| w.bits(0)); }
    info!("Program start");
    let _l0 = Spinlock0::claim();
    info!("Lock claimed");
    cortex_m::asm::delay(10000);
    cortex_m::peripheral::SCB::sys_reset();
}
```
Is there a way to solve this cleanly without a custom `#[entry]` macro or some `#[pre_init]` code?

jannic:
This could be a way to solve this issue: https://github.com/jannic/rp2040-entry/blob/master/src/lib.rs#L32-L34

Basically, provide an alternative `#[entry]` attribute, which forwards to `#[cortex_m::entry]` but adds code to release all spinlocks at program start.

For simplicity, I just added another wrapper function. This becomes visible in backtraces:
```
stack backtrace:
   0: rp2040_project_template::__rp2040_entry_main
        at src/main.rs:75:5
   1: rp2040_project_template::__cortex_m_rt___rp2040_entry_main_trampoline
        at src/main.rs:59:1
   2: main
        at src/main.rs:59:1
   3: Reset
```
It would probably be possible to inject the added code into `fn main()` or into the trampoline added by `cortex_m_rt`, instead.

jannic:
Updated the [example](https://github.com/jannic/rp2040-entry/) so it injects the unlocking code directly into `main()`.

thejpster:
Good catch!

I was already thinking about BSP entry functions, so a HAL entry is not unreasonable.

Does it get inlined in release mode using LTO?

jannic:
The latest version of the example directly injects the code into main, so there is no need to inline anything, and LTO doesn't matter.
[...]

PR #294, opened 2022-02-14
Add GPIO interrupt example

### LINKED ISSUE #275: Add interrupt-driven example for GPIO
There's a few examples for interrupts, but it would be good to have a nice simple one for GPIO to make getting started with them more approachable.

9names:
Add a GPIO interrupt example to the HAL
This was intended to be much more trivial, but it's hard to be simple, clear and safe at the same time.

Since this is intended to be the easiest guide to interrupts, please be critical of any complexity that might trip up beginners.
And if you are a beginner and you are reading this - your feedback is also very welcome, please comment!

Closes #275

### DIFF
--- a/rp2040-hal/examples/gpio_irq_example.rs
+++ b/rp2040-hal/examples/gpio_irq_example.rs
@@ -157,6 +157,7 @@ fn main() -> ! {
 
 #[interrupt]
 fn IO_IRQ_BANK0() {
+    // The `#[interrupt]` attribute covertly converts this to `&'static mut Option<LedAndButton>`
     static mut LED_AND_BUTTON: Option<LedAndButton> = None;
 
     // This is one-time lazy initialisation. We steal the variables given to us
--- a/rp2040-hal/examples/gpio_irq_example.rs
+++ b/rp2040-hal/examples/gpio_irq_example.rs
@@ -22,8 +22,6 @@
 #![no_std]
 #![no_main]
 
-use core::cell::RefCell;
-use cortex_m::interrupt::Mutex;
 // The macro for our start-up function
 use cortex_m_rt::entry;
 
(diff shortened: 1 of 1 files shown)


thejpster:
Fantastic!

I wonder if it would look better or worse using (LedPin, ButtonPin) as a type, with a single global static.

9names:
> I wonder if it would look better or worse using (LedPin, ButtonPin) as a type, with a single global static.

I think it looks much better, thanks for the suggestion!

rp2040-hal/examples/gpio_irq_example.rs
```diff
@@ -0,0 +1,187 @@
...
+    if LED_AND_BUTTON.is_none() {
+        cortex_m::interrupt::free(|cs| {
+            *LED_AND_BUTTON = GLOBAL_PINS.borrow(cs).take();
+        });
+    }
+
+    // Need to check if our Option<LedAndButtonPins> contains our pins
+    if let Some(gpios) = LED_AND_BUTTON {
+        // borrow led and button by *destructuring* the tuple
+        // these will be of type `&mut LedPin` and `&mut ButtonPin`, so we don't have
```
jannic, line 173: It took me about an hour to understand why those types would be mutable references.
Hint: It's because of magic behavior of the `#[interrupt]` attribute, which [converts](https://docs.rs/cortex-m-rt/latest/cortex_m_rt/attr.interrupt.html#properties) `static mut LED_AND_BUTTON: Option<LedAndButton>` to `let LED_AND_BUTTON: &mut Option<LedAndButton>`.

(And it probably would have taken me much longer, if I didn't work on the very similar `#[entry]` macro [during the last few days](https://github.com/rp-rs/rp-hal/issues/292#issuecomment-1038293127).)

Should this be mentioned here to avoid confusion?

9names: That's a good idea.
We shouldn't go into too much detail though, don't want to overwhelm beginners with too much info.

rp2040-hal/examples/gpio_irq_example.rs
```diff
@@ -0,0 +1,187 @@
...
+        // keep this example simple we'll just execute a `nop`.
+        // the `nop` (No Operation) instruction does nothing,
+        // but if we have no code here clippy would complain.
+        cortex_m::asm::nop();
+    }
+}
+
+#[interrupt]
+fn IO_IRQ_BANK0() {
+    static mut LED_AND_BUTTON: Option<LedAndButton> = None;
```
jannic, line 160:
```suggestion
    // The `#[interrupt]` attribute covertly converts this to `&'static mut Option<LedAndButton>`
    static mut LED_AND_BUTTON: Option<LedAndButton> = None;
```

rp2040-hal/examples/gpio_irq_example.rs
9names: I've added @jannic's suggested comment to the code to resolve this

thejpster (approved):

PR #298, opened 2022-02-22
Implement embedded-hal 1.0.0-alpha.7 traits

jannic:
- Several traits were removed from embedded-hal (https://github.com/rust-embedded/embedded-hal/pull/324/, https://github.com/rust-embedded/embedded-hal/pull/354), so this pull request also removes them.
- Some new functions were added to the I2C traits. Those have not been tested, yet.

### DIFF
--- a/rp2040-hal/CHANGELOG.md
+++ b/rp2040-hal/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- None
+- Updated embedded-hal alpha support to version 1.0.0-alpha.7
 
 ## [0.3.0] - 2021-12-19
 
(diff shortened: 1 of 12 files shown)


9names (approved):
Looks good to me.
Did you manage to find any driver crates that use e-h 1.0 traits?
I spent time adapting some while working on bl602 since it (originally) only supported e-h 1.0 alpha4, but wrangling dependencies sucked.

Also, which of the traits still need testing?
It would be good to have an issue with them so we don't forget to verify everything is working.

jannic:
No I don't have any drivers using this API, and so all traits need testing.
To be honest, I don't think anybody will use these traits while e-h is still alpha. I just implemented them to try out if there are any issues with the API which make it difficult to work with.
Therefore, merging them is not important. It would be equally fine to just have them on a branch for now, if you are worried about merging untested code.

9names:
>  would be equally fine to just have them on a branch for now, if you are worried about merging untested code.
I was more curious than cautious.
No, let's not leave them in a branch. I'm happier with untested code in-tree, as that is way more likely to be noticed and fixed.
And there's not much value in having the alpha traits unless they are the latest version, so I appreciate the effort here.

PR #300, opened 2022-02-23
Add an rp2040 specific #[entry] macro.

### LINKED ISSUE #292: Spinlocks not cleared by reset

jannic:
This macro extends the one from cortex-m-rt by code to unlock all spinlocks on boot.

I wonder if this should be part of the HAL, or if it belongs to rp2040-pac.

Closes #292

### DIFF
--- a/rp2040-hal-macros/src/lib.rs
+++ b/rp2040-hal-macros/src/lib.rs
@@ -1,0 +1,59 @@
+extern crate proc_macro;
+
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use quote::quote;
+use syn::{parse, parse_macro_input, Item, ItemFn, Stmt};
+
+#[proc_macro_attribute]
+pub fn entry(args: TokenStream, input: TokenStream) -> TokenStream {
+    let mut f = parse_macro_input!(input as ItemFn);
+
+    if !args.is_empty() {
+        return parse::Error::new(Span::call_site(), "This attribute accepts no arguments")
+            .to_compile_error()
+            .into();
+    }
+
+    let clear_locks: TokenStream = quote!(unsafe {
+        const SIO_BASE: u32 = 0xd0000000;
+        const SPINLOCK0: u32 = SIO_BASE + 0x100;
+        const SPINLOCK_COUNT: u32 = 32;
+        for i in 0..SPINLOCK_COUNT {
+            ::core::ptr::write_volatile((SPINLOCK0 + 4 * i) as *mut u32, 1);
+        }
+    })
+    .into();
+    let clear_locks = parse_macro_input!(clear_locks as Stmt);
+
+    // statics must stay first so cortex_m_rt::entry still finds them
+    let stmts = insert_after_static(f.block.stmts, clear_locks);
+    f.block.stmts = stmts;
+
+    quote!(
+        #[::cortex_m_rt::entry]
+        #f
+    )
+    .into()
+}
+
+/// Insert new statements after initial block of statics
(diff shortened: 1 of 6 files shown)


Sizurka:
One thing that might be nice is to be able to call the spinlock init (and any future required init if this gets extended) from outside a macro context.  Mostly I'm looking at RTIC, which [does it own codegen](https://github.com/rtic-rs/cortex-m-rtic/blob/57da1e0403510cafbdcf88e402b39ae6d3bf323e/macros/src/codegen.rs#L64) for `main`, so it'd be nice to have an easy way to call whatever the HAL needs to init easily from there.

That might be out of scope for this, though, since it's probably safe to assume that if you're not using cortex-m-rt you can figure out what needs to be done.

thejpster (approved):
LGTM!

rp2040-hal-macros/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+            .into();
+    }
+
+    let clear_locks: TokenStream = quote!(unsafe {
+        const SIO_BASE: u32 = 0xd0000000;
+        const SPINLOCK0: u32 = SIO_BASE + 0x100;
+        const SPINLOCK_COUNT: u32 = 32;
+        for i in 0..SPINLOCK_COUNT {
+            ::core::ptr::write_volatile((SPINLOCK0 + 4 * i) as *mut u32, 1);
+        }
```
9names, line 24: I think it would be more idiomatic to make a pointer to the first spinlock, use pointer arithmetic to calculate the offset to the current spinlock, and use write_volatile as an associated function for pointer (instead of passing it in as an argument)

```suggestion
        const SIO_BASE: u32 = 0xd0000000;
        const SPINLOCK0_PTR: *mut u32 = (SIO_BASE + 0x100) as *mut u32;
        const SPINLOCK_COUNT: usize = 32;
        for i in 0..SPINLOCK_COUNT {
            SPINLOCK0_PTR.wrapping_add(i).write_volatile(1);
        }
```

9names (approved):
> I wonder if this should be part of the HAL, or if it belongs to rp2040-pac.

This is firmly a HAL concern from my point of view. 
Is this still in draft due to that question?

Looks good to me too. The nitpick around raw pointer handling isn't a blocker for me.

jannic:
> Is this still in draft due to that question?

Not only because of that question. I just wasn't sure if this was a good approach in general, and wanted to wait for some feedback.

Now that there have been a few reviews, I'm confident that I didn't propose something stupid.

I rebased the PR because there were conflicts with changes on `main`.

9names (approved):
LGTM

PR #301, opened 2022-02-23
Allow to start multiple state machines in sync

### LINKED ISSUE #286: cannot synchronize PIOs with pio::StateMachine::synchronize_with()
Hi, I'm having trouble synchronizing PIOs with the current implementation of `pio::StateMachine::synchronize_with()`. Bounds require this to be run on `Stopped` state machines but I am not able to start them synchronously and thus they run a little out of sync. I can get them sychronized if I relax the bounds requirements and create a `pio::Synchronize::start` method which starts all state machines simultaneously. I have done so in the branch https://github.com/astraw/rp-hal/tree/hack-synchronize-with (which also includes an example program `pio_synchronized.rs` showing this functioning). However, this particular implementation is troublesome in that it starts the state machines without changing their type to `Running` from `Stopped`. I wonder if I am going about this wrong or if there is indeed not quite current API support for synchronizing PIO programs.

jannic:
This adds a bunch of boiler-plate structs to pio.rs, but is the easiest way I found to solve #286. Any idea for a better implementation is welcome.

The resulting API is fine though, as those types do not become visible:
```
    // Start both SMs at the same time
    let group = sm0.with(sm1).sync().start();

    // Stop both SMs at the same time
    let group = group.stop();

    // Start them again and extract the individual state machines
    let (sm1, sm2) = group.start().free();

    // Stop the two state machines separately
    let _sm1 = sm1.stop();
    let _sm2 = sm2.stop();
```
(Context: https://github.com/jannic/rp-hal/blob/9b625a573e0ad05c699845239ab3fdb2cae75fd1/rp2040-hal/examples/pio_synchronized.rs#L78-L93)

Closes #286

### DIFF
--- a/rp2040-hal/examples/pio_synchronized.rs
+++ b/rp2040-hal/examples/pio_synchronized.rs
@@ -39,10 +39,8 @@ fn main() -> ! {
     let pin1 = 1;
 
     // Define some simple PIO program.
-    let program = pio_proc::pio!(
-        32,
+    let program = pio_proc::pio_asm!(
         "
-    irq wait 0
 .wrap_target
     set pins, 1 [31]
     set pins, 0 [31]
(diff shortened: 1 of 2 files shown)


9names (approved):
Typestate always ends up this way. With only 4 options I think the tradeoff is worth it,
this is still less complex than an implementation based around Options.

Looks good to me.

9names (approved):

PR #303, opened 2022-02-26
Added so stopped PIO state machines can change their clock divider

korken89:


9names (approved):
Yep, that makes sense.
Thanks for the PR!

PR #305, opened 2022-02-27
Add Tiny2040 board support

MichaelBell:
I've added a BSP for the Pimoroni Tiny 2040, and a simple example blinking each colour of LED in turn.

I'm still quite new to Rust, so please check I haven't done anything silly - but this was mostly copy-paste so hopefully it's OK!


9names (commented):
This looks pretty good to me, nicely done.
I don't think you need most of these dev-dependencies though - maybe just embedded-hal, defmt and defmt-rtt?

MichaelBell:
Thanks - I had just taken the dependencies from the rp-pico, but many of those crates were only needed for its examples.  I have removed what I believe to be the unnecessary ones.

9names (approved):
Looks good to me.
Thanks for the PR!

PR #311, opened 2022-03-05
Remove the `alloc` requirement for `Core::spawn`

Liamolucko:
I've implemented `Core::spawn` for any `FnOnce` without a `Box`, ~~by making an extension to `FnOnce` which allows calling through a reference~~.
I modified the example slightly to demonstrate capturing the system tick frequency rather sending it through the FIFO.
I'm not at all an expert here, but it seems to work fine for the example.

Edit: I'm no longer using the hacky `FnOnce` extension; instead, I've made the core 1 startup function generic over the function type, which allows it to be copied onto the stack and then called.


Liamolucko:
I've started working on an example (https://gist.github.com/Liamolucko/bfbfbf863ceb72d4a7a9c876d8125c7f); however, I've run into a bit of an issue.
Basically, the example blinks one LED from each core, and I was trying to initialise them both before starting core 1, and then move one to core 1 by capturing it; the issue is, `Multicore::new` takes the whole of SIO, resulting in a compiler error because I've moved out the GPIO pins.
It only actually needs access to the FIFO; should I change the signature to take that instead?

Also, it might be a decent idea to move the stack parameter of `spawn` before the closure, since that'll make writing the closure inline cleaner.

9names:
Yeah, the existing peripheral requirements for a few of the HAL functions are problematic.
I've previously worked around it but if you see a way to split it up properly that would be much better.

Liamolucko:
Okay, I've implemented a simple example which blinks one LED on each core.

API changes to accommodate this:
- Make `Multicore::new` take `&mut SioFifo` instead of `&mut Sio`
- Flip the arguments to `Core::spawn`, so that the closure is the last argument. This makes multi-line closures look less ugly.

rp2040-hal/src/multicore.rs
```diff
@@ -165,10 +176,12 @@ impl<'p> Core<'p> {
...
 
-            push(wrapper as usize);
+            // We don't want to drop this, since it's getting moved to the other core.
+            let mut entry = ManuallyDrop::new(entry);
+
+            // Push the arguments to `core1_startup` onto the stack.
             push(stack.as_mut_ptr() as usize);
-            push(entry[1]);
-            push(entry[0]);
+            push(&mut entry as *mut _ as usize);
```
jannic, line 193: Isn't this a pointer to the stack of core0, which gets invalid as soon as `spawn` returns? But may still be accessed from `core1_startup`, later?
Even if `entry` is `ManuallyDrop` and so won't get dropped at the end of the method, its memory will still be invalidated.

Liamolucko: Yes, but core 0 waits for core 1 to signal that it's read from the pointer before returning: https://github.com/rp-rs/rp-hal/blob/0e5fdcfd9cbdd0efbf4af1315366a5b4c2e17b10/rp2040-hal/src/multicore.rs#L235-L236

9names: How would core1_startup access this after `spawn` returns?
We spin waiting for a signal from the other core before returning from `spawn`.
Entry is the entrypoint for core1, it can't send a reply until it has read it and started executing, and it would never be read again.

9names: Sorry, my reply was to jannic - apologies if it is confusing after Liam's comment

jannic: You're right: I missed the fact that `let entry = unsafe { ManuallyDrop::take(entry) };` is the point where the pointer gets dereferenced, and the contents get copied to a new location on core1's stack.
(ie. the argument to `ManuallyDrop::take` is a reference, but the new `let entry` is no longer a reference but an owned type.)

rp2040-hal/src/multicore.rs
```diff
@@ -140,13 +140,36 @@ impl<'p> Core<'p> {
...
+        F: FnOnce() -> bad::Never + Send + 'static,
+    {
+        if let Some((psm, ppb, fifo)) = self.inner.as_mut() {
+            // The first two ignored `u64` parameters are there to take up all of the registers,
+            // which means that the rest of the arguments are taken from the stack,
+            // where we're able to put them from core 0.
+            extern "C" fn core1_startup<F: FnOnce() -> bad::Never>(
+                _: u64,
+                _: u64,
+                entry: &mut ManuallyDrop<F>,
```
jannic, line 155: I wondered if it was ok to have `&mut` argument, but write `*mut` pointer to the stack. Even if it's an obvious choice to have the same memory representation for `*` and `&`, it's not necessarily guaranteed.

After doing some research I came to the conclusion that this is actually fine: https://rust-lang.github.io/unsafe-code-guidelines/layout/pointers.html says: "The layouts of &T, &mut T, *const T and *mut T are the same."

Is this common knowledge (of people reading low-level code), or should it be explained in a comment?

rp2040-hal/src/multicore.rs
```diff
@@ -140,13 +140,36 @@ impl<'p> Core<'p> {
...
-        if let Some((psm, ppb, sio)) = self.inner.as_mut() {
+    /// Spawn a function on this core.
+    pub fn spawn<F>(&mut self, stack: &'static mut [usize], entry: F) -> Result<(), Error>
+    where
+        F: FnOnce() -> bad::Never + Send + 'static,
+    {
+        if let Some((psm, ppb, fifo)) = self.inner.as_mut() {
+            // The first two ignored `u64` parameters are there to take up all of the registers,
+            // which means that the rest of the arguments are taken from the stack,
+            // where we're able to put them from core 0.
```
jannic, line 151: Still a hack, but better than the MULTICORE_TRAMPOLINE used before. :+1:

jannic (approved):
I don't have enough experience with multicore code to assess the usability of the API. But I tried and failed to find any unsoundness introduced by the changes. LGTM.

rp2040-hal/src/multicore.rs
9names: Agreed! My first impression was that it was too clever, but I've warmed up to it.

9names (approved):
Looks good to me.

rp2040-hal/src/multicore.rs
Liamolucko: I'm actually a little more concerned about how it's casted to a `usize` before writing and read back as a pointer - that seems like it's _probably_ equivalent to a `mem::transmute`, and under [strict provenance](https://doc.rust-lang.org/nightly/std/ptr/index.html#strict-provenance), [apparently that produces an invalid pointer](https://www.reddit.com/r/rust/comments/u1bbqn/pointers_are_complicated_iii_or_pointerinteger/iaarpcg/).

Liamolucko: To be safe, I've changed the stack setup to write the pointers directly rather than casting to `usize`.

9names: No, casting is absolutely *not* equivalent to transmute. The reddit comment by Ralf Jung you linked specifically states that pointer-int casts and int-pointer casts are something else entirely, and the rest of the thread discusses this several times.
Using the correct types for these doesn't cost us anything (and it's better to preserve semantic information than to discard it), but also keep in mind we're spending time (both changing the code and discussing it) to conform to memory semantics Rust does not now and might never have.

Liamolucko: > No, casting is absolutely *not* equivalent to transmute.

Sorry, I was unclear - I know that `as` casts aren't the same as transmutes, I meant that it seemed like a transmute for the function to reinterpret the `usize` written to the stack as an `&mut ...`.

> but also keep in mind we're spending time (both changing the code and discussing it) to conform to memory semantics Rust does not now and might never have.

Fair enough.

rp2040-hal/src/multicore.rs
```diff
@@ -157,14 +180,19 @@ impl<'p> Core<'p> {
...
-            };
+                stack_ptr.cast::<*mut usize>().write(stack.as_mut_ptr());
 
-            push(wrapper as usize);
-            push(stack.as_mut_ptr() as usize);
-            push(entry as usize);
+                // Push `entry`.
+                stack_ptr = stack_ptr.sub(1);
+                stack_ptr.cast::<&mut ManuallyDrop<F>>().write(&mut entry);
+            }
```
jannic, line 195: To be pedantic: Do we need a memory barrier at this point, to make sure the new stack values are actually written before core1 tries to load them?

9names: If we're being pedantic, yes.
A compiler fence is probaby enough but we might as well do it properly.

jannic:
```suggestion
            }
            // Make sure the compiler does not reorder the stack writes
            // in relation to the fifo writes, below.
            //
            // From the compiler perspective, this doesn't guarantee that the
            // second core actually sees those writes. However, we know that
            // the RP2040 doesn't have memory caches, and writes happen in-order.
            compiler_fence(Ordering::AcqRel);

```
(And add `use core::sync::atomic::{compiler_fence, Ordering};` above).

I did not use `core::sync::atomic::fence` because that doesn't provide additional guarantees without an additional atomic variable shared between the cores, as explained in https://doc.rust-lang.org/core/sync/atomic/fn.fence.html.

Liamolucko: Why does this need `Ordering::AcqRel`? To me, it seems like `Release` should be enough.

jannic: Good question. You are right, `Release` should be enough.
My reasoning was that I also want to prevent later accesses from being reordered to before this fence. But that's actually already guaranteed by `Release`: "with Release, preceding reads and writes cannot be moved past subsequent writes"  implies that subsequent writes can't be moved to before preceding writes, otherwise it doesn't make any sense.

```suggestion
            }
            // Make sure the compiler does not reorder the stack writes
            // in relation to the fifo writes, below.
            //
            // From the compiler perspective, this doesn't guarantee that the
            // second core actually sees those writes. However, we know that
            // the RP2040 doesn't have memory caches, and writes happen in-order.
            compiler_fence(Ordering::Release);

```

9names:
Thanks for the awesome PR @Liamolucko, going to merge this one now

PR #312, opened 2022-03-09
Prep for 0.4.0 release

9names:
Some last-minute tidy ups for the 0.4.0 release.

- Updated to pio 0.2.0 and pio-proc 0.2.0, and migrated the HAL examples to match
- IRQ example wasn't building since it required HAL to be build with features = ["rt"] but that wasn't specified
- Updated the changelog

### DIFF
--- a/boards/adafruit-feather-rp2040/Cargo.toml
+++ b/boards/adafruit-feather-rp2040/Cargo.toml
@@ -22,8 +22,7 @@ panic-halt= "0.2.0"
 embedded-hal ="0.2.5"
 nb = "1.0.0"
 smart-leds = "0.3.0"
-pio = "0.1.0"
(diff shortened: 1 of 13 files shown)


jannic:
Some checks failed because ws2812-pio-rs still depends on pio-0.1.0.
Should be easy to fix with https://github.com/ithinuel/ws2812-pio-rs/pull/10

jannic:
Same with i2c-pio-rs, fix available as https://github.com/ithinuel/i2c-pio-rs/pull/4
@9names @ithinuel can you coordinate merging of those changes?
As boards/rp-pico/Cargo.toml and several other board support crates point to git revisions of the affected crates, I think @ithinuel needs to merge first, so the git dependencies in rp-hal can be updated accordingly.

jannic:
One more detail breaking the CI checks fixed: https://github.com/9names/rp-hal/pull/1

9names:
Thanks for the review + help @jannic, and for the merges @ithinuel.
Going to merge this now and get this release out :tada:

PR #313, opened 2022-03-11
Update BSPs prior to release

9names:
Now that HAL 0.4.0 is out we can think about publishing updates for all the BSPs.
I thought it would be a good idea to try to get them a little bit more consistent first.

There's also an update for the HAL readme that I missed before pressing the publish button.
I won't push a new version of the HAL until we accumulate a bunch of bug-fixes

### DIFF
--- a/boards/adafruit-feather-rp2040/CHANGELOG.md
+++ b/boards/adafruit-feather-rp2040/CHANGELOG.md
@@ -15,9 +15,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - None
 
+## [0.2.0] - 2022-03-11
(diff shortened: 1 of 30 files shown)


thejpster (approved):

thejpster:
The HAL README is unfortunate as the old one is the one you see on crates.io. But it's probably not worth a minor bump just for that.

9names:
Yeah. It's a sad case of "things you notice as soon as you hit the publish button".
I had considered a patch version bump and release, but I assume we'll have sufficient cause for a new release soon enough.
The right sidebar on crates.io has the correct value, and the version displayed on github will be corrected, that's probably enough for now.

Issue #284, opened 2022-01-30
`uart::Writer` isn't `Send`, but I think it could be


danielzfranklin:
This is based on a long conversation I had with adamgrieg on matrix, which didn't end with a definite conclusion. I'm not confident, but I think that `uart::Write` not being `Send` is a coincidence and not an actual requirement.

`UartPeripheral::split` gives the reader half `self.device`, and clones a new device for the writer.

```rust
    pub fn split(self) -> (Reader<UART0, P>, Writer<UART0, P>) {
        let reader = Reader {
            device: self.device,
            pins: self.pins,
            config: self.config,
            effective_baudrate: self.effective_baudrate,
        };
        // Safety: reader and writer will never write to the same address
        let device_copy = unsafe { &*UART0::ptr() };
        let writer = Writer {
            device: device_copy,
            device_marker: core::marker::PhantomData,
            pins: core::marker::PhantomData,
        };
       (reader, writer)
    }
```

`self.device` is the pac token that (in my limited understanding) proves "moral ownership", which the pac asserts is `Send`. In contrast, `device_copy` is the raw pointer to the `RegisterBlock`, which isn't `Send`.

```rust
pub struct UART0 {
    _marker: PhantomData<*const ()>,
}
unsafe impl Send for UART0 {}
impl UART0 {
    #[doc = r"Pointer to the register block"]
    pub const PTR: *const uart0::RegisterBlock = 0x4003_4000 as *const _;
    #[doc = r"Return the pointer to the register block"]
    #[inline(always)]
    pub const fn ptr() -> *const uart0::RegisterBlock {
        Self::PTR
    }
}
```

As a result the `Reader` is `Send` but the `Writer` isn't. I think this is just a coincidence of the implementation details. I don't know enough about how the registers work to know if the current implementation is sound. I do think this is equally sound:

```diff
     pub fn split(self) -> (Reader<UART0, P>, Writer<UART0, P>) {
         let reader = Reader {
             device: self.device,
             pins: self.pins,
             config: self.config,
             effective_baudrate: self.effective_baudrate,
         };
         // Safety: reader and writer will never write to the same address
-       let device_copy = unsafe { &*UART0::ptr() };
+       let device_copy = unsafe { Peripherals::steal().UART0 };
         let writer = Writer {
             device: device_copy,
             device_marker: core::marker::PhantomData,
             pins: core::marker::PhantomData,
         };
         (reader, writer)
     }
```

[...]

PR #314, opened 2022-03-11
Using thread send safe UART* marker, when splitting

marius-meissner:
For details, see #284.

The approach was initially suggested by @danielzfranklin. I just filed the PR and did some hardware testing.
Changes were successfully tested by using UART for ESP8266 Wi-Fi offloading in combination with RTIC.

There are perhaps better ways to solve this problem, so I'm looking forward to your feedback.

### DIFF
--- a/rp2040-hal/src/uart/peripheral.rs
+++ b/rp2040-hal/src/uart/peripheral.rs
@@ -16,6 +16,7 @@ use rp2040_pac::{UART0, UART1};
 
 #[cfg(feature = "eh1_0_alpha")]
 use eh1_0_alpha::serial as eh1;
+use pac::Peripherals;
 
 /// An UART Peripheral based on an underlying UART device.
 pub struct UartPeripheral<S: State, D: UartDevice, P: ValidUartPinout<D>> {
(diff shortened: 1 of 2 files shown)


9names:
Looks okay to me.
I'll point out (in case you are unaware) that ::steal() has side effects. Shouldn't be a problem here, as you will have already called ::take() before this.
https://github.com/rust-embedded/cortex-m/issues/186

Another option is to create an new token by transmuting ()
```rust
let uart_token: UART0 = unsafe { core::mem::transmute( () ) };
```

9names (approved):
LGTM

marius-meissner:
Thanks for the quick approval and the feedback.
I had unfortunately forgotten to reply to the previous comment.

> Another option is to create an new token by transmuting ()

Right. But I think in this case the use of take slightly outweighs, especially because of the following indications:

> transmute is incredibly unsafe. There are a vast number of ways to cause [undefined behavior](https://doc.rust-lang.org/stable/reference/behavior-considered-undefined.html) with this function. transmute should be the absolute last resort.

danielzfranklin:
I mostly agree.

There's nothing about transmuting `()` to an arbitrary zero-sized type that's inherently unsound, the question is only whether forging a type a library assumes only it can create lets you "trick" code that relies on something for safety[1].

Either way we're "tricking" the pac in a way we know is sound based on internal implementation details. The `steal` call is extremely cheap though, and I feel like it reads cleaner.

[1]: https://internals.rust-lang.org/t/is-synthesizing-zero-sized-values-safe/11506/5

PR #315, opened 2022-03-12
Remove some unused fields from UartPeripheral and Reader

### LINKED ISSUE #310: `UartPeripheral` stores its config and baudrate for no reason
`UartPeripheral::config` and `UartPeripheral::effective_baudrate` are never actually used; the only time they're read is to pass them to `uart::Reader`, which never uses them either. These should be removed, since all they do is take up space (I'm pretty sure that `UartPeripheral` would be a ZST without them, too).

jannic:
As @Liamolucko noticed in #310, the UART code stores some config values in its structs which are never used.

While it's possible that future extensions actually need those fields, I don't think it's reasonable to keep them just in case, without a near-term plan to actually use them.

Closes #310


9names (approved):
LGTM

PR #317, opened 2022-03-13
Constrain pio::fifo::write value to valid types

### LINKED ISSUE #316: `pio::Tx::write` allows writes of arbitrary types, including types wider than a FIFO entry, to the FIFO
The type parameter `T` of [`pio::Tx::write`](https://github.com/rp-rs/rp-hal/blob/f8de8755ccf82a782c9430c343afa8daea671b24/rp2040-hal/src/pio.rs#L828) is unbounded and thus allows writing any type to the PIO SM's FIFO, even those wider than a FIFO entry.

This also leads to cases where passing a reference (e.g. `&u8`) to `write` fails to produce an error and writes the pointer to the FIFO - likely not the user's intention.

See also [discussion on Matrix](https://matrix.to/#/!zbjndUiYJDvhDRkPGw:matrix.org/$cWJ5Mb0NpwFhHqPpKpfcfnk7cRAbtzP422aLK762Yc0?via=matrix.org).

9names:
Use funty to limit the types of values that can be passed as a generic parameter to pio::fifo::write.
Previously it would allow many types that would almost certainly have been unsafe, plus it didn't catch when you accidentally passed in a ref or pointer.

Fixes #316

### DIFF
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -4,6 +4,7 @@ use crate::{
     atomic_register_access::{write_bitmask_clear, write_bitmask_set},
     resets::SubsystemReset,
 };
+use funty::{AtMost32, Integral};
 use pio::{Program, SideSet, Wrap};
 use rp2040_pac::{PIO0, PIO1};
 
@@ -825,7 +826,10 @@ impl<SM: ValidStateMachine> Tx<SM> {
     /// Write an element to TX FIFO.
     ///
     /// Returns `true` if the value was written to FIFO, `false` otherwise.
-    pub fn write<T>(&mut self, value: T) -> bool {
+    pub fn write<T>(&mut self, value: T) -> bool
+    where
+        T: Integral + AtMost32,
+    {
         // Safety: The register is never written by software.
         let is_full = self.is_full();
 
@@ -835,7 +839,7 @@ impl<SM: ValidStateMachine> Tx<SM> {
 
         unsafe {
             let reg_ptr = self.register_block().txf[SM::id()].as_ptr() as *mut T;
-            core::ptr::write_volatile(reg_ptr, value);
+            reg_ptr.write_volatile(value);
         }
 
         true
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -15,6 +15,7 @@ cortex-m-rt = ">=0.6.15,<0.8"
 embedded-hal = { version = "0.2.5", features = ["unproven"] }
 eh1_0_alpha = { version = "=1.0.0-alpha.7", package="embedded-hal", optional=true }
 embedded-time = "0.12.0"
+funty = { version = "2.0.0", default-features = false }
 itertools = { version = "0.10.1", default-features = false }
 nb = "1.0"
 rp2040-pac = "0.3.0"


rp2040-hal/src/pio.rs
```diff
@@ -825,7 +826,10 @@ impl<SM: ValidStateMachine> Tx<SM> {
     /// Write an element to TX FIFO.
     ///
     /// Returns `true` if the value was written to FIFO, `false` otherwise.
-    pub fn write<T>(&mut self, value: T) -> bool {
+    pub fn write<T>(&mut self, value: T) -> bool
+    where
+        T: Integral + AtMost32,
```
WaffleLapkin, line 831: I think this still allows writes of types with size > 4 bytes (32 bits) while being more restrictive than needed.

Both `funty::Integral` and `funty::AtMost32` are safe and non-sealed, which means that someone could implement them for an arbitrary big type like `struct Chunky([u8; 1024])` and then pass it to this function. I don't think that the pointer is valid for writes of arbitrary big types.

Also, `write_volatile` [safety] requires `dst` (`reg_ptr`) to be aligned, and again, since `Integral + AtMost32` can be implemented for arbitrary types, someone could pass a type with a bigger alignment that `u32` has.

[safety]: (https://doc.rust-lang.org/std/ptr/fn.write_volatile.html#safety)

This also makes it very tedious to pass custom structures like
```rust
#[repr(C)]
struct Example {
    a: u16,
    b: u8,
    c: u8,
}
```
even though this should be valid, afaiu.

I propose to change the bounds to `T: `[`Pod`](https://docs.rs/bytemuck/latest/bytemuck/trait.Pod.html) and adding the following assertions:
```rust
assert!(size_of::<T>() <= size_of::<u32>());
assert!(align_of::<T>() <= align_of::<u32>());
```
I'm not sure if ZST write would be valid here, so maybe this assertion is also needed:
```rust
assert!(0 < size_of::<T>());
```

Note that since `size_of` and `align_of` are const fns, these assertions will be optimized out for valid `T`s.

IMO my proposed solution is more robust and correct, but I may be wrong. I can open a PR with my proposed solution.

9names: Discussion in #rp-rs suggests we're going to go for having separate impls for u32, u16 and u8 instead to bypass this issue.
Thanks for writing up your thoughts on this, I'm sure I'll have to deal with this sort of problem again at some point, so talking about it is still valuable.

Context:
> thejpster: Why don't we just take a u32 and let the compiler expand U16 and u8 automatically?

> jannic: While this byte lane duplication can be a useful feature, it seems to be too much magic for a good rust API if it's just implicit. I'd prefer separate methods like write_u8_replicated(&mut self, value: u8), write_u16_replicated(&mut self, value: u16). Those should explain replication it their doc comments. (A seasoned embedded developer probably knows what it is, but for someone coming from a non-embedded context, it's very surprising.)
write(&mut self, value: T) could then be changed to write(&mut self, value: u32) and do a 32 bit register write, filling missing bits with zeroes when called with a smaller data type.

WaffleLapkin: BTW I think `read` can use `Pod` too, so I'd be `fn read<T: Pod>(&mut self) -> Option<T>` and this may be handy to use your own types with FIFO? Though I'm very unfamiliar with the pico, so maybe that's not that useful as I think, idk.

WaffleLapkin: When I was writing the last two comments I didn't see yours, weird ><

Yeah, having separate impls is a good choice. And I'm glad my thoughts were valuable C:

9names:
Converting to draft again for now

9names:
Closing this in favour of #318

PR #319, opened 2022-03-17
WIP: Multicore bootload

9names:
Existing Multicore code set up a trampoline function, then vectored into user code in the same executable from there.

This PR extracts the core functionality for launching code on another core from `inner_spawn` into `inner_bootstrap` (named after `cortex_m::asm::bootstrap` that does approximately the same work).
`inner_spawn` now sets up the stack and calculates the entry point for the trampoline before calling `inner_bootstrap`.

A new function `bootload` is used for starting a second program by using information from it's vector table - this one is named after [cortex_m::asm::bootload](https://github.com/rust-embedded/cortex-m/blob/e6c7249982841a8a39ada0bc80e6d0e492a560c3/src/asm.rs#L290)

There is currently no checking for alignment / other errors, will add those later.
I have briefly tested this code - it is mostly the same as some quick POCs i did to run code from [Flash](https://github.com/9names/rp2040_experiments-multicore_multiprogram/blob/main/core0-flash_core1-flash/src/main.rs) and [RAM](https://github.com/9names/rp2040_experiments-multicore_multiprogram/blob/main/core0-flash_core1-ram/src/main.rs) - I plan on updating this test code to validate this PR further

### DIFF
--- a/rp2040-hal/src/multicore.rs
+++ b/rp2040-hal/src/multicore.rs
@@ -108,14 +108,15 @@ fn core1_setup(stack_bottom: *mut usize) {
 /// Perform some basic validation of the program payload
 ///
 /// Validation performed:
-/// - Check `vector_table_addr` is correctly aligned. See
+/// - Check address `vector_table` points to is correctly aligned. See
 /// [CM0+ user guide](https://developer.arm.com/documentation/dui0662/b/The-Cortex-M0--Processor/Exception-model/Vector-table)
 /// for reference
 /// - Check `stack_addr` is correctly aligned
 fn validate_bootstrap_payload_alignment(
-    vector_table_addr: usize,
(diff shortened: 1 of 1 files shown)


9names:
Alignment + some quick checking of address ranges is now performed by validate_bootstrap_payload, which is invoked inside of bootload.
This code is tested to work with the examples in [my multicore example repo](https://github.com/9names/rp2040_experiments-multicore_multiprogram.git)
The existing multicore example in the HAL still works, too - in earlier versions of this PR I had accidentally transposed stack pointer and entry point arguments.

9names:
Marking this as draft until I can get it rebased on top of #311.

9names:
Dropping this PR as I currently have no projects that require it and no plans to pick it back up

PR #321, opened 2022-03-27
RAM-based interrupt vector tables

9names:
To improve performance for interrupts and allow for dynamically hooking interrupt vectors at runtime, provide some facilities for creating and using interrupt vector tables.
This should also increase what we can do with multicore.

This code is currently completely untested.


rp2040-hal/src/multicore.rs
```diff
@@ -107,6 +107,86 @@ impl<const SIZE: usize> Stack<SIZE> {
     }
 }
 
+/// Entry for a Vector in the Interrupt Vector Table.
+///
+/// Each entry in the Vector table is a union with usize to allow it to be 0 initialized via const initializer
+/// This union also allows for updating the stack pointer or
```
jannic, line 113: or?

rp2040-hal/src/multicore.rs
```diff
@@ -107,6 +107,86 @@ impl<const SIZE: usize> Stack<SIZE> {
...
+    }
+
+    /// Initialise our vector table by copying the current table on top of it
+    pub fn init(&mut self, ppb: &mut pac::PPB) {
+        let vector_table = ppb.vtor.read().bits();
+        unsafe {
+            crate::rom_data::memcpy44(
+                &mut self.table as *mut _ as *mut u32,
+                vector_table as *const u32,
+                192,
```
jannic, line 153: Something like `size_of::<VectorTable>()` would avoid hardcoding the size.

jannic:
Didn't test the code either, but it looks sensible.
Maybe find a better place, though? I don't think it's multicore-related.

rp2040-hal/src/multicore.rs
9names: Gosh, I have no idea what I was thinking at this point.
I'll remove that line, hopefully the train of thought returns when I get a chance to make progress on this code next.

rp2040-hal/src/multicore.rs
9names: Good idea, thanks!

jannic:
> I still don't have a good mental model of whether the padding bytes (256 bytes - 192) will be used by other static variables, so avoiding writing to them seems the safest option.

This seems to be an open question:
- https://github.com/rust-lang/rfcs/issues/1397
- https://github.com/rust-lang/unsafe-code-guidelines/issues/176

9names:
Thanks for those threads. I think the answer (right now) is that the padding bytes are unused.
And, it doesn't seem like any existing target will support stride != size, so that should hold in the future.
Would be nice to have a way to overalign the start of an array to avoid this though - maybe we'll get magic syntax for that some day but I don't want it so badly that I'll push for it.

rp2040-hal/src/vector_table.rs
```diff
@@ -0,0 +1,84 @@
...
+//! Vector Table Offset Register (VTOR) of the Cortex-M0+
+//! Also provides types and utilities for copying a vector table into RAM
+
+/// Entry for a Vector in the Interrupt Vector Table.
+///
+/// Each entry in the Vector table is a union with usize to allow it to be 0 initialized via const initializer
+///
+/// Implementation borrowed from https://docs.rs/cortex-m-rt/0.7.1/cortex_m_rt/index.html#__interrupts
+#[derive(Clone, Copy)]
+pub union Vector {
```
jannic, line 13: Does this need to be `pub`?

9names: No, this was me being lazy when copying the code.
Will remove pub, user doesn't need to see this.

jannic (approved):

PR #325, opened 2022-04-10
Provide an unsafe function for resetting all spinlocks

9names:
RTIC-based examples can't make use of the hal::entry macro to reset all spinlocks.
Provide an unsafe fn() that can be used to reset all spinlocks in init()

### DIFF
--- a/boards/rp-pico/examples/pico_rtic.rs
+++ b/boards/rp-pico/examples/pico_rtic.rs
@@ -27,6 +27,11 @@ mod app {
 
     #[init]
     fn init(c: init::Context) -> (Shared, Local, init::Monotonics) {
+        // Soft-reset does not release the hardware spinlocks
+        // Release them now to avoid a deadlock after debug or watchdog reset
+        unsafe {
+            hal::sio::spinlock_reset();
+        }
         let mut resets = c.device.RESETS;
         let mut watchdog = Watchdog::new(c.device.WATCHDOG);
         let _clocks = init_clocks_and_plls(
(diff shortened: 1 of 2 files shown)


ithinuel (changes requested):
I believe the `hal::entry` macro should be updated to make use of this function to avoid code repetition.

jannic:
IIRC this was something I tried an failed to do when writing the macro, because it's not easy to refer to a name from the hal from the macro create without causing a circular dependency. And the hal could be renamed when importing it, with no easy way for the macro crate to know that name. (But I know next to nothing about rust macros, so it may well be that there is a good solution which I just didn't find)

ithinuel:
I see, thank you for the explanation.
I had a look at what the `futures` crate does and it indeed seems filled with some magic I need to get my head around first. ([1](https://github.com/rust-lang/futures-rs/blob/9f73b8dceaebd86048995a674fc9824b3df8a6bc/futures-util/src/lib.rs#L45-L58) [2](https://github.com/rust-lang/futures-rs/blob/9f73b8dceaebd86048995a674fc9824b3df8a6bc/futures-util/src/async_await/join_mod.rs#L94-L97) [3](https://github.com/rust-lang/futures-rs/blob/9f73b8dceaebd86048995a674fc9824b3df8a6bc/futures-macro/src/join.rs#L125-L142))
It might not be worth doing this here.

ithinuel (approved):

jannic:
Thanks for the pointer! That indirection allows access to the $crate feature from a proc macro, quite clever. And no, probably not worth using that approach in the entry macro, as the added code for the workaround would be much longer (and complicated) than the code duplicated in the current PR.

Issue #322, opened 2022-04-02
Add other ADC modes: free-running, FIFO


9names:
The current ADC implementation in the HAL only provides support for blocking one-shot samples.

Tasks:
- Provide an API to configure the ADC into free-running mode.
- Provide an API to configure the ADC sample FIFO, pacing timer, round-robin sampling and right shift options.
- Add an example showing off each mode added

9names:
[C SDK reference for configuring the ADC](https://github.com/raspberrypi/pico-sdk/blob/2062372d203b372849d573f252cf7c6dc2800c0a/src/rp2_common/hardware_adc/include/hardware/adc.h#L178)
[C example using polling in continuous mode](https://github.com/raspberrypi/pico-examples/blob/2fb08a028f886624ebc9c938775e054860549159/adc/adc_console/adc_console.c#L17)

9names:
There is no example in the C SDK for capturing multiple channels or handling the samples in the interrupt handler, though they do have one that [uses DMA to capture the data](https://github.com/raspberrypi/pico-examples/blob/2fb08a028f886624ebc9c938775e054860549159/adc/dma_capture/dma_capture.c#L74)

riskable:
Free-running mode support would be *fantastic*!  Very much looking forward to this feature.

For reference, "the plan" is to hook up an analog multiplexer to a single analog pin on the RP2040 and then hook up additional multiplexers to each channel of the one that's connected to the RP2040 (i.e. multiplexers through a multiplexer).  In free-running mode (without round-robin) this should--in theory--allow one to effectively bypass the RP2040's internal ADC MUX and replace it with an external one with the fastest performance possible...  No need to issue blocking calls that tell the RP2040 which ADC pin to read from and enable non-blocking reads (well, reads that don't require 96 cycles).

Hopefully that'll be possible :smile:

9names:
Resolved by #626 and #636

PR #326, opened 2022-04-12
WIP: ADC FIFO

9names:
Trying to flesh out what ADC FIFO/free running IRQ functionality will look like.
Working towards #322

### DIFF
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -80,3 +80,8 @@ required-features = ["rt", "critical-section-impl"]
 # this example uses cortex-m-rt::interrupt, need rt feature for that
 name = "adc_fifo"
 required-features = ["rt", "critical-section-impl"]
+
+[[example]]
+# this example uses cortex-m-rt::interrupt, need rt feature for that
+name = "adc_fifo_dma"
+required-features = ["rt", "critical-section-impl"]
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -21,6 +21,7 @@ embedded-dma = "0.2.0"
 fugit = "0.3.6"
 itertools = { version = "0.10.1", default-features = false }
 nb = "1.0"
+num-traits = { version = "0.2.0", default-features = false }
 rp2040-pac = "0.4.0"
 paste = "1.0"
 pio = "0.2.0"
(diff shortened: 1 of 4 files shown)


9names:
ADC FIFO is now returning correct results for my test program (which reads 4 analog channels).
Back-to-back transfers generates data faster than I can consume it, so it's now set to 1000hz like the example in the datasheet.

9names:
Changed API so you don't have to calculate the divisor yourself, so instead of
```
// capture channel 1000 times/sec
adc.start_many_round_robin(0b1, 47999, 0);
```
you have 
```
adc.start_many_round_robin(0b1, 1000);
```

Also, the calculated divisor also factors in the number of channels - the previous example would have required
```
// capture both channels 1000 times/sec
adc.start_many_round_robin(0b11, 23999, 0);
```
to capture 2 channels at the same rate, now we can use
```
adc.start_many_round_robin(0b11, 1000);
```

badrbouslikhin:
I would benefit from this too.
I skimmed through your commits, can I help in any way?

9names:
> I would benefit from this too. I skimmed through your commits, can I help in any way?

Sure! This is functional now, but I'm not happy with the API and I'm not sure when I'll get back to it - so if you want to take over this PR please do.

nilclass:
@9names this is a great start! I took some inspiration from it and tried to come up with a better API, see #626.

To paraphrase your example:
```
adc.start_many_round_robin(0b11, 23999, 0);
```
would become:
```
let fifo = adc.free_running()
  .clock_divisor(23999, 0)
  .round_robin((&mut adc_pin0, &mut adc_pin1))
  .start_fifo();

// do something with the `fifo` object...
```

The `round_robin` method takes a tuple of channels instead of raw binary values, to avoid mistakes (similar to how `OneShot.read` takes a channel object instead of a raw number).

In addition to the `clock_divisor` method a `sample_freq` method could be added which computes the `int` and `frac` values (like you did in this PR).
I'm thinking such a method would need some bound checks though. From what I understand the minimum sample rate is something like 723 Hz (`48 MHz / (1 + 2^16 - 1 + (2^8 - 1 ) / 256)`), when the dividers are set to their maximum.

9names:
Closing this PR as #626 and #636 are better solutions

PR #328, opened 2022-04-17
Add defmt feature (as optionnal dep) and allow i2c::Error to be formatted

ithinuel:


rp2040-hal/Cargo.toml
```diff
@@ -30,6 +30,8 @@ critical-section = { version = "0.2.4", features = ["custom-impl"] }
 futures = { version = "0.3", default-features = false, optional = true }
 chrono = { version = "0.4", default-features = false, optional = true }
 
+defmt = { version = "0.2.0", optional = true }
```
9names, line 33: Could you specify the version as a range like we do with [embedded-hal-rt](https://github.com/rp-rs/rp-hal/blob/f67b650bb38044ff009c7bbabc510de90eab4910/rp2040-hal/Cargo.toml#L14)?
Can't use this as-is if using defmt 0.3 (that's what the template is on), and I think most users should also have moved over by now since newer probe-run won't even work with 0.2
```suggestion
defmt = { version = ">=0.2.0, <0.4", optional = true }
```

9names (approved):
LGTM

PR #329, opened 2022-04-17
Remove unmaintained implementation async i2c.

ithinuel:
I made an alternative prototype using `embedded-hal-async` here: https://github.com/ithinuel/rp2040-async-i2c/


9names (approved):
LGTM

PR #332, opened 2022-04-28
Add adafruit-macropad_blinky example

### LINKED ISSUE #302: Adafruit macropad examples directory has gone missing
I think the 'exercises' directory might have been accidentally removed: https://github.com/rp-rs/rp-hal/blob/main/boards/adafruit-macropad/examples gives a 404 error.

jannic:
Also adds const XOSC_CRYSTAL_FREQ = 12_000_000, which is the right
value according to https://learn.adafruit.com/assets/103270

Fixes #302


9names (approved):
LGTM.
I've got one of these rp2040 macropads turning up in a couple of weeks, this will be handy.

PR #336, opened 2022-05-07
Initial BSP for SparkFun Thing Plus RP2040

TyPott:
Provides some rudimentary support for the SparkFun Thing Plus RP2040 board. For the most part, this just follows the example set by the BSP for the Pro Micro RP2040, just changing the aliases where appropriate. I've confirmed that the Rainbow example works on the board, too.

### DIFF
--- a/boards/sparkfun-thing-plus-rp2040/README.md
+++ b/boards/sparkfun-thing-plus-rp2040/README.md
@@ -1,0 +1,95 @@
+# [sparkfun-thing-plus-rp2040] - Board Support for the [Sparkfun Thing Plus RP2040]
+
+You should include this crate if you are writing code that you want to run on
+a [Sparkfun Thing Plus RP2040] - a smaller [RP2040][Raspberry Silicon RP2040]
+board with USB-C and a WS2812B addressable LED.
+
+This crate includes the [rp2040-hal], but also configures each pin of the
+RP2040 chip according to how it is connected up on the Pro Micro RP2040.
+
+[Sparkfun Thing Plus RP2040]: https://www.sparkfun.com/products/17745
+[sparkfun-thing-plus-rp2040]: https://github.com/rp-rs/rp-hal/tree/main/boards/sparkfun-thing-plus-rp2040
+[rp2040-hal]: https://github.com/rp-rs/rp-hal/tree/main/rp2040-hal
+[Raspberry Silicon RP2040]: https://www.raspberrypi.org/products/rp2040/
+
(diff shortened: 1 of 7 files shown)


boards/sparkfun-thing-plus-rp2040/README.md
```diff
@@ -0,0 +1,95 @@
...
+[sparkfun-thing-plus-rp2040]: https://github.com/rp-rs/rp-hal/tree/main/boards/sparkfun-thing-plus-rp2040
+[rp2040-hal]: https://github.com/rp-rs/rp-hal/tree/main/rp2040-hal
+[Raspberry Silicon RP2040]: https://www.raspberrypi.org/products/rp2040/
+
+## Using
+
+To use this crate, your `Cargo.toml` file should contain:
+
+```toml
+sparkfun-thing-plus-rp2040 = "0.2.0"
```
9names, line 20: Update the version number to 0.1.0 to match version in Cargo.toml

boards/sparkfun-thing-plus-rp2040/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+    Gpio7 { name: scl },
+    Gpio8 { name: ws2812 },
+    Gpio16 { name: gpio16 },
+    Gpio17 { name: gpio17 },
+    Gpio18 { name: gpio18 },
+    Gpio19 { name: gpio19 },
+    Gpio20 { name: gpio20 },
+    Gpio21 { name: gpio21 },
+    Gpio22 { name: gpio22 },
+    Gpio25 { name: led_builtin },
```
9names, line 35: Could you rename Gpio25 from `led_builtin` to `led`?
Where possible I'm trying to keep the [template blinky code](https://github.com/rp-rs/rp2040-project-template/blob/f4fd405c917ee4d077f70bff63c46b76f6db1838/src/main.rs#L57) portable across BSPs, which is easier if it has the same name.

9names (changes requested):

boards/sparkfun-thing-plus-rp2040/src/lib.rs
TyPott: Can do. I used `led_builtin` to match SparkFun's schematic, which I think they did to disambiguate this LED from the WS2812, but keeping the names consistent makes sense to me too.

9names (approved):

9names (approved):

PR #341, opened 2022-05-20
Add missing copi pin for sparkfun-pro-micro-2040

Nashenas88:
The sparkfun-pro-micro-2040 pins definition was missing the copi pin on gpio 23, see https://cdn.sparkfun.com/assets/e/2/7/6/b/ProMicroRP2040_Graphical_Datasheet.pdf for reference from sparkfun.


9names (approved):
Looks good to me, thanks!

PR #344, opened 2022-05-24
Add missing GPIO I2C trait implmentations

chmanie:
This PR adds the missing I2C trait implementations for `Gpio`s 22-25 and 28+29. This is in line with the datasheet (p14) which lists all GPIOs to be I2C-capable.


9names (approved):
LGTM

Issue #342, opened 2022-05-23
PWM with servo seems not working properly


lucazulian:
I'm trying to move a micro servo with Raspberry Pi Pico and a simple micro servo motor but with the following code, the servo is not precise as the experiments I made with an Arduino board. I think the main reason would be the frequency with which I move the servo but I cannot find a way to have a similar precision.

```rust
//! # PWM Gear Steering Servo
//!
//! PWM attached to GP1

#![no_std]
#![no_main]

use cortex_m::delay::Delay;
// The macro for our start-up function
use cortex_m_rt::entry;

// GPIO traits
use embedded_hal::PwmPin;

// Time handling traits
use embedded_time::rate::*;

// Ensure we halt the program on panic (if we don't mention this crate it won't
// be linked)
use panic_halt as _;

// Pull in any important traits
use rp_pico::hal::prelude::*;

// A shorter alias for the Peripheral Access Crate, which provides low-level
// register access
use rp_pico::hal::pac;

// A shorter alias for the Hardware Abstraction Layer, which provides
// higher-level drivers.
use rp_pico::hal;

// The minimum PWM value we want
const LOW: u16 = u16::MIN;

// The maximum PWM value we want
const HIGH: u16 = u16::MAX;

/// Entry point to our bare-metal application.
///
/// The `#[entry]` macro ensures the Cortex-M start-up code calls this function
/// as soon as all global variables are initialised.
///
/// The function configures the RP2040 peripherals, then moves the servo in an
/// infinite loop.
#[entry]
fn main() -> ! {
    // Grab our singleton objects
    let mut pac = pac::Peripherals::take().unwrap();
    let core = pac::CorePeripherals::take().unwrap();

    // Set up the watchdog driver - needed by the clock setup code
    let mut watchdog = hal::Watchdog::new(pac.WATCHDOG);

    // Configure the clocks
    //
    // The default is to generate a 125 MHz system clock
    let clocks = hal::clocks::init_clocks_and_plls(
        rp_pico::XOSC_CRYSTAL_FREQ,
        pac.XOSC,
        pac.CLOCKS,
        pac.PLL_SYS,
        pac.PLL_USB,
        &mut pac.RESETS,
        &mut watchdog,
    )
    .ok()
    .unwrap();

    // The single-cycle I/O block controls our GPIO pins
    let sio = hal::Sio::new(pac.SIO);

    // Set the pins up according to their function on this particular board
[...]

PR #346, opened 2022-05-26
Add Pico PWM micro servo example

lucazulian:
Closes: #342

### DIFF
--- a/boards/rp-pico/examples/pico_pwm_servo.rs
+++ b/boards/rp-pico/examples/pico_pwm_servo.rs
@@ -12,18 +12,17 @@
 // The macro for our start-up function
 use cortex_m_rt::entry;
 
+use cortex_m::prelude::*;
+
 // GPIO traits
 use embedded_hal::PwmPin;
 
-// Time handling traits
-use embedded_time::rate::*;
+// Traits for converting integers to amounts of time
+use embedded_time::duration::Extensions;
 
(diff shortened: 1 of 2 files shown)


boards/rp-pico/examples/pico_pwm_servo.rs
```diff
@@ -0,0 +1,117 @@
...
+    let channel = &mut pwm.channel_b;
+    channel.output_to(pins.gpio1);
+
+    // Infinite loop, moving micro servo from one position to another.
+    // You may need to adjust the pulse width since several servos from
+    // different manufacturers respond differently.
+    loop {
+        // move to 0°
+        channel.set_duty(2500);
+        delay.delay_ms(400);
```
9names, line 101: Can you replace these delays with calls to the HAL CountDown timer instead? cortex_m::delay is only supposed to be used when you don't have a better delay source. It is quite inaccurate.
Here's an example that uses countdown:
https://github.com/rp-rs/rp-hal/blob/99b8845d7415875eb17e893d196db9d38bd531dc/boards/rp-pico/examples/pico_countdown_blinky.rs#L80

9names (changes requested):
Can you rename this PR to that reflects what it does. Something like "Add Pico PWM servo example".
I plan to squash-commit this PR, so this is important to make the git log useful.

9names (approved):

9names:
Thanks, LGTM

PR #347, opened 2022-05-27
HD44780 lcd display example from rp2040 setup specific for rp-pico

paddywwoof:
This is just an rp-pico specific version of the rp2040-hal example to use the hd44780-driver crate


9names:
Could you rename this example to make it more obvious who it is for?
Something like `pico_hd44780_display` would be better. Thanks!

boards/rp-pico/examples/pico_hd44780_display.rs
```diff
@@ -0,0 +1,137 @@
...
+//! been adapted from rp-hal/pr2040-hal/examples/lcd_display.rs for 
+//! the rp-pico board layout and pin assignment.
+//!
+//! See the `Cargo.toml` file for Copyright and license details.
+//!
+//! ```text              /--------------------------------------\
+//!  ____________        |        /-------------------------\   |
+//! | 1       GND|-------+---\    |           _|USB|_       |   |
+//! | 2       VDD|-------/   |    |          |1  R 40|-VBUS-o   v
+//! | 3        VS|-----------+----/          |2  P 39|       ||POT||
```
jannic, line 17: Are you sure about this wiring?
AFAIK, pin 2 (VDD) should be the fixed voltage, and pin 3 (often called V0) should be an adjustable value between GND and VDD. So it looks like you switched pins 2 and 3 on the display connector.

jannic:
Tried it on a display I happened to have: It works! (With switching pins 2/3 as suggested. I don't want to try it the other way around, because as I understand it, V0 > VDD violates the maximum ratings, and I don't want to break the display.)

paddywwoof:
@jannic Aaarg! Thank you so much, I've bamboozled myself with all this spaghetti (my "pot" is constructed from wires and resistors on the breadboard) and the construction of the ASCII wiring diagram.  Will fix it now.

boards/rp-pico/examples/pico_hd44780_display.rs
```diff
@@ -0,0 +1,137 @@
...
+//! alphanumeric LCD using the
+//! [HD44780](https://crates.io/crates/hd44780-driver) driver.
+//!
+//! It drives the LCD by pushing data out of six GPIO pins. This version has
+//! been adapted from rp-hal/pr2040-hal/examples/lcd_display.rs for 
+//! the rp-pico board layout and pin assignment.
+//!
+//! See the `Cargo.toml` file for Copyright and license details.
+//!
+//! ```text              /--------------------------------------\
```
jannic, line 13: Just a detail: In the generated docs, this line is not shown, but put into the `class` attribute, probably because it's in the same line as the block start marker:
```
<div class="example-wrap"><pre class="language-text              /--------------------------------------\"><code> ____________        |        /-------------------------\   |
| 1       GND|-------+---\    |           _|USB|_       |   |
| 2       VDD|-------+---+----/          |1  R 40|-VBUS-o   v
| 3        VS|-------/   |               |2  P 39|       ||POT||
| 4        RS|--\        o-----------GND-|3    38|-GND----------o
| 5        RW|--+--------/    /------GP2-|4  P 37|
```

jannic (approved):
LGTM

jannic (approved):

PR #353, opened 2022-06-04
Work around circular dependencies for ws2812-pio

jannic:
This is just a proof of concept - not really meant for merging, for now.

It intentionally includes the changes by @9names from https://github.com/rp-rs/rp-hal/pull/351


jannic:
#351 will be solved without vendoring, by referencing branches in rp-rs/i2c-pio-rs and rp-rs/ws2812-pio-rs, instead.

Issue #354, opened 2022-06-04
offset private in rp2040_hal::pio::InstalledProgram


fenax:
The offset a program is installed at in PIO memory is private, making me unable to use exec_instruction for JMP

9names:
This is fixed by #355

PR #355, opened 2022-06-04
accessor for offset of installed program on pio

fenax:
Make PIO installed program offset accessible with an accessor method
#354


jannic (approved):
Looks like a good addition, I don't see a reason why the offset should not be visible.

9names (approved):
I also don't see any good reason to not expose this to the user.
LGTM

PR #356, opened 2022-06-12
Add an example of using PWMs to generate an audio output

DrChat:
This PR merely adds a basic example of using the PWM controller to output an audio signal at 32kHz on GPIO0.

One of the more unique parts of this example is configuring the system clock to a nonstandard rate that is a multiple of the audio sample rate, so that the PWM clock can be divided by an integral number to deliver the audio samples.


paddywwoof:
Yes it would be very easy to synthesize the sound but I suppose @DrChat wants to demonstrate the ease of including sound files into the program. At least half the sample is empty, presumably to improve the impression when repeated (though that could be done just with a delay). If the blank parts are removed it would be 31.6k

The other thing is that it might be better using a WAV format file which would obviously be a sound sample and would have the advantage that it could be played by clicking on it in most file browsers. I can't find an option in Audactiy to save WAV as signed 8-bit so to do that the instructions would have to be
```rust
/// If you want to create your own, use Audacity to create a recording with a
/// sample rate of 32,000 Hz and then export it as WAV file coding Unsigned 8-bit PCM
const AUDIO: &[u8] = include_bytes!("pico_pwm_audio.wav");
```
and later
```rust
        for i in &AUDIO[44..] {
...
            let i = ((*i as u16) << 3).wrapping_add(2048) & 0xFFF;
```

DrChat:
Oh, huh that is clever - I didn't know you could just skip the wav header and directly play audio samples.
Let me play around with that, maybe I could even pull in a no_std wav parser and give it some flexibility.

As for the file size, I suppose technically it would be best to check it in as a git LFS file as that's the best practice for binary files.

And yeah, I figure it'll allow more creativity from users if the sample included an easy-to-replace audio file rather than a synthesized sound.

boards/rp-pico/examples/pico_pwm_audio.rs
```diff
@@ -0,0 +1,244 @@
...
+            cortex_m::interrupt::free(|_| {
+                // SAFETY: Interrupt cannot currently use this while we're in a critical section.
+                let channel = &mut unsafe { PWM.as_mut() }.unwrap().channel_a;
+                channel.set_duty(i);
+            });
+
+            // Throttle until the PWM channel delivers us an interrupt saying it's done
+            // with this cycle (the internal counter wrapped). The interrupt handler will
+            // clear the interrupt and we'll send out the next sample.
+            cortex_m::asm::wfi();
```
jannic, line 226: This only works if the PWM interrupt is the only one enabled.
Even for an example, I'd prefer a way which can be composed with other functionality.
Also, see [the documentation of WFI](https://developer.arm.com/documentation/dui0662/b/The-Cortex-M0--Instruction-Set/Miscellaneous-instructions/WFI): "WFI is intended for power saving only. When writing software assume that WFI might behave as a NOP operation."

jannic:
As my previous comments may sound somewhat negative: I really like the idea of an example outputting sounds! Just blinking LEDs gets boring. Bonus points for simple hardware (ie. not needing external i2s hardware or similar).

boards/rp-pico/examples/pico_pwm_audio.rs
```diff
@@ -0,0 +1,244 @@
+//! # Pico PWM Audio Example
+//!
+//! Drives GPIO0 with PWM to generate an audio signal for use with a speaker.
+//!
+//! Note that you will need to supply your own speaker. When hooked up to GPIO0,
+//! you should hear an audible chime.
```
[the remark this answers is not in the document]
paddywwoof: Yes I'm sure it's not good, especially something that might have inductance. I found a 2N2222 transistor with resistors already soldered from another project (tiny motor from arduino I think) and that makes it a bit louder (a *bad* thing after only a few seconds!) and safer. But it does add some external hardware. Maybe just mention that a transistor and resistors would be a good idea.

boards/rp-pico/examples/pico_pwm_audio.rs
paddywwoof: I agree, projects playing sounds are likely to have other hardware inputs such as ADC, buttons etc. so not ideal if the example stops working when other functionality is introduced.

WeirdConstructor:
Not too long ago I built this, it's two simple RC filters in a row and some code to generate a saw tooth:
image

paddywwoof:
@WeirdConstructor that looks a tidy example. I see on your repo you have switched from the `delay_us()` system in your screenshot to `nb::block!(cnt_down.wait())` which might be the way for @DrChat to go (though it probably removes the need for the clever clock setting section!)

It reminded me of something I did ages ago with at atmega32 https://www.youtube.com/watch?v=HrDFpDNN2cw I'm not sure where the code is for that now but I think I avoided floats by using lookup tables, based on this sketch by Adrian Freed adrianfreed.com/content/arduino-sketch-high-frequency-precision-sine-wave-tone-sound-synthesis

boards/rp-pico/examples/pico_pwm_audio.rs
DrChat: Ah good point. You guys think something like `wfe` and `sev` would be better instead?

boards/rp-pico/examples/pico_pwm_audio.rs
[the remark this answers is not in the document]
DrChat: Sure, that may be a good point! I hope I'm not damaging my Pi by wiring a speaker straight up to the GPIO pin haha.
What would be a better wording? My knowledge about software engineering is certainly better than my knowledge of electrical engineering.

boards/rp-pico/examples/pico_pwm_audio.rs
paddywwoof: The approach used in @WeirdConstructor's code https://github.com/WeirdConstructor/RustRP2040Code/blob/master/pwm_dac_saw_sampling/src/example_1000hz_saw.rs which I think is the same as the pwm_blinky example seems much simpler. However there might be disadvantages I'm unaware of.

9names:
A reminder for who-ever merges this: be sure to squash to avoid both audio files ending up in the repo history.

9names:
Section 3.1.4 of [Hardware Design with RP2040](https://datasheets.raspberrypi.com/rp2040/hardware-design-with-rp2040.pdf) outlines a good circuit for getting line-out audio with PWM.
https://github.com/rgrosset/pico-pwm-audio has an circuit that is simple enough to breadboard, which feels appropriate for the level of this example. At the very least, the fact that it has *some* DC blocking would make me more comfortable.

It also has a C project that uses the same system-frequency-matched PWM style that this example does, but they also switch the power supply from PFM to PWM mode to reduce switching noise (at the cost of increased power consumption).
It's only a single GPIO write, should we do that?

DrChat:
Oh, huh that is clever - I didn't know you could just skip the wav header
and directly play audio samples.
Let me play around with that, maybe I could even pull in a no_std wav
parser and give it some flexibility.

As for the file size, I suppose technically it would be best to check it in
as a git LFS file as that's the best practice for binary files.

And yeah, I figure it'll allow more creativity from users if the sample
included an easy-to-replace audio file rather than a synthesized sound.

PR #358, opened 2022-06-14
Bump version of board support crates

jannic:
As the BSPs re-export the hal, their version needs to be bumped after the release of rp2040-hal.

Also updates the changelogs and removes broken links.


9names (approved):
Ah, you removed all the tag links as well. Good.
Maintaining them was a lot of work for no benefit.

Looks good to me.

PR #361, opened 2022-06-17
Migrate to fugit

ithinuel:
This is an experimental migration to `fugit` in order to gather feedback and help decide if we actually want to migrate or not.

embedded-time was preserved to cover for `Baud` used in the uart module and that is not present in current version of `fugit`.

**Note:** The last commit needs to be trimmed off once the pio crates have been updated/approved.


rp2040-hal/src/xosc.rs
```diff
@@ -89,13 +83,12 @@ impl CrystalOscillator<Disabled> {
     }
 
     /// Initializes the XOSC : frequency range is set, startup delay is calculated and set.
-    pub fn initialize(self, frequency: Hertz) -> Result<CrystalOscillator<Initialized>, Error> {
+    pub fn initialize(self, frequency: HertzU32) -> Result<CrystalOscillator<Initialized>, Error> {
         const ALLOWED_FREQUENCY_RANGE: RangeInclusive<Megahertz<u32>> =
-            Megahertz(1)..=Megahertz(15);
-        const STABLE_DELAY: Milliseconds = Milliseconds(1_u32);
-        const DIVIDER: Fraction = Fraction::new(256, 1);
```
hmvp, line 96: I feel its somewhat worth it to keep the const even if its just `const DIVIDER: u32 = 256;`

rp2040-hal/src/xosc.rs
```diff
@@ -89,13 +83,12 @@ impl CrystalOscillator<Disabled> {
...
+            MegahertzU32::MHz(1)..=MegahertzU32::MHz(15);
```
hmvp, line 88: If you use `const ALLOWED_FREQUENCY_RANGE: RangeInclusive<HertzU32> =
            HertzU32::MHz(1)..=HertzU32::MHz(15);`

you can lose the convert on line 91

rp2040-hal/src/xosc.rs
```diff
@@ -107,20 +100,19 @@ impl CrystalOscillator<Disabled> {
         });
 
         //1 ms = 10e-3 sec and Freq = 1/T where T is in seconds so 1ms converts to 1000Hz
-        let delay_to_hz: Hertz = STABLE_DELAY.to_rate().map_err(|_| Error::BadArgument)?;
+        let delay_to_hz: HertzU32 = STABLE_DELAY.try_into_rate().ok_or(Error::BadArgument)?;
```
hmvp, line 103: You could move the `try_into_rate` call into line 89 (where STABLE_DELAY is declared..

Also you can just unwrap since we know the result is always ok and even if it isn't. It was not a bad argument...

rp2040-hal/src/xosc.rs
```diff
@@ -107,20 +100,19 @@ impl CrystalOscillator<Disabled> {
...
+        //startup_delay = ((freq_hz * STABLE_DELAY) / 256) = ((freq_hz / delay_to_hz) / 256)
         //See Chapter 2, Section 16, §3)
         //We do the calculation first.
         let startup_delay = frequency
-            .checked_div(&delay_to_hz.integer())
-            .and_then(|r| r.to_generic::<u32>(DIVIDER).ok())
+            .to_Hz()
+            .checked_div(delay_to_hz.to_Hz())
+            .map(|r| r / 256)
             .ok_or(Error::BadArgument)?;
```
hmvp, line 112: I think this can be simplified to `let startup_delay = frequency / delay_to_hz / 256;`

Again we know that delay_to_hz is never going to be 0, so checked_div makes less sense...

rp2040-hal/src/timer.rs
```diff
@@ -103,26 +103,26 @@ impl Timer {
...
-    type Time = embedded_time::duration::Microseconds<u64>;
+    type Time = MicrosDurationU64;
 
     fn start<T>(&mut self, count: T)
     where
         T: Into<Self::Time>,
     {
         self.period = count.into();
-        self.next_end = Some(self.timer.get_counter().wrapping_add(self.period.0));
+        self.next_end = Some(self.timer.get_counter().wrapping_add(self.period.ticks()));
```
hmvp, line 118: Isnt it nicer to use `to_micros()` to make it more explicit especially if someone might want to change `self.period` to another base?

rp2040-hal/src/timer.rs
```diff
@@ -103,26 +103,26 @@ impl Timer {
...
     }
 
     fn wait(&mut self) -> nb::Result<(), void::Void> {
         if let Some(end) = self.next_end {
             let ts = self.timer.get_counter();
             if ts >= end {
-                self.next_end = Some(end.wrapping_add(self.period.0));
+                self.next_end = Some(end.wrapping_add(self.period.ticks()));
```
hmvp, line 125: See above

rp2040-hal/src/clocks/macros.rs
```diff
@@ -352,12 +354,12 @@ macro_rules! stoppable_clock {
                     // Disable clock. On clk_ref and clk_sys this does nothing,
                     // all other clocks have the ENABLE bit in the same position.
                     self.disable();
-                    if (self.frequency > 0u32.Hz()) {
+                    if self.frequency > HertzU32::from_raw(0) {
```
hmvp, line 357: Feel more explicit even if practically useless to use `Hz(0)` performance wise it should not matter since its all const afaik

hmvp (commented):
Oh man fugit looks so much nicer than embedded_time... 

Some nitpicks from me (mostly based on my own try)

rp2040-hal/src/pll.rs
```diff
@@ -161,24 +150,26 @@ impl<D: PhaseLockedLoopDevice> PhaseLockedLoop<Disabled, D> {
             return Err(Error::PostDivOutOfRage);
         }
 
-        let ref_freq_range: Range<Hertz<u32>> = Hertz(5_000_000)..vco_freq.div(16);
+        use fugit::RateExtU32;
```
hmvp, line 153: I would put this at the top of the file

rp2040-hal/src/pll.rs
```diff
@@ -161,24 +150,26 @@ impl<D: PhaseLockedLoopDevice> PhaseLockedLoop<Disabled, D> {
...
+        let ref_freq_max_vco = (vco_freq.to_Hz() / 16).MHz();
```
hmvp, line 154: this should be `let ref_freq_max_vco = (vco_freq.to_Hz() / 16).Hz();` Note the dropped M from the last call

Since you convert vco_freq to Hz the value, even divided by 16 will be to big for HertzU32.Mhz() to handle

rp2040-hal/src/timer.rs
```diff
@@ -172,9 +172,9 @@ pub trait Alarm {
     /// reason, the minimum time that this function accepts is `10.microseconds()`
     ///
     /// [enable_interrupt]: #method.enable_interrupt
-    fn schedule<TIME: Into<Microseconds>>(
```
hmvp, line 175: Trying to use this I find the ergonomics a bit weird (might be me)...

So in converting I went from: 
`alarm0.schedule(TIME_TICK_MS.microseconds() * 1000);`
to
`alarm0.schedule(TIME_TICK_MS.micros() * 1000);`
which gave me an error because schedule is generic over the input duration NOM and DENOM. Which forces met to do:
`alarm0.schedule::<1, 1000_000>(TIME_TICK_MS.micros() * 1000)`

Since we convert to millis below, does it not make sense to just accept MillisDurationU32 since the caller can easily convert?

rp2040-hal/src/timer.rs
```diff
@@ -172,9 +172,9 @@ pub trait Alarm {
...
+    fn schedule<const NOM: u32, const DENOM: u32>(
```
hmvp, line 175: Trying to use this I find the ergonomics a bit weird (might be me)...

So in converting I went from: 
`alarm0.schedule(TIME_TICK_MS.microseconds() * 1000);`
to
`alarm0.schedule(TIME_TICK_MS.micros() * 1000);`
which gave me an error because schedule is generic over the input duration NOM and DENOM. Which forces met to do:
`alarm0.schedule::<1, 1000_000>(TIME_TICK_MS.micros() * 1000)`

Since we convert to millis below, does it not make sense to just accept MillisDurationU32 since the caller can easily convert?

hmvp: Mwa maybe not relevant. If I structure my code better this becomes a non issue...

rp2040-hal/src/timer.rs
```diff
@@ -240,11 +240,11 @@ macro_rules! impl_alarm {
...
             ///
             /// [enable_interrupt]: #method.enable_interrupt
-            fn schedule<TIME: Into<Microseconds>>(
+            fn schedule<const NOM: u32, const DENOM: u32>(
                 &mut self,
-                countdown: TIME,
+                countdown: Duration<u32, NOM, DENOM>,
             ) -> Result<(), ScheduleAlarmError> {
-                let duration = countdown.into().0;
+                let duration = countdown.to_millis();
```
hmvp, line 247: This should be `to_micros`

rp2040-hal/src/timer.rs
```diff
@@ -240,11 +240,11 @@ macro_rules! impl_alarm {
             /// The RP2040 has been observed to take a little while to schedule an alarm. For this reason, the minimum time that this function accepts is `10.microseconds()`
```
hmvp, line 240: Maybe also change the comment (and maybe others)

rp2040-hal/src/xosc.rs
ithinuel: That's true. I also realize we could also store `STABLE_DELAY` directly as a HertzU32 and skip all that.

rp2040-hal/src/timer.rs
ithinuel: You're probably right. Although, something feels wrong.
The assumption that the counter increments every us only holds if the user configures the ref clk of the watchdog to be at that frequency.
If that's not the case most of our code would be off.

rp2040-hal/src/timer.rs
ithinuel: I don't understand what you mean.

Doesn't it work if you use `alarm0.schedule(TIME_TICK_MS)` ? (note there's indeed an error with to_micro/to_millis)

rp2040-hal/src/timer.rs
hmvp: As far as I can see in the reference manual the watchdog and timer tick interval is based on ref_clock but internally divided to always be as close to 1us per tick.

rp2040-hal/src/timer.rs
hmvp: No in the example TIME_TICKS_MS was an u32. If you make that a proper duration then there is no issue.

However I found that the following is problematic:
`alarm0.schedule(10.micros())` 
You might argue that a constant could be created for that as well. But thats sometimes overkill

rp2040-hal/src/timer.rs
ithinuel: The division is controlled by [`init_init_clocks_and_plls`](https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/src/clocks/mod.rs#L332) which sets the `CYCLES` field of the `TICK` register described in [4.7.6](https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf#reg-watchdog-TICK).

So while unlikely, it is possible for a user to use a different value there.

Note:
I noticed that the `CYCLE` field is 9bits (8:0) wide while our method only takes and 8bit value.
I'll have to fix that, and introduce an extra `watchdog::Error::BadArgument` to `watchdog::Watchdog::enable_tick_generation`.

jannic:
The PR still contains the patch entries from https://github.com/rp-rs/rp-hal/commit/46d25400403ba342a189a0dda5cc5f5b569dda3d

ithinuel:
Yes it is stated in the description.
A new version of the -pio crates will be required. They need to be synced with this pr or it may break user's repo after a `cargo update` (or a fresh build without Cargo.lock)

9names (approved):
LGTM.
This yields a larger reduction in binary size than I expected, at least for the example I chose to benchmark. Feels good to get massive reductions, even if it's only for a toy example.

Before
```console
$ cargo size --release --example adafruit_itsy_bitsy_rainbow
Finished release [optimized] target(s) in 4.31s
   text    data     bss     dec     hex filename
  19808       0       4   19812    4d64 adafruit_itsy_bitsy_rainbow
```

After
```console
$ cargo size --release --example adafruit_itsy_bitsy_rainbow
Finished release [optimized] target(s) in 4.00s
   text    data     bss     dec     hex filename
  12640       0       4   12644    3164 adafruit_itsy_bitsy_rainbow

```

PR #362, opened 2022-06-19
BSP fixes

9names:
A small collection of changes from when testing splitting the BSPs out into separate repos.

### DIFF
--- a/boards/adafruit-feather-rp2040/README.md
+++ b/boards/adafruit-feather-rp2040/README.md
@@ -16,7 +16,7 @@ RP2040 chip according to how it is connected up on the Feather.
 To use this crate, your `Cargo.toml` file should contain:
 
 ```toml
-adafruit-feather-rp2040 = "0.2.0"
+adafruit-feather-rp2040 = "0.3.0"
 ```
 
 In your program, you will need to call `adafruit_feather_rp2040::Pins::new` to create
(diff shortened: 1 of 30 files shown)


9names:
I've left this with just the changes that I think are uncontroversial so we can get these minor doc/packaging fixes in place for the BSPs

boards/adafruit-feather-rp2040/README.md
```diff
@@ -16,7 +16,7 @@ RP2040 chip according to how it is connected up on the Feather.
 To use this crate, your `Cargo.toml` file should contain:
 
 ```toml
-adafruit-feather-rp2040 = "0.2.0"
+adafruit-feather-rp2040 = "0.3.0"
```
jannic, line 19: Could be:
```suggestion
adafruit-feather-rp2040 = "0.3.1"
```
(Doesn't matter much due to semver.)

jannic (approved):

boards/adafruit-feather-rp2040/README.md
9names: Yeah. I was relying on sem ver for these
Think I should change it to two figures (eg "0.3" to make this intention more clear?
We don't want to encourage PRs bumping just the patch version

9names: Going to merge without making a decision about this since I managed to break examples on main (by deleting the 0.5.0 branches prematurely)

jannic: Sure, it was just a minor nitpick, no need to delay the merge.

PR #364, opened 2022-06-22
Add bsp_pin example

### LINKED ISSUE #262: Explain the bsp_pins macro
Reading https://github.com/rp-rs/rp-hal/pull/259/files made me realise that the bsp_pins macro could probably do with some explanatory text. Like,

```
/// Calling the macro like this:
/// ```
/// bsp_pins! {
///    Gpio0 { name: gpio0 },
/// }
/// ```
/// 
/// Is equivalent to the following source code:
/// ```
/// <I have no idea what goes here>
/// ```
```

hmvp:
Fixes #262

### DIFF
--- a/rp2040-hal/src/gpio/pin.rs
+++ b/rp2040-hal/src/gpio/pin.rs
@@ -1075,7 +1075,56 @@ gpio!(
 /// [here](self#types) for a list of the available [`PinMode`] type aliases.
 ///
 /// # Example
-/// ...
+/// Calling the macro like this:
+/// ```rust
+/// use rp2040_hal::bsp_pins;
+/// bsp_pins! {
+///     #[cfg(feature = "gpio")]
+///     Gpio0 {
+///          /// Doc gpio0
+///          name: gpio0,
+///          aliases: { FunctionPio0: PioPin }
+///     },
+///     Gpio1 {
+///          name: led,
+///          aliases: { FunctionPwm: LedPwm }
+///      },
+/// }
+/// ```
+///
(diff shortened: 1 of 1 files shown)


rp2040-hal/src/gpio/pin.rs
```diff
@@ -1075,7 +1075,55 @@ gpio!(
...
+/// }
+/// ```
+///
+/// Is roughly equivalent to the following source code (excluding the docs strings below):
+/// ```
+/// use ::rp2040_hal as hal;
+/// use hal::gpio;
+/// pub struct Pins {
+///     # [cfg (feature != "gpio")]
+///     /// Doc gpio0"]
```
jannic, line 1100: The last two chars, `"]`, look a little bit weird. Is this intended / correct?

rp2040-hal/src/gpio/pin.rs
```diff
@@ -1075,7 +1075,55 @@ gpio!(
...
+///      },
+/// }
+/// ```
+///
+/// Is roughly equivalent to the following source code (excluding the docs strings below):
+/// ```
+/// use ::rp2040_hal as hal;
+/// use hal::gpio;
+/// pub struct Pins {
+///     # [cfg (feature != "gpio")]
```
jannic, line 1099: `#[cfg(not(feature = "gpio"))]` ?

hmvp: Yeah some hacky workaround, I removed it already

9names:
Looks good, thanks!

Issue #154, opened 2021-10-04
PIO example which loads a .pio file


thejpster:
If you are working with the Pico SDK, you might have some PIO code written out in a .pio file. We should have an example of some Rust code that can load the .pio file at compile time.

doraneko94:
Do you want to add this example to the `examples` in this repository? 
Or is it sufficient since it currently exists in `rp_hal_boards`?

ithinuel:
IMHO, rp-hal-boards is good enough.

jannic:
https://github.com/rp-rs/rp-hal-boards/blob/main/boards/rp-pico/examples/pico_pio_pwm.rs#L132 does exactly that. So I think this ticket can be closed?

PR #365, opened 2022-06-22
Add pio pwm example

hmvp:
This adds a more advance pio example which uses side set and instruction injection

### DIFF
--- a/boards/rp-pico/examples/pico_pio_pwm.rs
+++ b/boards/rp-pico/examples/pico_pio_pwm.rs
@@ -1,0 +1,162 @@
+//! # Pico PIO PWM Blink Example
+//!
+//! Fades the LED on a Pico board using the PIO peripheral with an pwm program.
+//!
+//! This will fade in the LED attached to GP25, which is the pin the Pico
+//! uses for the on-board LED.
+//!
+//! This example uses a few advance pio tricks such as side setting pins and instruction injection.
+//!
+//! See the `Cargo.toml` file for Copyright and license details. Except for the pio program which is subject to a different license.
+
+#![no_std]
+#![no_main]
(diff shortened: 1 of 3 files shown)


hmvp:
Moved pio code to seperate file so we can also close  #154

boards/rp-pico/examples/pico_pio_pwm.rs
```diff
@@ -0,0 +1,162 @@
...
+/// This uses a sneaky trick to set a second value besides the duty cycle.
+/// We first write a value to the tx fifo. But instead of the normal instructions we
+/// have stopped the state machine and inject our own instructions that move the written value to the ISR.
+fn pio_pwm_set_period<T: ValidStateMachine>(
+    sm: StateMachine<(hal::pac::PIO0, SM0), Running>,
+    tx: &mut Tx<T>,
+    period: u32,
+) -> StateMachine<(hal::pac::PIO0, SM0), Running> {
+    let mut sm = sm.stop();
+    tx.write(period);
```
jannic, line 54: Calling `pio_pwm_set_period` shortly after `pio_pwm_set_level` could have unintended results: If the value written to the tx fifo by `pio_pwm_set_level` has not yet been read by the PIO program before the state machine is stopped, the level will be interpreted as a period.

Perhaps add `while !tx.is_empty() {}` before the call to `sm.stop()`?

hmvp: Interesting.. You are completely right. I based the example on https://github.com/raspberrypi/pico-examples/blob/master/pio/pwm/pwm.c which has the same issue...

9names (approved):
LGTM

PR #371, opened 2022-06-27
Interpolator

fenax:
Support for the hardware interpolator of the RP2040
I am using it in one project and it works as expected.

Strange thing I did :
Each lane is accessed through a function returning a mutable reference to the lane to ensure the whole interpolator is borrowed as long as we hold to the reference as both lane are linked and can mutate each other. Maybe there is a cleaner way to achieve this.


rp2040-hal/src/sio.rs
```diff
@@ -667,3 +676,224 @@ pub unsafe fn spinlock_reset() {
...
+///Trait representing the functionnality of an interpolator.
+/// ```no_run
+/// use rp2040_hal::sio::{Sio,LaneCtrl,Lane};
+/// use rp2040_hal::pac;
+/// let mut peripherals = pac::Peripherals::take().unwrap();
+/// let mut sio = Sio::new(peripherals.SIO);
+///
+/// // by having the configuration const, the validity is checked during compilation.
+/// const config: u32 = LaneCtrl {
+///     mask_msb: 4, // Most significant bit is the fourt
```
jannic, line 776:
```suggestion
///     mask_msb: 4, // Most significant bit is the fourth
```
Nitpick: I wonder if 4th is the best way to describe a bit. While it feels ok to number bits 0,1,2,3,4..., personally I'd say something like "bit 0 is the first bit, bit 1 is the second bit....". So the fourth bit would be bit 3.

jannic (commented):
Code looks clean, but I never used the interpolator and have no idea how it works. Any reviewer available who knows more about the hardware?

9names:
Code looks good to me to, but without examples it's hard to gauge usability.
Do you think we could add a port of the example from https://github.com/raspberrypi/pico-examples/blob/master/interp/hello_interp/hello_interp.c?

boards/rp-pico/examples/pico_interpolator.rs
```diff
@@ -0,0 +1,441 @@
...
+    // reset the starting position
+    interp.get_lane0().set_accum(0);
+    interp.get_lane1().set_accum(0);
+    interp.set_base(texture.as_ptr() as u32);
+
+    for i in expected {
+        // This is unsafe and should be done extremely carefully
+        // remember to follow memory alignment,
+        // reading or writing an unaligned address will crash
+        if i != unsafe { *(interp.pop() as *const u8) } {
```
9names, line 434: interp.pop() always returns u32, right?
does it make sense to use raw pointers here, when `interp.pop() as u8` or `(interp.pop & 0xFF) as u8`(if you want to be explicit) would be equivalent?
if you want to do a u8 volatile read at that address as the C example does you'll have to drop down to pac level (or add a pop_u8 variant)

9names (changes requested):
Thanks for adding the example, you've done a great job porting it across.
I think it showcases the driver very well, and also reveals the API differences between the C SDK and Rust HAL which will come in handy for porting C drivers using this hardware.

ithinuel (changes requested):

boards/rp-pico/examples/pico_interpolator.rs
fenax: It is to be closest to the example in c++
it reads u32 from the pop register
which I cast into a pointer to u8.
the u8 read is from the array texture: [u8;16]

rp2040-hal/src/sio.rs
fenax: I reworded the whole comment, I hope it is clearer like this

rp2040-hal/src/sio.rs
```diff
@@ -667,3 +676,236 @@ pub unsafe fn spinlock_reset() {
...
+    pub mask_msb: u8,
+    #[doc = "Bits 5:9 - The least-significant bit allowed to pass by the mask (inclusive)"]
+    pub mask_lsb: u8,
+    #[doc = "Bits 0:4 - Logical right-shift applied to accumulator before masking"]
+    pub shift: u8,
+}
+
+impl LaneCtrl {
+    /// Default configuration. Normal operation, unsigned, mask keeps all bits, no shift.
+    pub const DEFAULT: Self = Self {
```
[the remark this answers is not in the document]
fenax: I agree, I changed it

rp2040-hal/src/sio.rs
```diff
@@ -667,3 +676,236 @@ pub unsafe fn spinlock_reset() {
         SPINLOCK0_PTR.wrapping_add(i).write_volatile(1);
     }
 }
+
+/// Configuration struct for one lane of the interpolator
+pub struct LaneCtrl {
+    #[doc = "Bit 22 - Only present on INTERP1 on each core. If CLAMP mode is enabled:
```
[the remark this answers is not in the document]
fenax: Done

boards/rp-pico/examples/pico_interpolator.rs
9names: Oh... wow, I totally misread what this was doing.
Thanks for clarifying, this makes sense now.
I think I might have a go at documenting this in the future, since I doubt I'll be the only one to get confused by this,

9names (approved):
This looks good to me :+1:

ithinuel (approved):
LGTM too.

Only needs a rebase (and maybe some nits picked by CI) before merge.

PR #380, opened 2022-07-09
Add pico-w BSP

jannic:
This is very much a WIP - don't take anything for granted.


boards/rp-pico-w/examples/pico_w_blinky.rs
```diff
@@ -0,0 +1,83 @@
+//! # Pico Blinky Example
+//!
+//! Blinks the LED on a Pico board.
+//!
+//! This will blink an LED attached to GP25, which is the pin the Pico uses for
+//! the on-board LED.
```
9names, line 6:
```suggestion
//! # Pico W Blinky Example
//!
//! Blinks the LED on a Pico W board.
//!
//! This will blink an LED attached to WL_GPIO0, which is the pin the Pico W uses for
//! the on-board LED. It is connected to the the Wifi chip so it cannot be set using RP2040 pins
```

9names (changes requested):

DavidBChristensen:
@jannic Great start! Are you developing the WiFi Driver, or are you looking for contributors?

9names:
@DavidBChristensen there is already a wifi driver that is working with embassy-rp
https://github.com/embassy-rs/cyw43
It currently uses gpio bitbanging to talk to the wifi chip. We need to bring that up with rp-hal, port the PIO driver from pico-sdk, and hook that in.
If you think you can do any of those tasks, feel free to do so.

jannic:
I definitely plan to somehow integrate the driver into the pico-w BSP, but I didn't start yet. Any help is appreciated!

jannic:
The version in this branch is now able to join a wireless network. It does respond to pings!

Probably other smoltcp features work as well, I just didn't test them yet. However, it's not yet stable: Bombarding the device with IP packets quickly leads to error messages from the cyw43 driver and subsequent firmware crashes.

FireAndIceFrog:
Is this still active? What happened here?

jannic:
It's currently on hold because I want rp-hal to work on stable rust, and some of the dependencies from embassy are still nightly-only. It might be possible to fork them and build a less ergonomic variant which works on stable. However, I don't have time to do that.
Sooner or later, either rustc will stabilize the needed features, or somebody (perhaps me) will become so impatient that a port to stable rust will be done :-)
If you want to use the pico-w with rust now, your best bet is using embassy.

9names:
What do we need in the HAL to support this?
If it's just because they're in the same workspace, can we add an exclude section in the workspace file for it, or put it in a repo by itself for the time being?

jannic:
> What do we need in the HAL to support this that requires nightly? If it's just because they're in the same workspace (or because it will make our CI messy), can put it in a repo by itself for the time being?

In fact I misremembered, the nightly requirement was already solved. `pico_w_blinky` compiles on stable rust and works.
What's preventing a release to crates.io are some unreleased dependencies:
- cyw43
- embassy-net
- embassy-executor

carloo1:
> some unreleased dependencies

Now that the embassy crates have been released, can this PR be completed?

PR #391, opened 2022-07-21
Adafruit macropad macros

mutantbob:
This patch creates a set of macros in the adafruit-macropad board crate which simplify instantiation of the hardware that is built in to the Adafruit Macropad.


mutantbob:
Doh, accidentally included the unmerged showcase.

PR #392, opened 2022-07-21
add Adafruit Macropad macros

mutantbob:
Add a bunch of macros in the adafruit-macropad board crate to make it easier to construct working peripheral objects.

With this patch your `main()` can start like this:

```Rust
let mut pac = pac::Peripherals::take().unwrap();
let core = pac::CorePeripherals::take().unwrap();
let mut watchdog = Watchdog::new(pac.WATCHDOG);
let clocks = macropad_clocks!(pac, watchdog);
let mut delay = cortex_m::delay::Delay::new(core.SYST, clocks.system_clock.freq().integer());
let pins = macropad_pins!(pac);
let timer = Timer::new(pac.TIMER, &mut pac.RESETS);
//
let mut disp = macropad_oled!(pins, clocks, delay, pac);
let mut led_pin = pins.led.into_push_pull_output();
let mut neopixels = macropad_neopixels!(pins, clocks, timer, pac);
//
let mut rotary = macropad_rotary_encoder!(pins);
let keys = macropad_keypad!(pins);
```

### DIFF
--- a/boards/adafruit-macropad/src/macros.rs
+++ b/boards/adafruit-macropad/src/macros.rs
@@ -8,7 +8,8 @@
 //! let mut watchdog = Watchdog::new(pac.WATCHDOG);
 //! let clocks = macropad_clocks!(pac, watchdog);
 //! let mut delay = cortex_m::delay::Delay::new(core.SYST, clocks.system_clock.freq().integer());
-//! let pins = macropad_pins!(pac);
+//! let mut sio = Sio::new(pac.SIO);
+//! let pins = macropad_pins!(pac, sio);
 //! let timer = Timer::new(pac.TIMER, &mut pac.RESETS);
 //! //
 //! let mut disp = macropad_oled!(pins, clocks, delay, pac);
@@ -33,6 +34,7 @@ pub use core::convert::Infallible;
 pub use embedded_hal::digital::v2::InputPin;
 pub use embedded_hal::spi::MODE_0;
 pub use embedded_time::rate::Extensions;
+pub use rotary_encoder_hal::Rotary;
 pub use sh1106::prelude::GraphicsMode;
 
 // XXX missing: speaker, I2C side port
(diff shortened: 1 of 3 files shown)


mutantbob:
I could use some help figuring out how to make the CI checks pass.  It looks like `cargo test --doc --target x86_64-unknown-linux-gnu` is malfunctioning.  

if I remove the code sample from the top of `macro.rs` the test passes fine, but then the user of the library will not be able to benefit from the example.

jannic:
You could add the `ignore` attribute to the code block. https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes
That way, it stays part of the documentation, but won't be run as a doc test.

boards/adafruit-macropad/src/macros.rs
```diff
@@ -0,0 +1,209 @@
...
+    pub key11: Pin<Gpio11, PullUpInput>,
+    pub key12: Pin<Gpio12, PullUpInput>,
+}
+
+impl KeysTwelve {
+    pub fn get_0based(&self, idx: i8) -> Option<&dyn InputPin<Error = Infallible>> {
+        self.get_1based(1 + idx)
+    }
+
+    pub fn get_1based(&self, idx: i8) -> Option<&dyn InputPin<Error = Infallible>> {
```
9names, line 175: Do you expect people to use both of these interfaces?
If you don't, it would best to pick the one you think is most useful, add a doc comment indicating how it's supposed to be used, and leave it with the name `get`

boards/adafruit-macropad/src/macros.rs
```diff
@@ -0,0 +1,209 @@
...
+            8 => Some(&self.key8),
+            9 => Some(&self.key9),
+            10 => Some(&self.key10),
+            11 => Some(&self.key11),
+            12 => Some(&self.key12),
+            _ => None,
+        }
+    }
+
+    pub fn array_0based(&self) -> [&dyn InputPin<Error = Infallible>; 12] {
```
9names, line 193: Doesn't really need the 0based suffix, there's no other base option and it's pretty obvious it's going to consume all the pins. Maybe name it `to_array` to match standard Rust naming conventions?

Also, since this moves the pins permanently out of `KeysTwelve`, do we want a way to convert from `[&dyn InputPin<Error = Infallible>; 12]` back to `KeysTwelve`?

9names (changes requested):

ithinuel:
This PR should be moved to https://github.com/rp-rs/rp-hal-boards.
I am closing it here but be assured these changes would be welcome over there :) .

PR #394, opened 2022-07-22
Implement conversion from Clock to Hertz using reference

jannic:
Implementing `impl From<SystemClock> for Hertz` is a footgun, as
SystemClock is not Copy, so the automatic conversion consumes the
owned clock.

This is visible in the example i2c.rs:

```
    let mut i2c = hal::I2C::i2c1(
        pac.I2C1,
        sda_pin,
        scl_pin, // Try `not_an_scl_pin` here
        400.kHz(),
        &mut pac.RESETS,
        clocks.peripheral_clock,
    );
```

If the user wants to use both `i2c0` and `i2c1` at the same time,
copying from this example won't work:

```
error[E0382]: use of moved value: `clocks.peripheral_clock`
   --> rp2040-hal/examples/i2c.rs:106:9
    |
97  |         clocks.peripheral_clock,
    |         ----------------------- value moved here
...
106 |         clocks.peripheral_clock,
    |         ^^^^^^^^^^^^^^^^^^^^^^^ value used here after move
    |
    = note: move occurs because `clocks.peripheral_clock` has type
`PeripheralClock`, which does not implement the `Copy` trait
```

As getting the frequency from a clock doesn't really need ownership,
changing it to `impl From<&SystemClock> for Hertz` is both more
logical and provides better usability.

This is, however, a breaking change: Code relying on this trait
implementation needs to be changed by adding a `&`.

### DIFF
--- a/boards/rp-pico/examples/pico_i2c_oled_display_ssd1306.rs
+++ b/boards/rp-pico/examples/pico_i2c_oled_display_ssd1306.rs
@@ -134,7 +134,7 @@ fn main() -> ! {
         scl_pin,
         400.kHz(),
         &mut pac.RESETS,
-        clocks.peripheral_clock,
+        &clocks.peripheral_clock,
     );
 
     // Create the I²C display interface:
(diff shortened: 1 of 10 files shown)


9names (approved):
LGTM. nice catch on the different clock source.
it's kinda depressing that the clock tree diagram in the manual doesn't actually cover all peripherals to make this easier to check.

PR #395, opened 2022-07-23
pimoroni display pack 2 example

jgmartin:
adds an example of using the [pimoroni display pack 2.0](https://shop.pimoroni.com/products/pico-display-pack-2-0?variant=39374122582099) with embedded_graphics.


boards/rp-pico/examples/pico_display_pack_2.rs
```diff
@@ -0,0 +1,170 @@
...
+    // create a filled circle
+    let circle =
+        Circle::new(Point::new(100, 60), 120).into_styled(PrimitiveStyle::with_fill(Rgb565::RED));
+
+    // draw the circle on a black background
+    display.clear(Rgb565::BLACK).unwrap();
+    circle.draw(&mut display).unwrap();
+
+    loop {
+        continue; // keep optimizer from removing in --release
```
jannic, line 168: I don't think this is still necessary. https://github.com/phil-opp/blog_os/issues/793

jannic:
Oops - sorry about that. I thought that clippy lint didn't apply to no_std crates?
https://github.com/rust-lang/rust-clippy/pull/5086 but it seems it was re-enabled later with https://github.com/rust-lang/rust-clippy/issues/6161

This is no longer a correctness issue, so you could just disable the clippy warning. Or add `cortex_m::asm::nop();`, as `pico_i2c_pio.rs` does. Or, of course, the `continue` you had before, just without the comment regarding the optimizer.

jannic:
@jgmartin, is https://github.com/rp-rs/rp-hal/pull/395/commits/962742c999327a17fbf4f6398d17afd4a9e73f90 fine with you?

jgmartin:
Perfect ty @jannic

jannic:
@jgmartin unfortunately, this branch needs an update: `embedded_time` was replaced by `fugit`. Do you want to update the example? I can do it as well, if you like. But I don't have a pimoroni display pack to test the result.

boards/rp-pico/examples/pico_display_pack_2.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    // Configure button inputs
+    // 12 - A button
+    let _a_pin = pins.gpio12.into_floating_input();
+    // 13 - B button
+    let _b_pin = pins.gpio13.into_floating_input();
+    // 14 - X button
+    let _x_pin = pins.gpio14.into_floating_input();
+    // 15 - Y button
+    let _y_pin = pins.gpio15.into_floating_input();
```
9names, line 106: These switches do not have external pull-ups, you should enable the internal ones or you'll never be able to detect a button press.
```suggestion
    // Configure button inputs
    // The push-buttons connect their respective pin to ground, this is often referred to as "active-low"
    // Since the board does not have external pull-up resistors, enable internal pull-up on each input
    // 12 - A button
    let _a_pin = pins.gpio12.into_pull_up_input();
    // 13 - B button
    let _b_pin = pins.gpio13.into_pull_up_input();
    // 14 - X button
    let _x_pin = pins.gpio14.into_pull_up_input();
    // 15 - Y button
    let _y_pin = pins.gpio15.into_pull_up_input();
```

9names:
This example is also compatible with Pimoroni Display Pack (1.0) as well - the pinout for LED, buttons and LCD controller for both is the same.
The LCD controller handles the differences between panels so the circle still renders correctly on the smaller screen even if you leave the screen size set to 320x240 (instead of actual screen resolution of 240x135).

jgmartin:
> @jgmartin unfortunately, this branch needs an update: `embedded_time` was replaced by `fugit`. Do you want to update the example? I can do it as well, if you like. But I don't have a pimoroni display pack to test the result.

No problem I will do this and test it.

boards/rp-pico/examples/pico_display_pack_2.rs
```diff
@@ -0,0 +1,170 @@
...
+    // 20 - LCD backlight
+    pins.gpio20.into_push_pull_output().set_high().unwrap();
+
+    // 22 - LCD hard reset
+    let mut rst_pin = pins.gpio22.into_push_pull_output();
+    rst_pin.set_low().unwrap();
+
+    // Setup SPI
+    let spi_cs = pins.gpio17.into_push_pull_output();
+    let spi_miso = pins.gpio16.into_push_pull_output();
```
ithinuel, line 133: This pin is not used as miso so I'd suggest giving it an name matching its use. eg: `di_data_command`

boards/rp-pico/examples/pico_display_pack_2.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    // Configure button inputs
+    // 12 - A button
+    let _a_pin = pins.gpio12.into_pull_up_input();
+    // 13 - B button
+    let _b_pin = pins.gpio13.into_pull_up_input();
+    // 14 - X button
+    let _x_pin = pins.gpio14.into_pull_up_input();
+    // 15 - Y button
+    let _y_pin = pins.gpio15.into_pull_up_input();
```
ithinuel, line 106: Those pins are not used (and are not configured for the purpose of another peripheral) so I think they should be removed from the example.

Alternatively, the example could draw on the display which button is pressed (as a letter or any other symbol).

boards/rp-pico/examples/pico_display_pack_2.rs
```diff
@@ -0,0 +1,170 @@
...
+use embedded_graphics::prelude::*;
+use embedded_graphics::primitives::*;
+
+// GPIO traits
+use embedded_hal::digital::v2::OutputPin;
+use embedded_hal::PwmPin;
+
+/// External high-speed crystal on the Raspberry Pi Pico board is 12 MHz. Adjust
+/// if your board has a different frequency
+const XTAL_FREQ_HZ: u32 = 12_000_000u32;
```
ithinuel, line 50: The definition provided by the `rp_pico` crate should be used instead.

ithinuel (changes requested):

ithinuel:
This PR should be moved to https://github.com/rp-rs/rp-hal-boards.
I am closing it here but be assured these changes would be welcome over there :) .

PR #398, opened 2022-07-25
Update examples to defmt 0.3

jannic:
Current version of probe-run is 0.3.3, which uses defmt 0.3.2.

With a firmware using defmt 0.2, this causes the following error
message:

```
Error: defmt wire format version mismatch: firmware is using 0.2, `probe-run` supports 3
suggestion: `cargo install` a different version of `probe-run` that supports defmt 0.2
```

Therefore, upgrade defmt dependency, ~~and also fix the linker script
in .cargo/config~~.

**EDIT:** instead of .cargo/config, use build.rs in individual board crates to set linker options as necessary.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -31,7 +31,6 @@ rustflags = [
     "-C", "link-arg=-Tlink.x",
     "-C", "inline-threshold=5",
     "-C", "no-vectorize-loops",
-    "-C", "link-arg=-Tdefmt.x",
 ]
 
 # This runner will make a UF2 file and then copy it to a mounted RP2040 in USB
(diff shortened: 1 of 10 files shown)


jannic:
Doesn't work yet, and I don't have a good idea how to fix this.
Either examples not using defmt will fail to build because defmt.x is not found, or examples using defmt will fail to build because defmt.x is not used.

9names:
I tried to do this too, same problem.
Maybe we should think about integrating logging functionality into the HAL or BSPs (behind a feature)?
Would be nice to have the examples "just work" with defmt or usb-uart output.

jannic:
Not sure. On the one hand, yes, it would be nice to have simple examples. On the other hand, I want examples to show typical usage and not something which would be done different in "real" code.

9names (approved):
Lgtm

PR #399, opened 2022-07-26
Fix handling of duty cycle while PWM channels is disabled

### LINKED ISSUE #390: PWM enable function resets the duty cycle
The function [`set_duty`](https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/src/pwm/mod.rs#L639) of a pwm channel directly writes the desired duty value into the respective peripheral register. However, [`enable`](https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/src/pwm/mod.rs#L628) writes the value `self.duty_cycle` in that register. That value, however, is only set by the `disable` function. 

So the following code actually disables the PWM:

```rust
let pwm_slices = rp2040_hal::pwm::Slices::new(pac.PWM, &mut pac.RESETS);
let mut pwm = pwm_slices.pwm2;
// Set prescalers for 50Hz Pwm
pwm.set_div_int(40u8);
pwm.set_div_frac(0u8);
pwm.set_top(62499);
pwm.enable();
let mut channel = pwm.channel_a;
channel.set_duty(6000); // <----
channel.output_to(pins.gpio4);
channel.enable(); // <----
```
I assume that pattern is not uncommon in code using `PWMPin`, so I'd suggest to update `self.duty_cycle` in the `set_duty` fn as well.

jannic:
Fixes #390

### DIFF
--- a/rp2040-hal/src/pwm/mod.rs
+++ b/rp2040-hal/src/pwm/mod.rs
@@ -572,6 +572,7 @@ pub struct Channel<S: SliceId, M: SliceMode, C: ChannelId> {
     slice_mode: PhantomData<M>,
     channel_id: PhantomData<C>,
     duty_cycle: u16,
+    enabled: bool,
 }
 
 impl<S: SliceId, M: SliceMode, C: ChannelId> Channel<S, M, C> {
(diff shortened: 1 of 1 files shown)


9names (approved):
LGTM

PR #402, opened 2022-07-27
Remove unnecessary cortex_m::interrupt::free in timer.rs

jannic:
The only thing accessed by those methods is `&mut self`, which is
guaranteed to be a unique reference. So an interrupt can not
interfere with the correctness of the operation.


9names (approved):
LGTM

PR #405, opened 2022-07-30
Fix usb missed ep0 out packet and add recommended sync delays

ithinuel:
The ep0-out buffer must not be marked as available unless required.
Otherwise, the controller will acknowledge the data-out packet but won't reflect that in its status registers.

This patch forces the controller to nack the data-out phase until we have processed the setup packet.

As per 4.1.2.5.1, the access to the DPSRAM should "be considered asynchronous and not atomic".
It is recommended to write to buffer control register in two steps.
- A first one to configure all bits but Available.
- Wait clk_sys/clk_usb (typically 125MHz/48MHz).
- Then set the available bit (if required).

The pico-sdk uses a blanket 12tick asm block so proposing a 12tick delay here too.

### DIFF
--- a/rp2040-hal/src/usb.rs
+++ b/rp2040-hal/src/usb.rs
@@ -352,8 +352,12 @@ impl Inner {
                 w.full_0().clear_bit();
                 w.pid_0().bit(!r.pid_0().bit())
             });
-            cortex_m::asm::delay(12);
-            buf_control.modify(|_, w| w.available_0().set_bit());
+            if index != 0 || len == ep.max_packet_size.into() {
+                // only mark as available on the control endpoint if and only if the packet was
+                // max_packet_size
+                cortex_m::asm::delay(12);
+                buf_control.modify(|_, w| w.available_0().set_bit());
+            }
             Ok(len)
         }
     }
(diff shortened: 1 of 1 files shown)


jannic:
> The pico-sdk uses a blanket 12tick asm block so proposing a 12tick delay here too.

This is a rather huge safety margin:

`clk_sys/clk_usb` should be at most 3 for supported frequencies (clk_usb is fixed at 48MHz and max for clk_sys is 133MHz).
A 12 tick delay would only be necessary for overclocking above 500MHz, far out of spec (but still in a range shown to be feasible on the rp2040).

And then, `cortex_m::asm::delay` currently waits for a larger number of clock cycles, to be safe: It doesn't know if it's running on some more advanced CPU which could compute the delay loop faster, so it takes the most conservative guess. If I count correctly, the `delay(12)` will actually wait for 20 clock cycles.
So the delay is wasting at least 17 clock cycles (when clk_sys is running at default speed).

But I guess this is code which would not run in a timing critical path? And even at a slower clk_sys of 48MHz (where no additional waits would be needed), it's only wasting about 0.4µs.

Conclusion: LGTM.

jannic (approved):

PR #406, opened 2022-07-30
Add suspend, resume and remote wakeup support.

ithinuel:
Every thing is in the title :) 

Although, it's worth noting that due to a bug in the currently published version of usb-device (0.2.8) this won't work properly.
[A fix](https://github.com/rust-embedded-community/usb-device/pull/97) has been merged already so until it's published on crates.io one has to 
```toml
[patch.crates-io]
usb-device = { git="https://github.com/rust-embedded-community/usb-device" }
```

### DIFF
--- a/rp2040-hal/src/usb.rs
+++ b/rp2040-hal/src/usb.rs
@@ -402,6 +402,14 @@ impl UsbBus {
             inner: Mutex::new(RefCell::new(Inner::new(ctrl_reg, ctrl_dpram))),
         }
     }
+
+    /// Generates a resume request on the bus.
+    pub fn remote_wakeup(&self) {
+        interrupt::free(|cs| {
+            let inner = self.inner.borrow(cs).borrow_mut();
+            inner.ctrl_reg.sie_ctrl.modify(|_, w| w.resume().set_bit());
+        });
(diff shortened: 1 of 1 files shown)


9names (approved):
Lgtm.
Will probably want to move all interrupt::free to critical-section::with at some point (conditionally) as Jannic has done, but in practice I don't think it's sensible to call the USB driver from 2 cores anyway

9names:
On second thought, adding the atomic aliases and removing the critical sections altogether is a better choice.

jannic:
What's the point of those suspend/resume operations? Isn't it meant to be some kind of power saving operation?
If so, shouldn't the firmware actually somehow change it's behaviour when in suspend more? So there should be some way for the driver to signal that to the rest of the firmware.
Or am I misunderstanding what this suspend mode is trying to achieve and it's all about suspending the USB communication itself?

ithinuel:
The usb-device stack has a [`fn state(&self) -> UsbDeviceState`](https://docs.rs/usb-device/latest/usb_device/device/struct.UsbDevice.html#method.state) member that the application should monitor for changes to and from suspended.

PR #409, opened 2022-08-02
Gate usb Suspend/Resume feature with a feature: usb-suspend

ithinuel:
Because this is broken in the current version of usb-device, we need to gate this until the upstream fix is published in a version.

**Note:**

I didn't get to test it yet. It'd be greatly appreciated if reviewer could build one of the example and confirm that it is broken without the gate (you'll need to make the usb host suspend the link, typically by putting a laptop host into sleep/suspend).

### DIFF
--- a/rp2040-hal/src/usb.rs
+++ b/rp2040-hal/src/usb.rs
@@ -564,7 +564,10 @@ impl UsbBusTrait for UsbBus {
             let sie_status = inner.ctrl_reg.sie_status.read();
             if sie_status.bus_reset().bit_is_set() {
                 return PollResult::Reset;
-            } else if sie_status.suspended().bit_is_set() {
+            }
+
+            #[cfg(feature = "usb-suspend")]
+            if sie_status.suspended().bit_is_set() {
                 return PollResult::Suspend;
             } else if sie_status.resume().bit_is_set() {
                 return PollResult::Resume;
--- a/rp2040-hal/src/usb.rs
+++ b/rp2040-hal/src/usb.rs
@@ -447,13 +447,15 @@ impl UsbBusTrait for UsbBus {
             // and when a setup packet is received
             // this should be sufficient for device mode, will need more for host.
             inner.ctrl_reg.inte.modify(|_, w| {
+                #[cfg(feature = "usb-suspend")]
                 w.buff_status()
-                    .set_bit()
-                    .bus_reset()
-                    .set_bit()
                     .dev_resume_from_host()
                     .set_bit()
                     .dev_suspend()
+                    .set_bit();
+                w.buff_status()
+                    .set_bit()
+                    .bus_reset()
                     .set_bit()
                     .setup_req()
                     .set_bit()
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -45,6 +45,7 @@ rt = ["rp2040-pac/rt"]
 rom-func-cache = []
 disable-intrinsics = []
 rom-v2-intrinsics = []
+usb-suspend = []
 
 [[example]]
 # irq example uses cortex-m-rt::interrupt, need rt feature for that


9names:
> confirm that it is broken without the gate

How do we confirm that it is broken, would you expect it to not enumerate on resume?
I've tested it this way (with and without the feature) and it seems fine.
I also tested it with XHCI HSETT, running get descriptor, suspend, get descriptor, resume, get descriptor.
Responds identically in both cases - before suspend descriptor fine, after suspend error, after resume fine.

jannic:
I was unable to see any difference in behavior between usb-device 0.2.8 and the current git version of usb-device.
Monitoring `[fn state(&self) -> UsbDeviceState](https://docs.rs/usb-device/latest/usb_device/device/struct.UsbDevice.html#method.state)` only returned `Default`, `Addressed` and `Configured`, never `Suspend`.

I tried it with both a windows and a linux noteboot. On windows, it didn't work at all because windows just powered down the USB port when suspending the notebook. On linux, I saw no state change on suspend and a sequence of `Default`, `Addressed` and `Configured` on resume, but it didn't matter which version of usb-device I was using.

ithinuel:
The way I did test suspend was:
- connect the device usb to a laptop
- connect the debug probe to my desktop
- probe-run the firmware showing the defmt traces (especially on [`usb_dev.state()`](https://docs.rs/usb-device/latest/usb_device/device/struct.UsbDevice.html#method.state) changes).
- make the laptop enter suspend
- monitor the usb dev state move to [`Suspend`](https://docs.rs/usb-device/latest/usb_device/device/enum.UsbDeviceState.html#variant.Suspend).
- wake up the laptop
- monitor the usb dev state.

So with the feature:
- disabled, you should not see the device enter [`Suspend`](https://docs.rs/usb-device/latest/usb_device/device/enum.UsbDeviceState.html#variant.Suspend) (as it won't be reported to the usb stack).
- enabled with :
  - usb-device=0.2.8 it should wake up into [`Default`](https://docs.rs/usb-device/latest/usb_device/device/enum.UsbDeviceState.html#variant.Default) state (that's the broken behaviour).
  - usb-device=git it should resume to the previous state it was in before entering [`Suspend`](https://docs.rs/usb-device/latest/usb_device/device/enum.UsbDeviceState.html#variant.Suspend).

jannic:
Yes, that's exactly what I did, but I never saw the usb dev state move to `Suspend`.

ithinuel:
I also used a logic analyser to monitor the bus to make sure the state reported by the drivers (both on the device and dmesg on linux) where consistent with what happened physically on the bus.

jannic:
With this patch, and the feature not activated, the pico_usb_twitchy_mouse just hangs.
It looks like the usb interrupt fires continuously. Probably because it's activated by setting the INTE.DEV_SUSPEND bit and then never cleared, because `suspend()` won't get called.

ithinuel:
Ah that is a good point. Let me update that PR quickly.

jannic:
BTW, do you know if https://github.com/rust-embedded-community/usb-device/issues/96 has any practical impact? What use-cases are broken by that bug?

jannic:
Close :-)
```
error[E0599]: no method named `dev_resume_from_host` found for struct `pac::usbctrl_regs::inte::BUFF_STATUS_W` in the current scope
   --> rp2040-hal/src/usb.rs:443:22
    |
443 |                     .dev_resume_from_host()
    |                      ^^^^^^^^^^^^^^^^^^^^ method not found in `pac::usbctrl_regs::inte::BUFF_STATUS_W<'_>`
```

rp2040-hal/src/usb.rs
```diff
@@ -447,13 +447,15 @@ impl UsbBusTrait for UsbBus {
             // and when a setup packet is received
             // this should be sufficient for device mode, will need more for host.
             inner.ctrl_reg.inte.modify(|_, w| {
+                #[cfg(feature = "usb-suspend")]
                 w.buff_status()
-                    .set_bit()
-                    .bus_reset()
-                    .set_bit()
                     .dev_resume_from_host()
```
jannic, line 452:
```suggestion
                    .set_bit()
                    .dev_resume_from_host()
```

jannic:
After adding the missing `.set_bit()`, the patch works as expected, in the following combinations:
- usb-device 0.2.8, feature `usb-suspend` not enabled → `Suspend` state not visible from firmware
- usb-device 0.2.8, feature `usb-suspend` enabled → `Suspend` state visible from firmware, wrong state after resume
- usb-device from git, feature `usb-suspend` not enabled → `Suspend` state not visible from firmware
- usb-device from git, feature `usb-suspend` enabled → `Suspend` state visible from firmware, correct state after resume

Even with the wrong state after resume, I didn't notice any ill effects. Maybe the linux kernel just works around the non-compliant behavior?

ithinuel:
closing in favour of https://github.com/rp-rs/rp-hal/pull/410

PR #411, opened 2022-08-02
Fix clearing usb sie_status flags

jannic:
Without those changes I was able to get `pico_usb_twitchy_mouse` into a state where the interrupt was triggered continuously and the main routine didn't make any progress.

### DIFF
--- a/rp2040-hal/src/usb.rs
+++ b/rp2040-hal/src/usb.rs
@@ -544,7 +544,7 @@ impl UsbBusTrait for UsbBus {
             inner
                 .ctrl_reg
                 .sie_status
-                .modify(|_, w| w.suspended().set_bit());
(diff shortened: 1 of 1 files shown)


ithinuel (approved):
Indeed, it appears that there not a 1:1 match of calls to `suspend` and `resume` and returns of PollResult::Suspend/Resumt

For example [here](https://github.com/rust-embedded-community/usb-device/blob/eaf9709b90e505e1451f74ae9b9c4badc329aba8/src/device.rs#L167-L171) although the [docs](https://github.com/rust-embedded-community/usb-device/blob/eaf9709b90e505e1451f74ae9b9c4badc329aba8/src/bus.rs#L109-L110) says
> This will be called after `poll` returns `PollResult::Suspend`

PR #413, opened 2022-08-07
rom_funcs results output over USB instead of UART

camrbuss:
- Added f64 tests for chips with rom_version>1
- Output results over USB Serial instead of hardware UART

### DIFF
--- a/rp2040-hal/examples/rom_funcs.rs
+++ b/rp2040-hal/examples/rom_funcs.rs
@@ -2,7 +2,10 @@
 //!
 //! This application demonstrates how to call functions in the RP2040's boot ROM.
 //!
-//! It may need to be adapted to your particular board layout and/or pin assignment.
+//! It may need to be adapted to your particular board layout and/or pin assignment,
+//! but only outputs via USB Serial port.
+//!
+//! Send characters to get results or send `!` to put the rp2040 in usb boot
 //!
 //! See the `Cargo.toml` file for Copyright and license details.
 
@@ -25,7 +28,12 @@ use hal::pac;
 
 // Some traits we need
 use core::fmt::Write;
-use hal::Clock;
+
+// USB Device support
+use usb_device::{class_prelude::*, prelude::*};
+
+// USB Communications Class Device support
+use usbd_serial::SerialPort;
 
 /// The linker will place this boot block at the start of our program image. We
 /// need this to help the ROM bootloader get our code up and running.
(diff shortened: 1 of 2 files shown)


9names:
I like the idea, but just replacing the UART output with USB-UART output doesn't really improve the situation.
A better option would be to have a optional `logger` feature in the HAL where you can choose between uart, usb-cdc, rtt or defmt for getting textual output from your program.
That way all the examples can output data in the user's preferred format, and end-users can also opt in to get a simpler debug output.
Also: make it a separate PR to updating the example.

rp2040-hal/examples/rom_funcs.rs
```diff
@@ -41,6 +49,26 @@ const XTAL_FREQ_HZ: u32 = 12_000_000u32;
 /// to work, this value must be of the form `2**N - 1`.
 const SYSTICK_RELOAD: u32 = 0x00FF_FFFF;
 
+struct Buffer<'a> {
+    buf: &'a mut [u8],
+    index: usize,
+}
+
+// write_str from https://stackoverflow.com/a/39491059
```
jannic, line 57: As far as I know, the default license of subscriber content published on stack overflow doesn't allow redistribution under MIT / Apache-2 license.

camrbuss: Oh my, I didn't realize that. How should I rectify the situation?

jannic: Perhaps just use https://docs.rs/heapless/0.7.16/heapless/struct.Vec.html or https://docs.rs/heapless/0.7.16/heapless/struct.String.html instead of implementing your own buffer?

9names: This might be a dumb question, but why are we buffering anyway?
Can't you push this straight into the SerialPort?

camrbuss: I don't believe [usbd_serial](https://docs.rs/usbd-serial/latest/usbd_serial/struct.SerialPort.html#impl-Write%3Cu8%3E) implements `write_str`

Is there another way that I am not seeing?

9names: I was assuming it wouldn't be difficult to build a wrapper struct for usbd_serial and implement write_str on that, but I think the generics would be a pain.
Maybe it would be easier to make a PR to get usbd_serial to (conditionally) impl write_str

9names: Looking at https://github.com/rust-embedded-community/usbd-serial/pull/16, seems this has already been proposed and declined since the blocking call would have to poll the USB device (which it can't).
So: would really require having some logger/stdin+stdout functionality, which would possibly also be part of how we could resolve #456

PR #414, opened 2022-08-10
Upgrade to critical-section 1.0.0

jannic:


9names (approved):
LGTM.
Should wait until #120 is merged so the USB changes don't conflict with that.

ithinuel (commented):
I see a feature critical-section-impl but disabling it does not (seem to) change much thing.
I see what the feature does in the code but I don't know what's the impact of it (without having to learn about the critical-section crate).
Is worth documenting this, either in the module level (although it is not a public module) or as a oneline comment in Cargo.toml?

jannic:
In general, features should add functionality, not remove them. (https://doc.rust-lang.org/cargo/reference/features.html#feature-unification)
I wonder if we could argue that "no-default-critical-section-impl" (probably with a different name) in fact adds functionality, namely the ability to provide a custom critical section impl? We could even name this feature "custom-impl", in perfect analogy to the same feature in critical-section 0.2.x.

ithinuel:
How about `cs-allow-alternative-locks` ?

We could even have a `cs-disable-interrupt-only` although I wouldn't be surprised if a dedicated crate already offered that :D

ExplodingWaffle:
[critical-section's docs](https://docs.rs/critical-section/latest/critical_section/#providing-an-implementation) seem to imply that the feature shouldn't be enabled by default which I think makes sense so you can use a different impl (like cortex-m's, in a single core app). cortex-m doesn't use a default feature either.

jannic:
I suggests to gate the implementation behind a feature, which we do. I don't think that gives us advice on whether that feature should be enabled by default. rp2040-hal is only used for one chip, which is a multi-core one, so using a multi-core-safe critical section by default seems reasonable.

If it was easy to disable a default feature, I'd be convinced that enabling it by default would be the right thing to do. But unfortunately, as @ithinuel mentioned, disabling a default feature on a crate which is also used as a transitive dependency by other crates is not easily possible.

But without enabling the implementation by default, we would have a default which would not work without additional work by the user of the lib. Also not very user-friendly.

So providing the impl by default, and having a feature that can be set if the caller want's to provide its own custom implementation could be a good workaround. Would work by default (and the default impl would be the safe one), but on demand, a different impl could be provided. If one crate enables this feature and provides its own custom impl, it works as expected. If two different crates try this at the same time, one would get a conflict, which is probably the best we could do in such a situation. Overall, I think this is a good solution.

ExplodingWaffle:
Forgive my lack of cargo knowledge- but wouldn't a `no-critical-section-impl` feature still have the transitive dependancy issue, but for libraries- so it'd have to be enabled in all library crates in order for things to work? I guess that could work, but I think it comes with the same "user friendly" problem, but for library authors, and instead of erroring when you try to use critical-section in your application, there's no error until the user tries to use c-s and there's conflicting impls. [see here](https://docs.rs/critical-section/latest/critical_section/index.html#usage-in-libraries)- but no library author is going to be reading those docs if they aren't using c-s. Also not sure what would happen if libraries had different versions of the hal- [this](https://docs.rs/critical-section/latest/critical_section/#duplicate-symbol-errors) seems to suggest there could be problems.
(cc @Dirbaio)

Dirbaio:
Yeah the problem with default features is that, it's too easy for one random lib to forget `default-features = false` and then make it impossible for the end user to disable the default feature. I guess you can add `default-features=false` to all crates and then hope (or send tons of PRs) that other authors don't forget it, but it's not quite nice...

In cortex-m 0.7.6, `critical-section-single-core` is not a default feature for this reason.

I sort of lean towards making it non-default. The case in rp-hal is less clear than in cortex-m though. In cortex-m it's somewhat likely a user might want to use another impl (using a multicore chip, using nrf softdevice, using a rtos..) but if you're using rp-hal it's because you're on a rp2040 so *the* correct critical section impl is rp-hal's. Can't think of many reasons you wouldn't want it. Perhaps if you're only using one core and want the faster perf?

jannic:
The `no-critical-section-impl` feature would not have the same issues because features are additive: as soon as at least one crate depending on rp2040-hal activates that feature, the hal would stop providing the impl. The exact property of feature resolution that bites us with the default features is actually useful here.

ithinuel:
I had a go at trying to make a feature that'd allow alternate custom cs impl but couldn't find a way to enable a feature on a dependency **if** a feature on the current crate is **not** enabled.

`critical-section/restore-state-u8` is required by the impl but needs to go to allow custom impl to enable their own state storage.

The only way forward I can see is that we need the critical section enabled manually like we do with `rt` and like the cortex-m crate does with `critical-section-single-core`. (even if that makes sense for cortex-m not to enable it by default).

jannic:
I opened [this issue](https://github.com/rust-embedded/critical-section/issues/25) to discuss the possible approaches of optionally providing a critical-section implementation.

jannic:
@ithinuel, this could be an option, basically as you suggested, but with the feature enabled by default from the board crates:
https://github.com/jannic/rp-hal/commit/280de7cb99be87f291b7b07c1c52cab8230a1a61

A rationale is given in the commit message.

ithinuel:
This commit seems perfect to me, I'd be happy to merge as it is.

ithinuel:
https://github.com/rp-rs/rp-hal/pull/444 opened with your suggestion :) (I didn't one could open a pr from someone else's repo :D)

PR #417, opened 2022-08-11
Fix clippy warnings

jannic:
Fixes a few instances of derive-partial-eq-without-eq and needless-borrow
clippy warnings.


jannic:
derive-partial-eq-without-eq was added in rust 1.63.0 and breaks the CI tests: 
https://github.com/rp-rs/rp-hal/runs/7796190750?check_suite_focus=true

9names (approved):

PR #418, opened 2022-08-12
Add meaningful names to steps and remove quasi-repetition

ithinuel:
The workflow only shows "Runs action-rs/cargo@v1" which does not tell us much about what's actually being done.
The list and clean up done after each build varies by only the debug/release part of the path that is repeated several times (2).

This PR gives each step a more meaningful name and simplify the list & clean up step with a generic one liner that can be copy/pasted without modification.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1,9 +1,13 @@
 on: [push, pull_request]
 name: Build and Test check
 jobs:
-  check:
-    name: cargo-check
+  builds:
+    name: "Build checks"
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


9names (approved):
:tada:

ithinuel:
Interestingly, git does not detect conflict between this PR and #120 and yield garbage result, so whichever comes first, the other will need to be updated.

9names (approved):
Oh is that what matrix does? I like it.

thejpster:
You cannot be told what the Matrix is. You have to see it for yourself.

thejpster (approved):

PR #421, opened 2022-08-18
Enable rp2040-hal/defmt feature in dev-dependencies

### LINKED ISSUE #420: probe-run doesn't like examples without defmt
Since the upgrade to defmt 0.3, `probe-run` doesn't like the examples which don't actually use defmt:
```
cargo run --example pico_blinky
   Compiling rp-pico v0.4.1 (/home/jan/rp2040/rp-rs/rp-hal/boards/rp-pico)
    Finished dev [unoptimized + debuginfo] target(s) in 0.26s
     Running `probe-run --chip RP2040 target/thumbv6m-none-eabi/debug/examples/pico_blinky`
Error: `.defmt` section found, but no version symbol - check your linker configuration
```

This is not a compiler error, but only happens when trying to run the resulting binary with `probe-run`. (Which is why I didn't notice it when testing https://github.com/rp-rs/rp-hal/pull/398 :disappointed:  )

This can be fixed by adding `use defmt_rtt as _;` to each example. Not nice, but would work. Any better idea?

jannic:
This is a little bit hacky, as it relies on rp2040-hal actually
using (and therefore linking) some defmt code when the defmt feature is
enabled.

However this solution has the advantage that it only affects
dev-dependencies and therefore has no impact on the board crates
themselves.

Closes #420

### DIFF
--- a/boards/pimoroni-plasma-2040/Cargo.toml
+++ b/boards/pimoroni-plasma-2040/Cargo.toml
@@ -18,6 +18,7 @@ cortex-m-rt = { version = "0.7", optional = true }
 embedded-time = "0.12.0"
 
 [dev-dependencies]
+rp2040-hal = { path = "../../rp2040-hal", version = "0.5.0", features = [ "defmt" ] }
 panic-halt= "0.2.0"
 embedded-hal ="0.2.5"
 smart-leds = "0.3.0"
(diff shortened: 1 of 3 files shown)


ithinuel (changes requested):
I would rather suggest to change the `build.rs` to make the addition of `defmt.x` gated by the feature.

```rust 

fn main() {
    if cfg!(feature = "defmt") {
        println!("cargo:rustc-link-arg-examples=-Tdefmt.x");
    }
}
```

ithinuel (approved):
Never mind, the better solution would be to only add defmt.x for examples that need it.
But it isn't possible (yet) with current cargo: https://github.com/rust-lang/cargo/issues/1430

jannic:
I agree that this is only a band-aid and we should keep looking for a better way.
Fortunately, as it only changes dev-dependencies, a better fix can just revert this change without affecting other crates depending on this one.

PR #422, opened 2022-08-18
Replace cortex_m::interrupt::Mutex with critical_section::Mutex in examples

ithinuel:
For consistency let's use `critical_section` everywhere.

~(There was also a `cortex_m::interrupt::free` left over in `timer.rs`)~

### DIFF
--- a/boards/rp-pico/CHANGELOG.md
+++ b/boards/rp-pico/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `rp2040-e5` feature enabling the workaround for errata 5 on the USB device peripheral.
+- Support for critical-section 1.0.0 in the examples.
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


rp2040-hal/src/timer.rs
```diff
@@ -268,7 +268,7 @@ macro_rules! impl_alarm {
                 if duration < MIN_MICROSECONDS {
                     return Err(ScheduleAlarmError::AlarmTooSoon);
                 } else {
-                    cortex_m::interrupt::free(|_| {
+                    critical_section::with(|_| {
```
jannic, line 271: I don't think this is a good change. The only reason interrupts are disabled here is that the timing is critical. It's not done to avoid concurrent accesses.
Changing this to a critical section only increases the overhead.
(In general, a global critical sections and disabling interrupts can serve different purposes. It may have been a mistake to return a CriticalSection token in both cases. But on single core, it's the same operation, so it's understandable.)

jannic (approved):
While technically `interrupt::free` is sufficient in those examples (they don't use the second core), it's probably a better example to use the more general solution. :+1:

PR #423, opened 2022-08-19
Use rp2040-hal in all example (possibly through their bsp)

ithinuel:
Some of the examples were using the cortex_m_rt::entry method which
misses the device specific spinlock re-initialisation.

This commits makes the usage more consistent by using rp2040_hal exported
macro as the only `entry` method used across examples.


jannic:
`rp2040-hal/Cargo.toml.orig` was accidentally added to this merge request.

boards/arduino_nano_connect/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Changelog
```
jannic, line 1: I'm not sure if changelogs for individual board crates are really useful. Updating them will be a lot of work, and in most cases they will all contain the same entries. IIRC there was a discussion about this in the past, but I don't remember where and what the conclusion was.

ithinuel:
With this extra commit, it should no longer happen

jannic (changes requested):
I like the idea to use `rp2040_hal::entry` in the examples. They provide a better default.

jannic (approved):
The previous "request changes" was due to the `.orig` files, which you already removed in the meantime.
Comments regardings changelogs and the way the entry macro is called are just ideas, not blockers.

PR #424, opened 2022-08-20
Experiment: Remove defmt

jannic:
This is not (yet?) meant for merging but just an experiment to see how using rtt-target instead of defmt would look like.
Removal of the defmt specific link flags simplifies the examples quite a bit. I think we should at least consider such a change.

Additionally, not requiring defmt would make the log messages usable for developers with an existing setup which supports rtt, but not defmt. (ie. many setups from the C world, without rust specific tooling). This was what initially triggered the discussion on defmt vs. rtt-target on the matrix channel.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,3 +28,4 @@ rp2040-hal = { path ="./rp2040-hal" }
 
 [patch.crates-io]
 rp2040-hal = { path ="./rp2040-hal" }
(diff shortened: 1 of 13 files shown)


jannic:
The code is in rp-hal-boards now, and the question if we want to replace defmt is still open. However, as nobody commented on the merge request, I guess there is no demand for such a change?

jannic:
Related: https://github.com/knurling-rs/defmt/issues/749
Defmt sometimes is another thing that can go wrong, which can become really confusing to beginners, when our examples just don't work because of some dependency issue.

jannic:
Closing this for now.

PR #425, opened 2022-08-21
Fix clocks docs "extended usage" to match `init_clocks_and_plls`

camrbuss:
The current extended usage clocks example is written correctly but the XOSC needs to be setup first in order for it to work properly. This matches the layout in `init_clocks_and_plls`.


jannic (approved):
The failing clippy check is unrelated. I got the same error on a completely different change. Probably caused by a new check added in a recent version of clippy.

jannic:
> The failing clippy check is unrelated. I got the same error on a completely different change. Probably caused by a new check added in a recent version of clippy.

It's an actual error, caused by a change in the `embedded-sdmmc` dependency. Still not related to this pull request. I'll care about that in a separate PR.

Thanks for contributing!

PR #426, opened 2022-08-21
Fix build of pico_spi_sd_card

jannic:
The build failure was triggered by this commit:
https://github.com/rust-embedded-community/embedded-sdmmc-rs/commit/566f2023653dace195a43b05a6429f47e6be8a50

@ostenning, does this fix look reasonable? Was it intended that all callers of `Controller::new` would need to add those type parameters? (I'm not complaining about the breaking change. We are using a git dependency, therefore we have to live with breaking changes.)

### DIFF
--- a/boards/rp-pico/examples/pico_spi_sd_card.rs
+++ b/boards/rp-pico/examples/pico_spi_sd_card.rs
@@ -273,7 +273,7 @@ fn main() -> ! {
     blink_signals(&mut led_pin, &mut delay, &BLINK_OK_LONG);
 
     info!("Init SD card controller...");
-    let mut cont = Controller::new(block, DummyTimesource::default());
+    let mut cont: Controller<_, _, 4, 4> = Controller::new(block, DummyTimesource::default());
 
     blink_signals(&mut led_pin, &mut delay, &BLINK_OK_LONG);
 


boards/rp-pico/examples/pico_spi_sd_card.rs
```diff
@@ -273,7 +273,7 @@ fn main() -> ! {
     blink_signals(&mut led_pin, &mut delay, &BLINK_OK_LONG);
 
     info!("Init SD card controller...");
-    let mut cont = Controller::new(block, DummyTimesource::default());
+    let mut cont: Controller<_, _, 4, 4> = Controller::new(block, DummyTimesource::default());
```
ithinuel, line 276: I don't really know why it is required since the consts have default value.

Surprisingly (to me) this is enough to make it compile without any complaint:

```suggestion
    let mut cont: Controller<_, _> = Controller::new(block, DummyTimesource::default());
```

jannic: I think this nicely explains why this happens: https://gankra.github.io/blah/defaults-affect-inference/#default-type-parameters

As I understand it: When declaring the type of `cont` with `let mut cont: Controller<_, _>`, rustc will use the default values.

But when calling `Controller::new( ... )`, the default type parameters are not used to constrain the possible values of the type parameters of `Controller`. All 4 of them need to be determined through type inference, and there is nothing in the existing code which restricts the two numeric parameters.

ithinuel: Thanks! Today I learn this!

jannic:
Made this a draft, for now. Let's wait for embedded-sdmmc 0.4.0, now that #428 fixed the builds.

jannic:
With embedded-sdmmc upgraded to 0.4.0 with https://github.com/rp-rs/rp-hal-boards/pull/5, this PR is obsolete.

PR #427, opened 2022-08-22
Make the i2c error render as text if eh1_0_alpha is on.

ithinuel:


rp2040-hal/src/i2c.rs
```diff
@@ -80,6 +83,38 @@ pub enum Error {
     AddressReserved(u16),
 }
 
+#[cfg(feature = "eh1_0_alpha")]
+impl core::fmt::Debug for Error {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
```
9names, line 88: Use the same name parameter name `fmt` for the formatter for both core::fmt::Formatter and defmt, no reason these should be different

9names (changes requested):

rp2040-hal/src/i2c.rs
ithinuel: The difference comes from how the parameter is called in the respective trait definitions. 
Any reason to chose `fmt` over `f` ?

9names: No reason, just would prefer that it's the same one for consistency. 
I don't believe in making your project conform to the internal naming conventions of libraries.
Use f for both if you prefer brevity, fmt for both would be fine, as would formatter.

ithinuel: opted for `fmt` as requested :)

9names (approved):
Lgtm

jannic (approved):

PR #429, opened 2022-08-23
remove unused dependencies & add GHA flow

ithinuel:
Should it be made part of the existing jobs or parallel ones?

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -2,7 +2,7 @@ on: [push, pull_request]
 name: Build and Test check
 jobs:
   builds:
-    name: "Build checks"
+    name: Build checks
     runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


ithinuel:
Either way, it make the checks much much longer, so I'm not sure it's entirely worth it to have it as workflow.
We might just be good having it run manually every now and then.

9names (approved):
Glad to drop all the unused deps.
Looks like all the extra time is just building cargo-udeps.
I couldn't manage to get toolchain caching working correctly, but it should be possible to do this once and cache it for subsequent runs...

ithinuel:
Seems like the `use-tool-cache` option is deprecated ( https://github.com/actions-rs/tool-cache/ is archived )
And support using `@action/cache` has been waiting for about 2years: https://github.com/actions-rs/core/pull/125 and https://github.com/actions-rs/install/pull/5

9names:
Ah. Maybe using a prebuilt version instead? There are binary releases on github (need to untar them though).
I'll check how fast that is...

9names:
takes 1.5 seconds on my machine.
wish i didn't have to encode the version number in the string, maybe we could regex it out somehow but it might be more cursed...

```system
wget -q -O - https://github.com/est31/cargo-udeps/releases/download/v0.1.32/cargo-udeps-v0.1.32-x86_64-unknown-linux-gnu.tar.gz | tar -xz && mv ./cargo-udeps-v0.1.32-x86_64-unknown-linux-gnu/cargo-udeps ~/.cargo/bin && rm ./cargo-udeps-v0.1.32-x86_64-unknown-linux-gnu -r
```

ithinuel:
I found https://github.com/aig787/cargo-udeps-action :D seems to work as expected

PR #432, opened 2022-08-25
Temporarily disable out_sticky in set_pindirs

jannic:
Otherwise, it will change unrelated pindirs as well, if sticky is set


9names (approved):
LGTM

PR #439, opened 2022-08-29
Use TimerInstant in Timer::GetCounter & add Alarm::schedule_at

ithinuel:


rp2040-hal/src/timer.rs
```diff
@@ -289,6 +311,45 @@ macro_rules! impl_alarm {
...
+            /// The RP2040 is unable to schedule an event taking place in more than
+            /// `u32::max_value()` microseconds.
+            ///
+            /// [enable_interrupt]: #method.enable_interrupt
+            fn schedule_at<const FREQ_HZ: u32>(
+                &mut self,
+                timestamp: TimerInstantU64<FREQ_HZ>,
+            ) -> Result<(), ScheduleAlarmError> {
+                const MIN_MICROSECONDS: u64 = 10;
+                let timestamp = timestamp.ticks();
```
jannic, line 332: Perhaps I don't understand the `fugit` API, but doesn't that only work if `FREQ_HZ == 1_000_000`?
If so, `timestamp` should be a `TimerInstantU64<1_000_000>`.

rp2040-hal/src/timer.rs
```diff
@@ -199,6 +205,22 @@ pub trait Alarm {
         countdown: Duration<u32, NOM, DENOM>,
     ) -> Result<(), ScheduleAlarmError>;
 
+    /// Schedule the alarm to be finished at the given timestamp. If [enable_interrupt] is
+    /// called, this will trigger interrupt whenever this timestamp is reached.
+    ///
+    /// The RP2040 has been observed to take a little while to schedule an alarm. For this
+    /// reason, this function requires the timestamp to be at least 10 microseconds in the
+    /// future.
```
jannic, line 213: This claim sounds a little bit dubious. Why should the rp2040 take microseconds to schedule an alarm?
Is there some example available which demonstrates this delay?

jannic (changes requested):

rp2040-hal/src/timer.rs
ithinuel: This is the same claim as on `::schedule`. I did not witness that myself.
It'd rather see the sequence in a time critical block:

- set timestamp
- check if armed
- check if counter is > timestamp
- check if still armed
- if yes => we set the alarm in the past => error

rp2040-hal/src/timer.rs
ithinuel: It should Provided that the watchdog was configured properly.
It'd also make sense to provide an alias to this fully defined type.
I'll update accordingly.

jannic: I think we can assume that the watchdog is properly configured at this point. Otherwise the whole API doesn't make much sense. get_counter also assumes a 1MHz tick.

rp2040-hal/src/timer.rs
```diff
@@ -289,6 +311,51 @@ macro_rules! impl_alarm {
...
+                }
+
+                cortex_m::interrupt::free(|_| {
+                    let alarm = &timer.$timer_alarm;
+
+                    // safety: This is the only code in the codebase that accesses memory address $timer_alarm
+                    alarm.write(|w| unsafe { w.bits(timestamp_low) });
+
+                    // If it is not set, it has already triggered.
+                    if (timer.armed.read().bits() & $armed_bit_mask) != 0 {
```
jannic, line 345: Is this line needed?
It could be seen as an optimization, as it avoids the call to `get_counter` in case the alarm already triggered. However, this seems rather unlikely, as it was only scheduled a few cycles earlier.

I think the check below is sufficient. If `now < timestamp`, we are sure that the alarm was scheduled in the future and everything is fine. If `now < timestamp`, it _might_ have been scheduled in the past (and will trigger in ~72 minutes), or it _might_ have been scheduled properly and already triggered in the mean time. Those two cases are properly separated by the `(timer.armed.read().bits() & $armed_bit_mask) != 0` comparison below.

ithinuel: Indeed, thanks!

rp2040-hal/src/timer.rs
jannic: The new code looks much more sensible, thanks!
→ Now the comment is outdated :smile:

rp2040-hal/src/timer.rs
```diff
@@ -289,6 +311,52 @@ macro_rules! impl_alarm {
...
             }
 
+            /// Schedule the alarm to be finished at the given timestamp. If [enable_interrupt] is
+            /// called, this will trigger interrupt `
+            #[doc = $int_name]
+            /// ` whenever this timestamp is reached.
+            ///
+            /// The RP2040 has been observed to take a little while to schedule an alarm. For this
+            /// reason, this function requires the timestamp to be at least 10 microseconds in the
+            /// future.
```
jannic, line 321: Outdated, as well.

ithinuel: I hope that it'll be all good now xD

jannic (approved):

PR #440, opened 2022-08-29
Add a method to allow setting the PIO's clock divisor without floats

ithinuel:


rp2040-hal/src/pio.rs
```diff
@@ -1769,10 +1765,14 @@ impl<P: PIOExt> PIOBuilder<P> {
...
     ///
-    /// The is based on the sys_clk. Set 1 for full speed. A clock divisor of `n` will cause the state machine to run 1
-    /// cycle every `n` clock cycles. For small values of `n`, a fractional divisor may introduce unacceptable jitter.
-    pub fn clock_divisor(mut self, divisor: f32) -> Self {
-        self.clock_divisor = divisor;
+    /// The clock is based on the `sys_clk`. Set 1 for full speed. A clock divisor of `n` will
+    /// cause the state machine to run 1 cycle every `n` clock cycles. For small values of `int`,
+    /// a fractional divisor may introduce unacceptable jitter.
+    /// A value of 0 for the `int`eger part requires a `frac`tional part of 0. This is interpreted
+    /// by the device as the integer 65536.
```
jannic, line 1772: What about adding an equation: `PIO clock = sys_clk / (int + frac / 256)`

ithinuel: I reworded the comment.
I left the `clock_divisor_fixed_point` method. I don't want to have the same issue that we had with the migration to fugit (synchronous updates of the crates). I'm open to alternative names if this one is too long.

jannic (approved):

jannic (approved):

PR #444, opened 2022-09-01
Enable critical-section-impl by default only from board crates, not from hal

ithinuel:
There are a lot of non-binary crates depending on rp2040-hal. That way, the default-features of rp2040-hal may be activated unintentionally through an indirect dependency. Therefore, a binary crate which wants to disable the `critical-section-impl` feature to provide its own one could have a hard time to do so.

In contrast, the board support crates are usually only used by top-level binary crates. So disabling the default features on those should usually just work.

Binary crates depending on rp2040-hal directly, which don't use any board support crate, might need to activate the feature manually. This is reasonable because those binary crates need to replicate some boilerplate from the board crates anyhow.

### DIFF
--- a/boards/adafruit-feather-rp2040/Cargo.toml
+++ b/boards/adafruit-feather-rp2040/Cargo.toml
@@ -25,6 +25,7 @@ smart-leds = "0.3.0"
 ws2812-pio = "0.3.0"
 
 [features]
-default = ["boot2", "rt"]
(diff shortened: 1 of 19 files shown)


9names (approved):
I don't like that folks that use rp2040-hal directly will now have to specify this feature - I know that quite a few people do this and it will making upgrading more painful.
But the alternatives and the status quo are worse.
LGTM.

PR #447, opened 2022-09-07
pio: Move interrupt related (en|dis)abling/forcing methods to the statemachine

ithinuel:
The SM knows its id using the type system so there's no need for checking it.

 This commit also adds a `PioIRQ` enum to select the output IRQ.

EDIT: For an example of how I use this: https://github.com/ithinuel/rp2040-async-i2c/blob/main/src/pio.rs

### DIFF
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -51,18 +51,6 @@ pub trait PIOExt:
         (
             PIO {
                 used_instruction_space: 0,
-                interrupts: [
-                    Interrupt {
-                        id: 0,
-                        block: self.deref(),
-                        _phantom: core::marker::PhantomData,
-                    },
-                    Interrupt {
-                        id: 1,
-                        block: self.deref(),
-                        _phantom: core::marker::PhantomData,
-                    },
-                ],
                 pio: self,
             },
             sm0,
@@ -80,7 +68,6 @@ impl PIOExt for PIO1 {}
 pub struct PIO<P: PIOExt> {
     used_instruction_space: u32, // bit for each PIO_INSTRUCTION_COUNT
     pio: P,
-    interrupts: [Interrupt<P>; 2],
 }
 
 impl<P: PIOExt> core::fmt::Debug for PIO<P> {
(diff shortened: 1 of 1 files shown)


rp2040-hal/src/pio.rs
```diff
@@ -118,6 +118,18 @@ impl<P: PIOExt> PIO<P> {
         &self.interrupts
     }
 
+    /// Enable interrupts raised by state machines.
+    ///
+    /// The PIO peripheral has 4 outside visible interrupts that can be raised by the state machines. Note that this
+    /// does not correspond with the state machine index; any state machine can raise any one of the four interrupts.
```
jannic, line 124: This description confuses me. (And, to be honest, the datasheet confuses me, too.)
Let me explain what I _think_ how the PIO interrupts work:

There are 8 interrupt flags which can be accessed from all state machines. That are the 8 possible values which can be set or queried with the `IRQ` or `WAIT` instructions, and that are outside visible in the PIO IRQ register.

4 of them can be routed to two externally visible interrupts, which you call `PioIRQ::Irq0` and `PioIRQ::Irq1`.
Those four are, a little bit confusingly, called `sm0`—`sm3` in the pio interrupt registers (`IRQ0_INTE` etc.), but do _not_ correspond to the 4 state machines.

If that's correct, I would not call it "4 outside visible interrupts", I only see two interrupts from the outside. But they can be connected to 4 of the 8 interrupt flags visible from inside the PIO.

rp2040-hal/src/pio.rs
```diff
@@ -1407,6 +1505,10 @@ impl<P: PIOExt> Interrupt<P> {
     ///
     /// Each of the 4 state machines have their own TX FIFO. This interrupt is raised when the TX FIFO is not full, i.e.
     /// one could push more data to it.
+    #[deprecated(
+        since = "0.7.0",
+        note = "Use the dedicated method on the state machine"
+    )]
     pub fn enable_tx_not_full_interrupt(&self, id: u8) {
         match id {
             0 => self.irq().irq_inte.modify(|_, w| w.sm0_txnfull().set_bit()),
```
jannic, line 1514: Not new in this PR, but I just noticed it:

I think this doesn't fulfill the safety contract stated above:
```
// Safety: `StateMachine` is marked Send so ensure all accesses remain atomic and no new concurrent
// accesses are added.
```
`irq_inte.modify(...)` is not atomic. Would it be possible to use the atomic bit operations instead, like you did in you new implementations?

ithinuel: 🤦   I had the fix committed (4min before your review) but didn't push it 🤦

rp2040-hal/src/pio.rs
```diff
@@ -1366,168 +1477,123 @@ impl<P: PIOExt> Interrupt<P> {
     /// The PIO peripheral has 4 outside visible interrupts that can be raised by the state machines. Note that this
     /// does not correspond with the state machine index; any state machine can raise any one of the four interrupts.
     pub fn enable_sm_interrupt(&self, id: u8) {
-        match id {
-            0 => self.irq().irq_inte.modify(|_, w| w.sm0().set_bit()),
-            1 => self.irq().irq_inte.modify(|_, w| w.sm1().set_bit()),
-            2 => self.irq().irq_inte.modify(|_, w| w.sm2().set_bit()),
-            3 => self.irq().irq_inte.modify(|_, w| w.sm3().set_bit()),
-            _ => panic!("invalid state machine interrupt number"),
+        assert!(id < 3, "invalid state machine interrupt number");
```
jannic, line 1480: Should be `id < 4`?

rp2040-hal/src/pio.rs
ithinuel: Your understanding matches my confused understanding of the datasheet.

rp2040-hal/src/pio.rs
ithinuel: Good catch! it should now be fixed.

jannic (approved):

PR #448, opened 2022-09-11
fix concurrent accesses to sm_execctrl and sm_instr when sideset isn't optional

ithinuel:
This also includes an alternative implementation of the install function with the hope that using pattern matching will give more opportunity to the compiler to prune the branches we are not interested in.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,3 +28,4 @@ rp2040-hal = { path = "./rp2040-hal" }
 
 [patch.crates-io]
 rp2040-hal = { path = "./rp2040-hal" }
+i2c-pio = { git = "https://github.com/ithinuel/i2c-pio-rs", branch = "fix-pio-exec-instruction" }
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -476,22 +476,20 @@ impl<SM: ValidStateMachine> UninitStateMachine<SM> {
 
     // Safety: The Send trait assumes this is the only write to sm_clkdiv
     fn set_clock_divisor(&self, int: u16, frac: u8) {
-        self.sm().sm_clkdiv.write(|w| {
-            unsafe {
-                w.int().bits(int);
-                w.frac().bits(frac);
-            }
-
-            w
-        });
+        // Safety: This is the only write to this register
+        unsafe {
+            self.sm()
+                .sm_clkdiv
+                .write(|w| w.int().bits(int).frac().bits(frac));
+        }
     }
 
-    fn sm(&self) -> &rp2040_pac::pio0::SM {
-        unsafe { &*self.sm }
+    unsafe fn sm(&self) -> &rp2040_pac::pio0::SM {
+        &*self.sm
     }
 
-    fn pio(&self) -> &rp2040_pac::pio0::RegisterBlock {
-        unsafe { &*self.block }
+    unsafe fn pio(&self) -> &rp2040_pac::pio0::RegisterBlock {
+        &*self.block
     }
 }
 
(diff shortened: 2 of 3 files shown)


rp2040-hal/src/pio.rs
```diff
@@ -531,17 +527,25 @@ impl<SM: ValidStateMachine, State> StateMachine<SM, State> {
...
-    pub fn exec_instruction(&mut self, instruction: u16) {
-        // TODO: clarify what happens if the instruction stalls.
-        self.sm.exec_instruction(instruction);
+    /// If an instruction written to INSTR stalls, it is stored in the same instruction latch used
+    /// by OUT EXEC and MOV EXEC, and will overwrite an in-progress instruction there. If EXEC
+    /// instructions are used, instructions written to INSTR must not stall.
+    pub fn exec_instruction(&mut self, instruction: Instruction) {
+        let instruction = instruction.encode(self.program.side_set);
+
+        critical_section::with(|_| {
```
jannic, line 543: I don't like all those added critical sections.
Most of those instances are already guarded from concurrent accesses by the fact that the functions take a mutable reference to the state machine.

The only exception I found is `Tx::drain_fifo(&mut self)`. `Tx` doesn't contain a reference to the state machine, and the write access is realized with help of an unsafe access in `Tx::register_block(&self)`.
What about moving `drain_fifo` to the state machine, or alternatively adding a `&mut StateMachine` parameter, to guarantee exclusive access. That would make the critical sections unnecessary.

BTW, this safety comment is very sus, as the same function appears in both `Tx` and `Rx`:
```
    fn register_block(&self) -> &pac::pio0::RegisterBlock {
        // Safety: The register is unique to this Tx instance.
        unsafe { &*self.block }
    }
```

ithinuel: Yes, that's the only reason we need those critical sections. I was tempted to make those calls non-mut but I didn't like that.

It seemed weird to have the drain fifo for tx on the SM without one for Rx, but I'm not against it.
Do you think we should go this way to get rid of the critical sections ?

jannic: Yes, I very much prefer the solution without the critical sections.
While the additional overhead is probably not significant for those functions, critical sections are a global resource: They could affect other users of critical sections, and add noticeable interrupt latency.
I think any critical section in library code should be avoided if possible, and if really needed, they should be limited to code which usually only runs once during initialization, or be as short as possible (eg. a read/modify/write operation).

rp2040-hal/src/pio.rs
```diff
@@ -1225,7 +1328,6 @@ unsafe impl<SM: ValidStateMachine + Send> Send for Tx<SM> {}
 // are added.
 impl<SM: ValidStateMachine> Tx<SM> {
     fn register_block(&self) -> &pac::pio0::RegisterBlock {
-        // Safety: The register is unique to this Tx instance.
```
jannic, line 1228: Just removing the comment doesn't make the code correct :-)
(But I'll have a closer look if there's really something unsafe here, and if necessary, open a ticket about it. That's not related to this PR)

Cargo.toml
```diff
@@ -28,3 +28,4 @@ rp2040-hal = { path = "./rp2040-hal" }
 
 [patch.crates-io]
 rp2040-hal = { path = "./rp2040-hal" }
+i2c-pio = { git = "https://github.com/ithinuel/i2c-pio-rs", branch = "fix-pio-exec-instruction" }
```
jannic, line 31: I guess the plan is to remove this as soon as the branch is merged?

ithinuel: Yes

rp2040-hal/src/pio.rs
```diff
@@ -480,19 +486,13 @@ impl<SM: ValidStateMachine> UninitStateMachine<SM> {
...
-            .sm_instr
-            .write(|w| unsafe { w.sm0_instr().bits(instruction) })
-    }
-
     fn sm(&self) -> &rp2040_pac::pio0::SM {
         unsafe { &*self.sm }
     }
+
+    fn pio(&self) -> &rp2040_pac::pio0::RegisterBlock {
+        unsafe { &*self.block }
```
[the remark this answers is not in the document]
ithinuel: Yes, `fn sm(…)`, `fn pio(…)` and `fn register_block(…)` all serve the same purpose. Safety cannot be documented at that point because it depends on what operation is actually done on the registers.

having them all as `unsafe fn` seems fair to me.

jannic (approved):
Thanks, @ithinuel, for all the work you spent on answering my review comments.
Documenting all the unsafe accesses individually is a huge improvement.

PR #449, opened 2022-09-12
Added set_fifos/set_rx_watermark/set_tx_watermark

papyDoctor:
I added the possibility to enable/disable the UART Rx/tx FIFOs as well as setting the watermark for the Rx/Tx  FIFO independently.
I removed the watermark setting on the enable_rx_interrupt function. Since the default value for the watermark is the same as the one set in the enable_rx_interrupt function, it is backward compatible.

### DIFF
--- a/rp2040-hal/src/uart/reader.rs
+++ b/rp2040-hal/src/uart/reader.rs
@@ -2,7 +2,7 @@
 //!
 //! This module is for receiving data with a UART.
 
-use super::{UartDevice, ValidUartPinout};
+use super::{FifoWatermark, UartDevice, ValidUartPinout};
 use rp2040_pac::uart0::RegisterBlock;
 
 use embedded_hal::serial::Read;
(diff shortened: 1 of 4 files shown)


rp2040-hal/src/uart/reader.rs
```diff
@@ -52,16 +52,37 @@ pub(crate) fn is_readable<D: UartDevice>(device: &D) -> bool {
...
+/// See DS: Table 423
+pub fn set_rx_watermark(rb: &RegisterBlock, watermark: FifoWatermark) {
+    let wm = match watermark {
+        FifoWatermark::Bytes4 => 0,
+        FifoWatermark::Bytes8 => 1,
+        FifoWatermark::Bytes16 => 2,
+        FifoWatermark::Bytes24 => 3,
+        FifoWatermark::Bytes28 => 4,
+    };
+    rb.uartifls.modify(|_r, w| unsafe { w.rxiflsel().bits(wm) });
```
ithinuel, line 79: This unsafe block could be avoided by making the field an enum in the pac.

ithinuel (commented):
LGTM, Thank you for your PR :)

The comment bellow can be left for a possible later improvement.

jannic (approved):

rp2040-hal/src/uart/reader.rs
jannic: As this requires changes in a different crate first, let's merge it as it is. The pac change can be done separately.

PR #451, opened 2022-09-13
timer: Mark Timer & Alarm* Send and Sync

ithinuel:


rp2040-hal/src/timer.rs
```diff
@@ -340,6 +335,18 @@ macro_rules! impl_alarm {
...
+            fn drop(&mut self) {
+                self.disable_interrupt();
+                release_alarm($armed_bit_mask)
+            }
+        }
+
+        // safety: there can only be only 1 instance of this alarm so no shared mutable access
+        unsafe impl Send for $name {}
+        // safety: all write operations are synchronised and all reads are atomic
+        unsafe impl Sync for $name {}
```
jannic, line 349: AFAIK Alarm* already implement Send and Sync even without explicit declaration, as those are auto traits.
https://docs.rs/rp2040-hal/latest/rp2040_hal/timer/struct.Alarm0.html#impl-Send-for-Alarm0

rp2040-hal/src/timer.rs
```diff
@@ -14,10 +14,26 @@ use crate::atomic_register_access::{write_bitmask_clear, write_bitmask_set};
...
 
 /// Instant type used by the Timer & Alarm methods.
 pub type Instant = TimerInstantU64<1_000_000>;
 
+static ALARMS: AtomicU8 = AtomicU8::new(0x0F);
+fn take_alarm(mask: u8) -> bool {
+    critical_section::with(|_| {
+        let alarms = ALARMS.load(Ordering::Relaxed);
+        ALARMS.store(alarms & !mask, Ordering::Relaxed);
+        (alarms & mask) != 0
```
jannic, line 27: I think this should check that only one bit in `mask` is set. (Or, easier, don't pass a mask to this function, but the bit number).
Otherwise one could call it with multiple bits set, which would return `true` in case one of the alarms was still available, but the caller wouldn't know which one.
Of course this is not critical as this function isn't `pub`.

rp2040-hal/src/timer.rs
```diff
@@ -65,47 +77,30 @@ impl Timer {
...
-            self.alarms[3] = false;
-            Some(Alarm3(PhantomData))
-        } else {
-            None
-        }
+        take_alarm(1 << 3).then_some(Alarm3(PhantomData))
     }
 }
+// safety: there can only be only 1 instance of this alarm so no shared mutable access
+unsafe impl Send for Timer {}
```
jannic, line 101: Also already implemented, https://docs.rs/rp2040-hal/latest/rp2040_hal/timer/struct.Timer.html#impl-Send-for-Timer
(but not Sync)

rp2040-hal/src/timer.rs
```diff
@@ -14,10 +14,26 @@ use crate::atomic_register_access::{write_bitmask_clear, write_bitmask_set};
...
+use core::sync::atomic::{AtomicU8, Ordering};
 
 /// Instant type used by the Timer & Alarm methods.
 pub type Instant = TimerInstantU64<1_000_000>;
 
+static ALARMS: AtomicU8 = AtomicU8::new(0x0F);
+fn take_alarm(mask: u8) -> bool {
+    critical_section::with(|_| {
+        let alarms = ALARMS.load(Ordering::Relaxed);
+        ALARMS.store(alarms & !mask, Ordering::Relaxed);
```
jannic, line 26: I'm always unsure about this: Does `critical_section` provide any ordering guarantees?
The docs seem to be quiet about that topic. So maybe al the critical section guarantees is that no two threads can be inside a critical section at the same time.
If that's true, it may be necessary to use a stronger ordering here.
(At least formally, if we only consider rust guarantees. In practice, the rp2040 probably acts sufficiently benign that Relaxed will do...)

rp2040-hal/src/timer.rs
ithinuel: This is an internal method so I figured it'd be less change to reuse the `$armed_bit_mask` already defined in the macro.

rp2040-hal/src/timer.rs
ithinuel: hmm indeed, I got a bit heavy handed :P

rp2040-hal/src/timer.rs
ithinuel: the cortex-m0 does not reorder instructions or data accesses and our implementation (and cortex_m::interrupt::enable/disable) has compiler fences that should prevent bad things to happen.

jannic: Yes, discussed that on the rust embedded matrix channel as well. Even though the docs don't spell it out, the critical section should guarantee ordering.

jannic (approved):

PR #453, opened 2022-09-17
simplify jump offset calculation when loading PIO programs

jannic:


ithinuel:
I  was hoping in the previous PR that using pattern matching would let rust eliminate the unecessary decode branches.

Do you know how much is saved by this change?

jannic:
Good point. No, I didn't check that, will do that tomorrow.

jannic:
As a test, I used `cargo build --release --example pio_blink` with and without this change. (That example doesn't even use a `jmp` instruction, so in theory, the compiler could optimize the whole logic away. In practice, it doesn't.)
The resulting firmware sizes were:
- 8420 bytes without the patch
- 7840 bytes with the patch

With `cargo bloat` I can see that the following functions are no longer included in the new code:
- `pio pio::InstructionOperands::decode` (280 bytes)
- `std <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold` (192 bytes)
- `pio pio::Instruction::decode` (174 bytes)

Interestingly, the code size for `rp2040_hal::pio::PIO<P>::install` grew by 68 bytes.

ithinuel:
That makes sense, the 600 bytes are probably left for the validation/panics that decode still does if it does look like a jmp instruction but has invalid parameters.
The fact that install grew is most likely because of the now inlined code :) 

I'm happy with the extra 600 that make sure the hal is not patching instruction that are incorrect but i'm not against the PR either 🤷 

I was thinking of adding a type to pio-rs to differentiate verified instructions (through encode, pio-proc, the assembler etc) and raw instruction injected into a program but I haven't found the time to do so yet.

jannic:
With `lto=true` the difference becomes even larger:
```
 File  .text   Size       Crate Name
12.2%  78.4% 2.0KiB   pio_blink pio_blink::__cortex_m_rt_main
 1.1%   6.8%   176B         pio pio::InstructionOperands::encode
 0.7%   4.7%   120B         pio pio::Instruction::encode
```
vs.
```
File  .text   Size       Crate Name
7.2%  80.5% 1.1KiB   pio_blink pio_blink::__cortex_m_rt_main
```
The rest stays identical, binary size 3352 bytes vs. 2108 bytes.

The difference shrinks again if I add `#[inline(always)]` to several functions in `pio-rs`. So I guess the difference is not really inherent to the code, but due to the way rustc optimizes it.

For firmware running from flash I'd agree that a few 100 bytes don't matter much. In case one wants to load the code to RAM, it may become more important to safe some space.

BTW, was able to shrink the binary by a few more bytes by doing the calculations using u8 instead of usize. (Honestly didn't expect that, because the registers are 32bit anyways.) Now down to 2056 bytes (with LTO).

Of course, these are all micro optimizations. I wouldn't consider them at all if the code became significantly less readable. But the resulting code doesn't look unreadable at all. Without comments and the assert, the logic boils down to:
```
                    if instr & 0b1110_0000_0000_0000 == 0 {
                        let address = (instr & 0b11111) as u8;
                        let address = address + offset;
                        instr & (!0b11111) | address as u16
                    }
```
Sure, to understand the bit masks one needs to know the PIO instruction encoding (or look it up in the datasheet), but isn't that a reasonable expectation when working on HAL code?

ithinuel (approved):

PR #454, opened 2022-09-18
Fix servo example in rp2040-servo, remove unusable aliases.

Nashenas88:
The servo example included phase correction, which does not work with the provided divider. Updated comments to explain numbers and also explain alternate calculation with phase correction enabled.

Remove unusable aliases from all servo pins. The board includes a 180 Ohm resistor on all servo data pins, which makes the lines not very usable for Spi, Uart or I2C. The aliases were removed given they will rarely, if ever, be used with this board.

### DIFF
--- a/boards/pimoroni-servo2040/examples/pimoroni_servo2040_pwm_servo.rs
+++ b/boards/pimoroni-servo2040/examples/pimoroni_servo2040_pwm_servo.rs
@@ -67,10 +67,31 @@ fn main() -> ! {
     const MAX_PULSE: u16 = 2000;
 
     let mut pwm: hal::pwm::Slice<_, _> = pwm_slices.pwm0;
-    pwm.set_ph_correct();
-    // 50Hz
+
+    // 50Hz desired frequency
+    // Rp2040 clock = 125MHz
+    // Top = 65_535, resolution for counter (maximum possible u16 value)
+    // Wrap = Top+1 (number of possible values)
+    // Phase correction multiplier = 2 if phase correction enabled, else 1
+
+    // Divider = rp2040_clock / (Wrap * phase_correction_multiplier * desired_frequency)
+    // Divider = 125,000,000/(65_536 * 1 * 50)
+    // Divider = 38.14639
+    // Divider int = 38
+    // Divider frac = 3 (3/16 = 0.1875, smallest frac greater than desired clock divider).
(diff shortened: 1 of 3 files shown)


Nashenas88:
Should I also include the gifs with this PR, or should I do that as a follow-up?
servo-2040-rainbow
servo-2040-servo
servo-2040-pwm-signal
PXL_20220917_195236027

boards/pimoroni-servo2040/examples/pimoroni_servo2040_pwm_servo.rs
```diff
@@ -67,10 +67,31 @@ fn main() -> ! {
...
+    // Wrap = Top+1 (number of possible values)
+    // Phase correction multiplier = 2 if phase correction enabled, else 1
+
+    // Divider = rp2040_clock / (Wrap * phase_correction_multiplier * desired_frequency)
+    // Divider = 125,000,000/(65_536 * 1 * 50)
+    // Divider = 38.14639
+    // Divider int = 38
+    // Divider frac = 3 (3/16 = 0.1875, smallest frac greater than desired clock divider).
     pwm.set_div_int(38);
     pwm.set_div_frac(3);
```
jannic, line 83: Is it a good choice to use fractional dividers when using pwm to control a servo?
I know close to nothing about servo motors, but as I understand it they do not really care about the exact frequency as long as it's in the right ballpark. So I wonder if the jitter introduced by the fractional divider is a bigger issue than a slightly different frequency would be. (But then, with a top value of 65535, the clock jitter would probably average out over the length of a pulse, so it doesn't really matter either.)

jannic (approved):

jannic:
> Should I also include the gifs with this PR, or should I do that as a follow-up?

Both would be fine with me.

9names:
Just so we're clear: the gifs should be added to the readme via the github web interface editor so they are stored outside of git

Nashenas88:
@9names I didn't see any option to upload the images through the web editor. I just used the links to the content in my comment above. Let me know if there's a different way I should approach that.

9names:
> I just used the links to the content in my comment above.

Should be fine. I think github keeps the content as long as it is regularly accessed, so it should stay online.
I'll make some manual backups just in case.

9names (approved):

PR #455, opened 2022-09-19
added SPI is_busy function

papyDoctor:
When using non blocking read/write with CS managed by user, this function permits to deassert CS at the right time.


ithinuel (approved):
LGTM.

jannic (approved):

Issue #13, opened 2021-02-02
DMA on the way!


AsafFisher:
Implement the DMA api for all the peripherals  :memo:

tdittr:
Things I had on my list for the DMA:
- Example: https://docs.rs/stm32f1xx-hal/0.7.0/stm32f1xx_hal/dma/index.html
- Traits: https://crates.io/crates/embedded-dma

AsafFisher:
Thank you for the guidance! I will look in to it (:

AsafFisher:
I dont like `https://crates.io/crates/embedded-dma`'s traits why should I follow them?

tdittr:
Well, I guess it would allow more interoperation, similar to the embedded-hal traits.

maximeborges:
What is the status of this implementation?

9names:
@maximeborges The branch where this work took place is https://github.com/rp-rs/rp-hal/tree/rp-dma
It has not been touched in 7 months. I would call the status of this implementation "abandoned".
Use it as a starting point if you wish, but if you a need for DMA I suggest starting a new branch and working on it rather than waiting for this to be finished.

mgottschlag:
See #209 for a more complete take on DMA which is already usable.

lightsing:
How is it going? Is there any working on progress?

9names:
No progress

eloraiby:
@AsafFisher any news on this one?

9names:
AsafFisher has not been active on this project in over a year.
No progress to report

ogoffart:
I have been using the PR #209 in my project (details in this comment: https://github.com/rp-rs/rp-hal/pull/209#issuecomment-1156582772 ) I have rebased the change to the recent branch in https://github.com/slint-ui/rp-hal/tree/dma-0.6

9names:
I did a rebase of your rebase earlier today.
https://github.com/9names/rp-hal/tree/dma-0.6
The plan was going to write + run some tests to see what works, maybe clean up the implementation and write some docs

ogoffart:
That would be nice to see that merged. I also tried to upgrade to embedded_dma 0.2 in https://github.com/slint-ui/rp-hal/tree/embedded_dma2 (which i just pushed now),  but i haven't tried it yet.

9names:
Closed via #457

PR #457, opened 2022-09-23
DMA (alternative to #209)

9names:
Rebased #209 onto current trunk.
Updated to embedded-dma 0.2

Current plan is to drop everything except for SingleChannel, since I think that should end up being simpler and more flexible than the current approach.
To get bidirectional functionality from SingleChannel, will need to update peripherals to provide TX/RX endpoints so that the user can have 2 channels targeting the same peripheral for reading/writing.
For double-buffering I was thinking to pass the second DMA channel as an Option to the constructor, so the same interface is used regardless of single-shot/double buffering.

Will probably rename SingleChannel to Channel or Transfer or something to match the above changes.

EDIT:
  After splitting it up, it looks like it already resembles the above somewhat so this might only require minor changes

Other goals:
- support transfer widths other than 32bit for peripherals
- address concerns with current interface raised in #13 and #209


9names:
embedded-dma constrains DMA traits to WORD sizes, where WORD is u8/u16/u32/u64/i8/i16/i32/i64.
64bit sizes compile fine but break at runtime. There are now tests that show this is broken, not sure about the best way to proceed to fix that.

9names:
Removed target tests from this branch since there's no practical way to keep CI happy with them in there. Will keep them in a separate repo for now.

9names:
Turns out the solution to the CI problem was to exclude the on-target test project from the workspace.
There's still the rest of the problems that come from having a top-level .cargo/config, but those are at least potentially resolvable.

rp2040-hal/src/dma/single_channel.rs
```diff
@@ -0,0 +1,208 @@
...
+                .write(|w| unsafe { w.bits(dest) });
+        } else {
+            self.ch().ch_write_addr.write(|w| unsafe { w.bits(dest) });
+        }
+    }
+
+    fn set_chain_to_enabled<CH2: SingleChannel>(&mut self, other: &mut CH2) {
+        // We temporarily pause the channel when setting CHAIN_TO, to prevent any race condition
+        // that could occur, as we need to check afterwards whether the channel was successfully
+        // chained to this channel or whether this channel was already completed. If we did not
```
jannic, line 182: How does that interact with DREQ grants? Are they still counted while the channel is disabled?

If they are note, the DREQ counter could get out of sync.

rp2040-hal/src/pio.rs
```diff
@@ -59,10 +60,21 @@ pub trait PIOExt:
...
-impl PIOExt for PIO0 {}
-impl PIOExt for PIO1 {}
+impl PIOExt for PIO0 {
+    fn id() -> usize {
+        0
+    }
+}
+impl PIOExt for PIO1 {
+    fn id() -> usize {
+        0
```
jannic, line 75: 1?

9names: Yeah, makes sense

rp2040-hal/src/dma/single_channel.rs
9names: What do you mean by "DREQ grants"? I can't find this term anywhere.

9names:
I merged in https://github.com/9names/rp-hal/pull/4 and rebased on latest main, apologies for the noise

rp2040-hal/src/dma/single_channel.rs
jannic: I did some experiments, temporarily pausing a channel and reading DBG_CTDREQ. Looks like the DREQ mechanism keeps counting while a the `en` flag is cleared, so this shoud be fine!

9names: Ah, good. I read that section of the datasheet + the pico-sdk and it seemed like it would just keep counting, but I was going to test it out to be sure. Now that you've tested it I'll trust that it works. :+1:

9names:
> Any issues preventing it from being merged?

Well there's at least one issue: it's not finished yet.
The examples are not finished, tests only cover memory-to-memory operations, not all peripherals are implemented yet, and parts of the DMA driver are not complete.
Are you offering to help?

jounathaen:
I could help a little. So, what is left to do?
In the code, I find the following:
- [ ] SPI example TODOS?
- [ ] Ring transfer
- [ ] Drop for `SingleBuffering`
- [ ] Documentation improvements
- [ ] Fix/enhance `pio_dma.rs` example
- [ ] Fix/enhance `spi_dma.rs` example
- [ ] Callbacks?

> The examples are not finished, tests only cover memory-to-memory operations, not all peripherals are implemented yet, 

In any case, we should not wait until everything is polished and final before merging, as a partially working DMA support is still better than no DMA support.

9names:
MVP as far as I'm concerned:
- Remove support for everything that isn't finished.
- Fix up documentation for everything that is left.
- Add an RTIC example to ensure that it is compatible with resources.
- Add an example that uses a peripheral as a source (ADC would be a good candidate).
- Add an example that does TX and RX on the same peripheral (SPI/UART would be good candidates for this)
- Add an example that does peripheral to peripheral transfers (ADC -> UART would be my choice)
- Add an example that does mem-mem transfers (this can be copy-paste from the on-target tests)
- Verify after we've done all this that everything still works for anyone currently using code from this branch (slint is the only project I'm aware of).

I'm planning on doing a merge from trunk today, then remove the incomplete stuff and start on the docs, if you want to help with the examples that would be much appreciated.

9names:
To implement Bidirectional support correctly we really need something to start multiple channels simultaneously.
single_channel only allows setting one bit at a time.
https://github.com/9names/rp-hal/blob/7c832ea7f21784c062bd2171ef9a5bd62d6d4776/rp2040-hal/src/dma/single_channel.rs#L206

The examples + tests I've added "work" but are unsafe and inherently racy.

rp2040-hal/src/dma/bidirectional.rs
```diff
@@ -0,0 +1,99 @@
+use super::{single_channel::SingleChannel, Pace, ReadTarget, WriteTarget};
```
jounathaen, line 1: Honestly, why do we need the bidirectional config at all?
In your SPI example, you showed that the same can be achieved with two unidirectional configs as well.
This sounds to me like application level code, not HAL level code.

rp2040-hal/examples/spi_dma.rs
```diff
@@ -0,0 +1,126 @@
...
+    let dma = pac.DMA.split(&mut pac.RESETS);
+    // Configure GPIO25 as an output
+    let mut led_pin = pins.gpio25.into_push_pull_output();
+    let mut delay = cortex_m::delay::Delay::new(core.SYST, clocks.system_clock.freq().to_Hz());
+
+    // Use DMA to transfer some bytes (single buffering).
+    let tx_buf = singleton!(: [u8; 16] = [0x42; 16]).unwrap();
+    let rx_buf = singleton!(: [u8; 16] = [0; 16]).unwrap();
+
+    // BidirectionalConfig isn't finished, so we can't use that.
```
jounathaen, line 87: I'd throw out the bidirectional config in the example anyway. IMHO, the example should be a starting point with "simple" examples, and should not demonstrate all possible uses. If an example for BidirectionalConfig is desired, why not put it in the documentation of that module?

rp2040-hal/src/spi.rs
```diff
@@ -45,10 +46,29 @@ impl State for Disabled {}
...
+pub trait SpiDevice: Deref<Target = pac::spi0::RegisterBlock> + SubsystemReset {
+    /// The DREQ number for which TX DMA requests are triggered.
+    fn tx_dreq() -> u8;
+    /// The DREQ number for which RX DMA requests are triggered.
+    fn rx_dreq() -> u8;
+}
 
-impl SpiDevice for pac::SPI0 {}
-impl SpiDevice for pac::SPI1 {}
+impl SpiDevice for pac::SPI0 {
```
jounathaen, line 56: One could use https://docs.rs/rp2040-pac/latest/rp2040_pac/dma/ch/ch_ctrl_trig/enum.TREQ_SEL_A.html here, instead of "magic numbers"

rp2040-hal/src/dma/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+/// The DREQ value for I2C1's RX FIFO
+pub const DREQ_I2C1_RX: u8 = 35;
+/// The DREQ value for the ADC
+pub const DREQ_ADC: u8 = 36;
+/// The DREQ value for the XIP Streaming FIFO
+pub const DREQ_XIP_STREAM: u8 = 37;
+/// The DREQ value for the XIP SSI TX FIFO
+pub const DREQ_XIP_SSITX: u8 = 38;
+/// The DREQ value for the XIP SSI RX FIFO
+pub const DREQ_XIP_SSIRX: u8 = 39;
```
jounathaen, line 346: As this becomes the public API, I'd strongly suggest to use the PAC enum here as well:
https://docs.rs/rp2040-pac/latest/rp2040_pac/dma/ch/ch_ctrl_trig/enum.TREQ_SEL_A.html

9names: Yes, having these here pre-dates them being in the PAC.
Should definitely move to using the PAC constants.

rp2040-hal/src/dma/bidirectional.rs
9names: > This sounds to me like application level code, not HAL level code.

I disagree. Having to resort to fabricating a SPI singleton and then throwing it away after the transaction, and having to do hacks to get both channels started simultaneously is not what something we want to push on our users.
This is enough to test my tests, but I would not recommend it as a general solution to the problem. We can do better.

jounathaen:
Another suggestion: Could you rename `SingleBuffering` to `SingleBufferedTransfer` (and `Bidirectional` accordingly)?
Found the other names a little clumsy when trying it out.

rp2040-hal/src/dma/mod.rs
9names: These are now removed

rp2040-hal/src/pio.rs
```diff
@@ -59,10 +60,21 @@ pub trait PIOExt:
...
+impl PIOExt for PIO0 {
+    fn id() -> usize {
+        0
+    }
+}
+impl PIOExt for PIO1 {
+    fn id() -> usize {
+        1
+    }
+}
```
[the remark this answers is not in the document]
9names: Resolved in [573e62e](https://github.com/rp-rs/rp-hal/pull/457/commits/573e62ea42c01be568faee4867e7548f4f8ca2ac)

rp2040-hal/src/spi.rs
9names: Resolved in [243af46](https://github.com/rp-rs/rp-hal/pull/457/commits/243af465d2e2715ca2a3167b7943e8a2a83d85d2)

9names:
Bidir is now working for SPI, at least with my trivial tests/examples.
One blocker down.

rp2040-hal/src/dma/mod.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// source in a single DMA operation.
+    ///
+    /// For peripherals, the count should likely be u32::MAX. If a data source implements
+    /// EndlessReadTarget, it is suitable for infinite transfers from or to ring buffers. Note that
+    /// ring buffers designated for endless transfers, but with a finite buffer size, should return
+    /// the size of their individual buffers here.
+    ///
+    /// # Safety
+    ///
+    /// This function has the same safety guarantees as `StaticReadBuffer::static_read_buffer`.
```
jannic, line 166: `StaticReadBuffer` doesn't exist in the latest version of embedded_dma. Should probably be changed to `ReadBuffer::read_buffer`? At least that's how I understand the [changelog](https://github.com/rust-embedded/embedded-dma/blob/master/CHANGELOG.md): "Replace less strict ReadBuffer and WriteBuffer definitions with those of StaticReadBuffer and StaticWriteBuffer. This removes the separate static traits."

rp2040-hal/src/dma/mod.rs
```diff
@@ -0,0 +1,265 @@
...
+
+impl<B: ReadBuffer> ReadTarget for B {
+    type ReceivedWord = <B as ReadBuffer>::Word;
+
+    fn rx_treq() -> Option<u8> {
+        None
+    }
+
+    fn rx_address_count(&self) -> (u32, u32) {
+        // Safety: We only call the function once per buffer.
```
jannic, line 187: I don't understand this safety comment: As `rx_address_count` is not unsafe, what keeps some random safe code from calling this method multiple times?

jannic (approved):
I still haven't had time to try this code. But as it reads nicely, contains test cases, and has been commented by multiple persons, I'd say we shouldn't wait for a more in-depth review before merging it. It has been waiting long enough.

rp2040-hal/src/dma/mod.rs
9names: That makes sense with my understanding of the 0.2 changes.
There's a bunch more documentation to fix up as well, unfortunately. Hoping that we'll get more interest for that with the code in-tree than we have with PRs.

rp2040-hal/src/dma/mod.rs
9names: Comment was from the original PR, but agree that it doesn't make sense.
Maybe the original wasn't going to expose this to end-users?
Also not sure what the side-effect is that was hoped to be avoided, I think I'll write some tests (after merge) to exercise this condition to see if it's a problem.

9names: ah.
```rust
/// # Safety
///
/// Once this method has been called, it is unsafe to call any `&mut self`
/// methods on this object as long as the returned value is in use (by DMA).
unsafe fn read_buffer(&self) -> (*const Self::Word, usize);
```
but this is also true if the addresses returned are copied and used for another DMA transfer - the function in and of itself is safe. I'll remove the comment as it's more confusing than helpful.

jounathaen:
I suggest resolving https://github.com/rp-rs/rp-hal/issues/528 before merging, as this would be a major API change afterwards.

9names:
Alright, it's now updated to the API discussed in #528.
Looks okay to me :) but would appreciate another pair of eyes to review how the API looks @jounathaen

jounathaen:
I think it is very nice now! Couldn't find anything more to complain about :wink: 

Thank you so much for your efforts!

9names:
I did a test rebase and it looked sensible, so I'm going to trust that Github's "Squash and merge" is going to do the right thing :crossed_fingers:

PR #459, opened 2022-09-25
Add rtic-monotonic support for timer & alarms (feature gated)

ithinuel:


rp2040-hal/src/timer.rs
```diff
@@ -381,3 +381,49 @@ impl_alarm!(Alarm3 {
...
+
+        const DISABLE_INTERRUPT_ON_EMPTY_QUEUE: bool = false;
+
+        fn now(&mut self) -> Instant {
+            self.0.get_counter()
+        }
+
+        fn set_compare(&mut self, instant: Instant) {
+            if let Err(super::ScheduleAlarmError::AlarmTooLate) = self.1.schedule_at(instant) {
+                panic!("Instant too far in the future, it cannot be scheduled.");
```
jannic, line 414: Is it really necessary to panic at this point?

I checked a few other implementations. Some just set the comparator value of the timer to the past without checking it at all, some do set the timer to the next tick. But I didn't find a single one which panics. 

I wonder what "Note: This method does not need to handle race conditions of the monotonic, the timer queue in RTIC checks this." in the trait docs really means, does it cover this situation?

ithinuel: It turns out it is not. From what I understand, the interrupt may trigger any time before the instant and should trigger short after the target time (eg Systick-monotonic does not even do anything in set_compare because the interrupt gets triggered everytime anyways).

So I updated this method accordingly to have an intermediate trigger if the target instant is too far in the future.

rp2040-hal/src/timer.rs
```diff
@@ -381,3 +381,50 @@ impl_alarm!(Alarm3 {
...
+        }
+
+        fn set_compare(&mut self, instant: Instant) {
+            // The alarm can only trigger up to 2^32 - 1 ticks in the future.
+            // So, if `instant` is more than 2^32 - 2 in the future, we use `max_instant` instead.
+            let max_instant = self.0.get_counter() + 0xFFFF_FFFE.micros();
+            let wake_at = core::cmp::min(instant, max_instant);
+
+            // Cannot fail
+            let _ = self.1.schedule_at(wake_at);
```
jannic, line 416: Why "cannot fail"?
What about AlarmTooSoon?

rp2040-hal/src/timer.rs
```diff
@@ -381,3 +381,50 @@ impl_alarm!(Alarm3 {
...
+            self.1.enable_interrupt();
```
jannic, line 417: I'm not sure how the hardware reacts if the alarm expires first, and then the interrupt gets enabled.
Wouldn't it be safer to enable the interrupt before scheduling the alarm?

rp2040-hal/src/timer.rs
```diff
@@ -381,3 +381,50 @@ impl_alarm!(Alarm3 {
...
+        const DISABLE_INTERRUPT_ON_EMPTY_QUEUE: bool = false;
+
+        fn now(&mut self) -> Instant {
+            self.0.get_counter()
+        }
+
+        fn set_compare(&mut self, instant: Instant) {
+            // The alarm can only trigger up to 2^32 - 1 ticks in the future.
+            // So, if `instant` is more than 2^32 - 2 in the future, we use `max_instant` instead.
+            let max_instant = self.0.get_counter() + 0xFFFF_FFFE.micros();
```
jannic, line 412: Why not `0xFFFF_FFFF`?

rp2040-hal/src/timer.rs
ithinuel: I do not see any detail about TooSoon cased documented, [rp2040-monotonic](https://github.com/korken89/rp2040-monotonic/blob/master/src/lib.rs#L60) treats it as being after overflow, so 🤷 

What do you think it should do?

rp2040-hal/src/timer.rs
ithinuel: The flag is set by hardware regardless of the interrupt being enabled (in `INTR`). When the interrupt is enabled in `INTE`, the signal will propagate to the NVIC where the rest of the magic happens. (that is at least what I'd expect).

rp2040-hal/src/timer.rs
ithinuel: To be conservative mostly and it does not hurt to wake 1us earlier :)

rp2040-hal/src/timer.rs
jannic: I don't know how rtic uses the monotonic timer. If it can handle additional interrupts, the safest bet would be to trigger one immediately, I guess?
I see two cases where a TooSoon error could occur:
1) Some logic bug in the time calculations. In that case, it could be better to somehow signal the error, but I don't think there is a good way to do this here? If the calculated time was wrong in the first place, triggering the alarm immediately wouldn't make things worse, I guess.
2) The alarm time was correctly calculated as an instant very close in the future, and something delayed processing so `schedule_at` was called at or after that instant. In that case, triggering the interrupt immediately is the best thing to do.

ithinuel: That makes me think, if we can detect it to be TooSoon, we could as well manually pend the interupt in the NVIC 🤔 

Or we could:
1. Pause the timer
2. if timestamp > counter => set alarm to timestamp
3. else set alarm = counter
4. resume timer

This'd required to use a critical_section::with context to synchronize the two cores (in case they have an alarm each) as they'd be accessing in write to the pause reg.

jannic: Pausing the timer sounds like it wouldn't help the accuracy of that timer.
But just pending the interrupt could be a good solution. Or setting the INTF flag? (And clear it again in clear_compare_flag, I guess.)

ithinuel: What do you think about the latest changes?

jannic (approved):

PR #461, opened 2022-09-28
Upgrade eh alpha

jannic:


rp2040-hal/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Update embedded-hal alpha support to version 1.0.0-alpha.8 - @jannic
```
ithinuel, line 12: Shouldn't that be alpha.9 ?

jannic: Of course :-)

ithinuel (approved):

PR #462, opened 2022-10-03
Document HAL features in Config.toml(s), and enable ROM function caching in BSPs

9names:
The features of the rp2040-hal were not documented in any of the Config.toml files.
Also, some features were not re-exposed by the BSPs so would be difficult to discover.
Finally, updated the default features of each BSP to include the USB enumeration errata and ROM function pointer caching, since it is likely that most users will want these but will have difficulty finding them on their own.


ithinuel (approved):
LGTM

9names:
Removed USB enumeration errata feature from defaults in the last commit, no other changes

jannic (approved):

PR #464, opened 2022-10-09
Fix watchdog counter load

Sizurka:
With the migration to fugit in #361 the watchdog counter calculation was inadvertently changed.  The watchdog tick rate is set to 1 microsecond by [`init_clocks_and_plls`](https://github.com/rp-rs/rp-hal/blob/eac51cdd160cfd6db6937a8d22fdb76a761e823c/rp2040-hal/src/clocks/mod.rs#L333) however the reload counter was changed to milliseconds with the switch.  The misleading field name `delay_ms` presumably contributed to this.

This PR fixes the counter reload units back to microseconds and renames the field to `delay_us` to remove the confusion.

### DIFF
--- a/rp2040-hal/src/watchdog.rs
+++ b/rp2040-hal/src/watchdog.rs
@@ -41,7 +41,7 @@ use fugit::MicrosDurationU32;
 /// Watchdog peripheral
 pub struct Watchdog {
     watchdog: WATCHDOG,
-    delay_ms: u32,
+    delay_us: u32,
 }
 
 impl Watchdog {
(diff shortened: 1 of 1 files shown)


rp2040-hal/src/watchdog.rs
```diff
@@ -122,17 +122,17 @@ impl watchdog::WatchdogEnable for Watchdog {
 
         // Due to a logic error, the watchdog decrements by 2 and
         // the load value must be compensated; see RP2040-E1
-        self.delay_ms = period.into().to_millis() * 2;
+        self.delay_us = period.into().to_micros() * 2;
```
jannic, line 125: In theory this could overflow, which could be fixed by checking the max value before doing the multiplication.
In practice it probably doesn't matter because most firmwares will just hard code some sane value.

jannic (approved):

PR #468, opened 2022-10-12
Check watchdog load value max value before multiplying by 2, to avoid integer overflow

### LINKED ISSUE #465: TODO: Add a sanity check handling overflow for the period passed to Watchdog::start
see https://github.com/rp-rs/rp-hal/pull/464#discussion_r991043893

jannic:
Fixes #465


9names (approved):
Lgtm

PR #469, opened 2022-10-13
Spi api changes

pietgeursen:
There are few things that I discovered I need from the SPI. 

This PR is a bit of conversation starter. I'm happy to split it up / close it if it's not useful. But I thought I should at least try and get my work back upstream :)

## `Spi::set_baudrate` is slow

Slow is relative of course, but it takes ~100us at runtime on rp2040. That's a deal breaker for me because I have multiple devices on a single SPI bus and I need to be able to reconfigure the SPI and write to it it every 50us.

I extracted the guts of the calculation into `const fn calc_spi_clock_divider_settings_for_baudrate`. This returns a `SpiClockDividerSettings` which can then be passed to `pub fn set_baudrate_from_settings(&mut self, settings: &SpiClockDividerSettings)` which is what actually writes registers. 

This enables you to only do that calculation when you actually need to.

`set_baudrate` is refactored to use these functions internally.

## `Spi::set_mode(&mut self, mode: Mode)`

I _think_ this is safe to do. It's just a convenience function to update the `Mode`on an enabled `Spi`without having to `disable` it and create a new one.

##  `Spi::pub fn complete_transfer(&mut self) -> Result<(), nb::Error<Infallible>> `

I found a footgun with the [embedded hal spi trait](https://docs.rs/embedded-hal/latest/embedded_hal/spi/trait.FullDuplex.html)

The rp2040 has fifo of width 8. Imagine you do this:

```rs
chip_select.set_low().unwrap();
nb::block!(spi.send(byte1)).unwrap();
nb::block!(spi.send(byte2)).unwrap();
nb::block!(spi.send(byte3)).unwrap();
chip_select.set_high().unwrap();
```

Assuming the fifo was empty to start with, those `sends` will never block. The implementation only returns `WouldBlock` when the transmit buffer is full, not when the transfer is complete. That's arguably the correct thing to do. But my chip select will get asserted in the middle of the `sends`.

Now you can do this:

```rs
chip_select.set_low().unwrap();
nb::block!(spi.send(byte1)).unwrap();
nb::block!(spi.send(byte2)).unwrap();
nb::block!(spi.send(byte3)).unwrap();
nb::block!(spi.complete_transfer()).unwrap()
chip_select.set_high().unwrap();
```


rp2040-hal/src/spi.rs
```diff
@@ -78,51 +78,131 @@ impl<S: State, D: SpiDevice, const DS: u8> Spi<S, D, DS> {
...
+        });
+
+        // enable the device
+        self.device.sspcr1.modify(|_, w| w.sse().set_bit());
+    }
+
+    /// Checks if all transmission buffers are empty.
+    ///
+    /// Useful when you need to wait to de-assert a chipselect or reconfigure the device after sending some bytes.
+    pub fn complete_transfers(&self) -> Result<(), nb::Error<Infallible>> {
```
jannic, line 136: `embedded_hal` calls this method `flush`.
Perhaps we should use the same name?
https://docs.rs/embedded-hal/1.0.0-alpha.9/embedded_hal/spi/trait.SpiBusFlush.html#tymethod.flush

pietgeursen: Ooooh nice!

jannic: BTW, https://docs.rs/embedded-hal/1.0.0-alpha.9/embedded_hal/spi/trait.SpiDevice.html#tymethod.transaction might interest you as well. It covers the use case of having multiple transfers while keeping CS asserted.

However, the 1.0.0-alpha traits are not yet fully implemented in this HAL.

rp2040-hal/src/spi.rs
```diff
@@ -78,53 +78,131 @@ impl<S: State, D: SpiDevice, const DS: u8> Spi<S, D, DS> {
...
+                .sph()
+                .bit(mode.phase == Phase::CaptureOnSecondTransition)
+        });
+
+        // enable the device
+        self.device.sspcr1.modify(|_, w| w.sse().set_bit());
+    }
+
+    /// Wait until all operations have completed and the bus is idle.
+    pub fn flush(&self) -> Result<(), nb::Error<Infallible>> {
```
jannic, line 134: Another remark: The future of `nb` is a little bit unclear. It might get replaced by an `async` API.
While I'd say that we should continue supporting the `nb` traits of embedded-hal for the foreseeable future, I wouldn't add methods using it to the the rp2040-hal impls.

So this method should have the same signature as https://docs.rs/embedded-hal/1.0.0-alpha.9/embedded_hal/spi/trait.SpiBusFlush.html#tymethod.flush, `fn flush(&mut self) -> Result<(), Self::Error>`.

ptpaterson: `Self::Error` is [currently being implemented](https://github.com/rp-rs/rp-hal/blob/2fb3e2a7e96cc9901d38c24b4a60b9960d90e866/rp2040-hal/src/spi.rs#L230) as `Infallable`, so the proposed signature change would mean `flush` would _have_ to be blocking, right?

```rust
    fn flush(&mut self) -> Result<(), Self::Error> {
        while self.device.sspsr.read().bsy().bit() {}
        Ok(())
    }
```

If that's the case, why not make the inherent impl return `()` and allow an eh alpha `SpiBusFlush` implementation to use that.  See[ _Fallability_ discussion](https://github.com/rust-embedded/embedded-hal/blob/74f9b9bd600f195456f6fe588fbc8f59b02b3894/MIGRATING-0.2-1.0.md#fallibility) in proposed migration guide.

```rust
    fn flush(&mut self) {
        while self.device.sspsr.read().bsy().bit() {}
    }
```

Or for that matter, if it is helpful to have an inherent non-blocking implementation (I honestly don't know if that's helpful) why not let the inherent impl be non-blocking (keep the current signature with `Result<(), nb::Error<Infallible>>`) to be consistent with the other inherent impls and the `SpiBusFlush` trait impl loop over it to make it blocking.

ptpaterson: Just realized this PR is based on a relatively old commit -- this needs a rebase. 

So if rebased, `Spi<Enabled, D, DS>` now has an `is_busy` method which can be used to implement flush. I think it makes sense to only implement `flush` for enabled SPI anyway.

```rust
impl<D: SpiDevice, const DS: u8> Spi<Enabled, D, DS> {
    /* ... */
    fn flush(&self) {
        while self.is_busy {}
    }
}
```

which would allow later:

```rust
    #[cfg(feature = "eh1_0_alpha")]
    impl<D: SpiDevice> eh1::SpiBusFlush for Spi<Enabled, D, $nr> {
        fn flush(&mut self) -> Result<(), Self::Error> {
            self.flush();
            Ok(())
        }
    }
```

jannic: > So if rebased, `Spi<Enabled, D, DS>` now has an `is_busy` method which can be used to implement flush.

Exactly, and `is_busy` is functionally equivalent to the non-blocking version of `flush`, so it's sufficient to have the blocking `flush` method.

rp2040-hal/src/spi.rs
```diff
@@ -78,53 +78,131 @@ impl<S: State, D: SpiDevice, const DS: u8> Spi<S, D, DS> {
...
     }
 }
 
+/// Clock divider settings
+pub struct SpiClockDividerSettings {
+    /// The prescaler for writing to sspcpsr
+    pub prescale: u8,
+    /// The postdiv for writing to sspcr0
+    pub postdiv: u8,
+}
```
ithinuel, line 149: Making fields public is a strong commitment (see rust's [future proofing](https://rust-lang.github.io/api-guidelines/future-proofing.html#structs-have-private-fields-c-struct-private) guidelines).

Also, I'm not even sure allowing users to set them to arbitrary values is a good idea since they'd need to be validated before set in the registers (eg only even values are allowed in `CPSDVSR`).

rp2040-hal/src/spi.rs
```diff
@@ -78,53 +78,131 @@ impl<S: State, D: SpiDevice, const DS: u8> Spi<S, D, DS> {
...
+/// This calculation takes ~100us on rp2040 at runtime and is used by [set_baudrate] every time
+/// it's called. That might not be acceptable in
+/// situations where you need to change the baudrate often.
+///
+/// Note that this is a const function so you can use it in a static context if you don't change your
+/// peripheral clock frequency at runtime.
+///
+/// If you do change your peripheral clock at runtime you can store the [SpiClockDividerSettings] and only re-calculate it
+/// when the peripheral clock frequency changes.
+pub const fn calc_spi_clock_divider_settings_for_baudrate(
```
ithinuel, line 163: I think this would make a nice inherent method to `SpiClockDividerSettings` like

```rust
impl SpiClockDividerSettings {
    pub const fn new(peri_frequency: HertzU32, baudrate: HertzU32) -> Self {
        todo!()
    }
}
```

rp2040-hal/src/spi.rs
```diff
@@ -78,53 +78,131 @@ impl<S: State, D: SpiDevice, const DS: u8> Spi<S, D, DS> {
...
+        if self.device.sspsr.read().bsy().bit() {
+            Err(nb::Error::WouldBlock)
+        } else {
+            Ok(())
+        }
     }
 }
 
+/// Clock divider settings
+pub struct SpiClockDividerSettings {
```
ithinuel, line 144: This method should most likely derive:

- Copy, Clone
- PartialEq, Eq
- Debug
- defmt::Format (when the feature is enabled)

ithinuel (changes requested):

PR #471, opened 2022-10-15
Implement UartConfig::new constructor method

jannic:
The change of the baudrate type is up for discussion, if it's controversial, I'll roll it back.

### DIFF
--- a/rp2040-hal/src/uart/utils.rs
+++ b/rp2040-hal/src/uart/utils.rs
@@ -1,8 +1,7 @@
-use fugit::HertzU32;
-
 use crate::pac::{uart0::RegisterBlock, UART0, UART1};
 use crate::resets::SubsystemReset;
 use core::ops::Deref;
+use fugit::HertzU32;
 
 /// Error type for UART operations.
 #[derive(Debug)]
@@ -82,6 +81,23 @@ pub struct UartConfig {
 
     /// The parity that this uart should have
     pub parity: Option<Parity>,
+}
+
+impl UartConfig {
+    /// Create a new instance of UartConfig
+    pub fn new(
+        baudrate: HertzU32,
+        data_bits: DataBits,
+        parity: Option<Parity>,
+        stop_bits: StopBits,
+    ) -> UartConfig {
+        UartConfig {
+            baudrate,
+            data_bits,
+            stop_bits,
+            parity,
+        }
+    }
 }
 
 /// Rx/Tx FIFO Watermark
--- a/rp2040-hal/src/uart/utils.rs
+++ b/rp2040-hal/src/uart/utils.rs
@@ -85,7 +85,7 @@ pub struct UartConfig {
 
 impl UartConfig {
     /// Create a new instance of UartConfig
-    pub fn new(
+    pub const fn new(
         baudrate: HertzU32,
         data_bits: DataBits,
         parity: Option<Parity>,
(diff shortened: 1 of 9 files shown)


rp2040-hal/src/uart/utils.rs
```diff
@@ -84,6 +85,47 @@ pub struct UartConfig {
...
+            baudrate,
+            data_bits,
+            stop_bits,
+            parity,
+        }
+    }
+
+    /// Create a new instance of UartConfig by copying an existing one and
+    /// setting a new baudrate value.
+    pub fn baudrate(&self, baudrate: u32) -> UartConfig {
```
ithinuel, line 106: This looks a bit like the builder pattern except that it usually takes and returns an `&mut self` : https://doc.rust-lang.org/1.12.1/style/ownership/builders.html

rp2040-hal/src/uart/utils.rs
```diff
@@ -72,7 +73,7 @@ pub enum Parity {
 #[non_exhaustive]
 pub struct UartConfig {
     /// The baudrate the uart will run at.
-    pub baudrate: HertzU32,
+    pub baudrate: u32,
```
ithinuel, line 76: If the type does not carry the unit, I think it should be at least documented.

```suggestion
    /// The baudrate in bit per second the uart will run at.
    pub baudrate: u32,
```

jannic: The unit of a baudrate is a baud, not bit per second. So it's already in the name.
https://en.wikipedia.org/wiki/Baudrate

For an uart which only transfers one bit per symbol, the numeric values are the same, so the units are often confused. But I'd rather not increase that confusion by documenting the unit of a variable named "baudrate" with a unit of "bit per second".

rp2040-hal/src/uart/utils.rs
jannic: That would make using it more complicated: `UartPeripheral::enable(...)` takes an owned `UartConfig`, so something like `uart.enable(UartConfig::default().baudrate(9600))`, would no longer work. Alternatives would be:
```
  let mut config = UartConfig::default();
  config.baudrate(9600); // which isn't better than config.baudrate = 9600;
  uart.enable(config);
```
or
```
  uart.enable(UartConfig::default().baudrate(9600).build())
```
with
```
  impl UartConfig {
    fn build(&self) -> UartConfig {
      ...
    }
  }
```

In the end, I don't like this builder-like usage too much anyways, as mentioned above: It's not possible to see the other parameters besides the baudrate without looking into the docs or implementation of the `default()` function.

So perhaps I should just remove those functions entirely.

jannic:
I reverted both the baudrate type change and the builder-like methods, as they are not core parts of this pull request.
That way, they can be discussed separately.
Also, this avoids making a breaking change.

rp2040-hal/src/uart/utils.rs
ithinuel: Yes, baud is symbols per second. For uart a symbol is a bit so baud is equal to bit per second.
Also baudrate is already redundant in itself (since baud is already a rate) but that's commonly used in the industry 🤷 .

rp2040-hal/src/uart/utils.rs
```diff
@@ -84,6 +83,23 @@ pub struct UartConfig {
     pub parity: Option<Parity>,
 }
 
+impl UartConfig {
+    /// Create a new instance of UartConfig
+    pub fn new(
```
ithinuel, line 88: Would it make sense to make this method const ?

jannic: Good point, that may be useful.

ithinuel (approved):
LGTM

PR #472, opened 2022-10-16
Fix UART read error discard data

Sizurka:
The current handling of a UART error does not accurately report the data before the error.  It currently just returns the whole buffer passed, which will potentially have been partly written by any data read before the error.  Fix this by returning the slice of data that has been read so far.

Also fix a spelling error in the field name.


jannic:
LGTM. Can you add a changelog entry? The fix for the spelling error should be mentioned because it's (formally) a breaking change.

jannic (approved):

jannic:
Thank you!

Issue #481, opened 2022-10-23
Provide a way to construct a DynPin (for C interop)


9names:
cr1901 asked on #rust_embedded if there was a way to construct a pin for use with C code.
As in: existing C project. new code written in Rust, assuming it's called through an `extern "C" fn`

We're almost able to construct a DynPin from scratch now, but the DynRegisters struct is private.

Here's an example of the *almost working* syntax we currently have:
```rust
let pin = DynPinId {
    group: Bank0,
    num: 1,
};
let mode = DynPinMode::Output(PushPull);
// This doesn't work as DynRegisters is private
let regs = DynRegisters::new(pin);
let dynpin1 = DynPin { regs, mode };
dynpin1.toggle();
```

It's a bit verbose.
If we were to provide such a facility (and I think we should) how should we enable this?
Should we add an (unsafe) function that allows you to conjure a DynPin based solely on pin number and mode, along with documentation on how it will explode if you get these wrong?
Would appreciate any thoughts on whether we should do this, and how.

cr1901:
For more context: the idea is that the C code provides a header that defines the pin layout. I will parse this on the Rust side to create structs for appropriately initializing the used hardware on the Rust side (and provide a context pointer back to the C side after init is done).

Since there's no real way to enforce this on the C side, it's up to the C side to not call hw init functions that Rust is using.

jannic:
Without commenting on the implementation details, as I didn't have time to read the code yet: l think this is a valid use case and should be supported somehow. If it's only possible with some (unsafe) function which crates the DynPin out of thin air, so be it.

Are there alternatives? What about some container storing the real DynPins on the rust side, and handling the pin id as some kind of handle? That way C code could refer to the pin by its id, but could still only access pins which were previously registered in that structure. But that's probably just overcomplicating the issue.

PR #482, opened 2022-10-23
Make it possible to construct DynPin from scratch

9names:
Basic functionality for #481

### DIFF
--- a/rp2040-hal/src/gpio/dynpin.rs
+++ b/rp2040-hal/src/gpio/dynpin.rs
@@ -305,7 +305,7 @@ impl DynPin {
     /// must be at most one corresponding [`DynPin`] in existence at any given
     /// time.  Violating this requirement is `unsafe`.
     #[inline]
-    unsafe fn new(id: DynPinId, mode: DynPinMode) -> Self {
+    pub unsafe fn new(id: DynPinId, mode: DynPinMode) -> Self {
         DynPin {
             regs: DynRegisters::new(id),
             mode,
--- a/rp2040-hal/examples/gpio_c_interop.rs
+++ b/rp2040-hal/examples/gpio_c_interop.rs
@@ -1,0 +1,103 @@
+#![no_std]
+#![no_main]
+
+// Ensure we halt the program on panic (if we don't mention this crate it won't
+// be linked)
+use panic_halt as _;
+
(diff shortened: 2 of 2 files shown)


cr1901:
I finally got around to testing this... I don't think it's going to be enough; ultimately what I need is "a way to pass in `DynPin`s to e.g. the I2C peripheral, which take type-level `Pins`. There's a `try_from` for `DynPin`, but the problem is _I don't know which type-level pin I need to convert to until runtime._

The SPI peripheral doesn't seem to take in any pins to initialize, so I can safely use that with `DynPin`. What can be done about e.g. `I2c` and `UartPeripheral`?

jannic:
> ultimately what I need is "a way to pass in `DynPin`s to e.g. the I2C peripheral, which take type-level `Pins`. There's a `try_from` for `DynPin`, but the problem is _I don't know which type-level pin I need to convert to until runtime._

That one looks like a separate (but related) issue:
Currently, [struct I2C {...}](https://docs.rs/rp2040-hal/latest/rp2040_hal/i2c/struct.I2C.html) encodes the actual pin used in its type. So it's impossible to create an instance without knowing the pins *at compile time*.

I think this requirement is too strict. I don't see an inherent reason why it shouldn't be possible to construct an I2C peripheral with DynPins. Of course, that either needs some run-time checks (not zero cost - but really not expensive either), or would allow the user to create an I2C peripheral with nonsensical pins. Even the latter wouldn't be a catastrophe, it's not unsound or such, it just won't work.

rp2040-hal/src/gpio/dynpin.rs
```diff
@@ -305,7 +305,7 @@ impl DynPin {
     /// must be at most one corresponding [`DynPin`] in existence at any given
     /// time.  Violating this requirement is `unsafe`.
```
jannic, line 306: Just a side note, as it wasn't changed in this PR: The comment "Violating this requirement is `unsafe`" is a little bit off. The function is `unsafe` because the caller must guarantee that `DynPin`s are unique. Violating such a condition is not `unsafe`, it's plain wrong. It may cause UB, or other bugs. But saying that violation the safety conditions of an `unsafe` function is `unsafe` doesn't convey any useful meaning.

PR #484, opened 2022-10-26
Add ability to modify installed program wrap bounds

davidcole1340:
The wrap source and target cannot change after installing the program
into the PIO at present, even though this is possible with the C/C++
HAL. This is useful when using the same program across multiple state
machines that have different wrap sources and targets.

### DIFF
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -292,7 +292,25 @@ pub struct InstalledProgram<P> {
 }
 
 impl<P: PIOExt> InstalledProgram<P> {
-    /// Get the warp target (entry point) of the instaled program.
+    /// Change the source and/or target for automatic program wrapping.
+    ///
+    /// This replaces the current wrap bounds with a new set. This can be useful if you are running
+    /// multiple state machines with the same program but using different wrap bounds.
+    ///
+    /// # Returns
+    ///
+    /// * [`Ok`] containing a new program with the provided wrap bounds
+    /// * [`Err`] containing the old program if the provided wrap was invalid (outside the bounds of
+    ///   the program length)
+    pub fn set_wrap(self, wrap: Wrap) -> Result<Self, Self> {
+        if (self.offset + wrap.source) < self.length && (self.offset + wrap.target) < self.length {
+            Ok(InstalledProgram { wrap, ..self })
+        } else {
+            Err(self)
+        }
+    }
+
+    /// Get the wrap target (entry point) of the installed program.
     pub fn wrap_target(&self) -> u8 {
         self.offset + self.wrap.target
     }
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -303,7 +303,7 @@ impl<P: PIOExt> InstalledProgram<P> {
     /// * [`Err`] containing the old program if the provided wrap was invalid (outside the bounds of
     ///   the program length)
     pub fn set_wrap(self, wrap: Wrap) -> Result<Self, Self> {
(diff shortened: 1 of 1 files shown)


rp2040-hal/src/pio.rs
```diff
@@ -292,7 +292,25 @@ pub struct InstalledProgram<P> {
...
+    /// This replaces the current wrap bounds with a new set. This can be useful if you are running
+    /// multiple state machines with the same program but using different wrap bounds.
+    ///
+    /// # Returns
+    ///
+    /// * [`Ok`] containing a new program with the provided wrap bounds
+    /// * [`Err`] containing the old program if the provided wrap was invalid (outside the bounds of
+    ///   the program length)
+    pub fn set_wrap(self, wrap: Wrap) -> Result<Self, Self> {
+        if (self.offset + wrap.source) < self.length && (self.offset + wrap.target) < self.length {
```
jannic, line 306: Why is the offset relevant in those checks?
Aren't wrap.source and wrap.target relative to the offset, the same way length is?
So both source and target should be between 0 and length-1, inclusive.

davidcole1340: Thanks, I misunderstood how that worked. Changed now

ithinuel:
> This is useful when using the same program across multiple state
machines that have different wrap sources and targets.

I have the feeling that changing the state on the installed program isn't the right thing.

Maybe an Installed pogram should remain unchanged and the state machines take a ref type (to the program) that holds the SM specific wrap points ?

davidcole1340:
> > This is useful when using the same program across multiple state
> > machines that have different wrap sources and targets.
> 
> I have the feeling that changing the state on the installed program isn't the right thing.
> 
> Maybe an `InstalledProgram` should remain unchanged and the state machines take a ref type (to the program) that holds the SM specific wrap points ?

I agree - have only done it like this since each state machine requires its own instance of `InstalledProgram` when using `PIOBuilder` (via `unsafe fn share()`). So would be interested to see what others think.

jannic:
I agree with @ithinuel that the `set_wrap` on `InstalledProgram` feels a little bit out of place. But at the same time, I don't know a clearly better alternative. @ithinuel, could you explain "the state machines take a ref type (to the program) that holds the SM specific wrap points" a little more? I don't understand what you are suggesting.

And, thinking about it a little longer, I didn't find a reason why this feels wrong. In fact, it seems to be the logical place to do it. After all, an `InstalledProgram` is just a `Program` which "was placed in the PIO's instruction memory". It's not yet assigned to a state machine, it's not yet running. So why not provide the same `set_wrap` method on `InstalledProgram` as on `Program`? It doesn't affect the instructions, so it can be done after installing the program in instruction memory.

I wonder if the feeling is caused by some mismatch between the way  the hardware works and our abstraction. When looking at the registers, the wrap points are not part of the program (ie. some part of the instruction space), but of the state machine. Multiple state machines can share a program, but they always have their own wrap points. But in the API, we configure the wrap points as part of the program, because most of the time, for a given program, there is only one set of wrap points which makes sense.

jannic (approved):
I'd say this change is ok. As far as I can tell it should work as expected. And while the API is not without alternatives, it isn't wrong or ugly.

@ithinuel, you had some doubts. What do you think, shall we merge it as it is, or do you suggest some API improvements?

ithinuel:
Yes, I think you pointed at the right thing in
> the feeling is caused by some mismatch between the way the hardware works and our abstraction

I probably won't have the time to have a serious think about it before jan' 23 so I think we can merge this as a first step towards inabling the usecase.

PR #487, opened 2022-10-28
Specify required feature critical-section-impl for some examples

jannic:
Not perfect, but at least leads to a meaningful error message when the feature is missing:
```
error: target `vector_table` in package `rp2040-hal` requires the features: `rt`, `critical-section-impl`
Consider enabling them by passing, e.g., `--features="rt critical-section-impl"`
```


ithinuel (approved):

9names:
Definitely much better

ithinuel:
I was wondering why this wasn't caught by CI and I think that's because when we build the examples, the BSP features get all merged together which enables `critical-section-impl`.

#487 covers that caveat

PR #489, opened 2022-10-28
Amend CI to build rp2040-hal without the bsp's features.

ithinuel:
requires https://github.com/rp-rs/rp-hal/pull/487 first

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1,6 +1,25 @@
 on: [push, pull_request]
 name: Build and Test check
 jobs:
+  rp2040-hal-builds:
+    name: Build rp2040-hal examples without bsp
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -1,6 +1,25 @@
...
+        with:
+          toolchain: stable
+          target: thumbv6m-none-eabi
+          override: true
+          profile: minimal
+      - name: Build rp2040-hal's examples
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: -p rp2040-hal ${{ matrix.mode }} --examples --features rt,critical-section-impl
```
jannic, line 22: Does this change much if those two features are enabled anyways?

ithinuel: This is intended to catch future changes. EG introduction of an extra features that'd break examples if abscent.

jannic (approved):

PR #490, opened 2022-10-30
Extend pico_usb_serial example with some string formatting

jannic:
Not sure if it should be added here, or as a separate example. But as the question how one can write formatted strings to USB serial has been asked more than once, having some example available would be nice.


9names (approved):
I remember this being one of my first questions as well.
LGTM!

PR #492, opened 2022-11-05
UsbBus::poll clears suspended interrupt when ints.dev_suspend is set

jhillyerd:
My device often sees USBCTRL_IRQ interrupt storms after reset.  This appears to
be because INTS DEV_SUSPEND is set, but SIE.SUSPEND is not cleared by
UsbBus::poll().

### DIFF
--- a/rp2040-hal/src/usb.rs
+++ b/rp2040-hal/src/usb.rs
@@ -615,6 +615,7 @@ impl UsbBusTrait for UsbBus {
 
             // check for bus reset and/or suspended states.
             let sie_status = inner.ctrl_reg.sie_status.read();
+            let ints = inner.ctrl_reg.ints.read();
             let mut buff_status = inner.ctrl_reg.buff_status.read().bits();
 
             if sie_status.bus_reset().bit_is_set() {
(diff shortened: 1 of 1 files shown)


rp2040-hal/src/usb.rs
```diff
@@ -629,7 +630,7 @@ impl UsbBusTrait for UsbBus {
                 #[cfg(not(feature = "rp2040-e5"))]
                 return PollResult::Reset;
             } else if buff_status == 0 && sie_status.setup_rec().bit_is_clear() {
-                if sie_status.suspended().bit_is_set() {
+                if sie_status.suspended().bit_is_set() || ints.dev_suspend().bit_is_set() {
                     inner.ctrl_reg.sie_status.write(|w| w.suspended().set_bit());
                     return PollResult::Suspend;
```
jannic, line 635: I wonder if it's correct to return PollResult::Suspend if sie_status.suspended() is not set.
Clearing the interrupt flag when only ints.dev_suspend is set looks like a valid fix, good catch!
Out of curiosity: what's the USB host you use when this issue occurs?

jhillyerd: I agree that this is probably not the correct way to fix this. Still discussing in matrix room.

The issue repros on a USB 2.0 hub, or direct to a USB 2.0 port on host, or USB 3.0 port on host.

jannic: What kind of host? It might be operating system or even driver dependent.

jhillyerd: Linux fractal 5.15.70 #1-NixOS SMP Fri Sep 23 12:15:52 UTC 2022 x86_64 GNU/Linux

jhillyerd:
Obsoleted by #493

PR #493, opened 2022-11-07
usb: Read ints rather than sie_status in poll

ithinuel:
If the latency is too high, there is a risk that the status reverts to a previous state while keeping the interrupt flag up. This fixes that by relying on ints instead of sie_status to process the events.

This should fix the issue reported by @jhillyerd in #492.

### DIFF
--- a/rp2040-hal/src/usb.rs
+++ b/rp2040-hal/src/usb.rs
@@ -614,12 +614,12 @@ impl UsbBusTrait for UsbBus {
             }
 
             // check for bus reset and/or suspended states.
-            let sie_status = inner.ctrl_reg.sie_status.read();
+            let ints = inner.ctrl_reg.ints.read();
             let mut buff_status = inner.ctrl_reg.buff_status.read().bits();
 
-            if sie_status.bus_reset().bit_is_set() {
+            if ints.bus_reset().bit_is_set() {
                 #[cfg(feature = "rp2040-e5")]
-                if sie_status.connected().bit_is_clear() {
+                if inner.ctrl_reg.sie_status.read().connected().bit_is_clear() {
                     inner.errata5_state = Some(errata5::Errata5State::start());
                     return PollResult::None;
                 } else {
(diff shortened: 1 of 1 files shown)


jhillyerd:
I tested this on my device, and can confirm it resolves the interrupt storm I was seeing.

jannic:
Could this sequence of events happen?

1. USB is in suspended state
2. resume is signaled by the hardware, `ints.dev_resume_from_host` is set
3. before the interrupt is handled (eg. because interrupts were temporary disabled), suspend is signaled by the hardware, `ints.dev_suspend` is set
4. interrupt handler runs, `poll` is called
5. `poll` first checks `ints.dev_suspend`, sees that it is set, clears the flag, and returns `PollResult::Suspend`. This is ignored by usb-device as the device is already in suspended state. 
6. interrupt handler runs again, because `ints.dev_resume_from_host` is still set, `poll` is called
7. as `ints.dev_suspend` is no longer set, poll now reacts to `ints.dev_resume_from_host`, clears that flag, and returns `PollResult::Resume`

If this happened, the order of `PollResult::Suspend` and `PollResult::Resume` would be wrong. This would cause the suspend event to get lost. The software stack would think the device state was `Configured` (or whatever the state was before the initial suspend event) even though it should be `Suspend`.

jannic:
I wonder if the old handling of SIE_STATUS had the same (perhaps theoretical) issue. At least according to the documentation, the flags in SIE_STATUS are WC bits (set by hardware, cleared by writing a 1 bit from the CPU).

If that's true, in the example given above, SIE_STATUS.RESUME and SIE_STATUS.SUSPENDED would both be set, and we also would not know the current hardware state.

ithinuel:
Yes, I'd need some time to figure a sensible way to handle that (if there is one).

jannic:
So shall we just merge this PR as it is, to solve the irq storm issue from #492?
Especially as it looks like this is not a regression and the undefined situation where two events are signaled at once can already happen when reading SIE_STATUS?

jannic (approved):

jannic:
@ithinuel I guess this can be merged, right?

ithinuel:
Yes I think so.

PR #494, opened 2022-11-09
Allow setting clock divisors on running state machines

jannic:
The documentation of CLKDIV_RESTART contains the following sentence:

"Note also that CLKDIV_RESTART can be written to whilst the state machine is running, and this is useful to resynchronise clock dividers after the divisors (SMx_CLKDIV) have been changed on-the-fly."

This implies that it's allowed to change the value of CLKDIV on a running state machine.

### DIFF
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -636,6 +636,20 @@ impl<SM: ValidStateMachine, State> StateMachine<SM, State> {
             }
         }
     }
+
+    /// Change the clock divider of a state machine.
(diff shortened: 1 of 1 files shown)


ithinuel:
Setting the clock div on the fly alone does not reset the phase (the counter).
If we are to move allow live change of the clock, I think we should have something in the doc warning that it may have unexpected side effects.

jannic:
Related comment on the raspberry forum:
https://forums.raspberrypi.com/viewtopic.php?p=1989421&hilit=change+CLKDIV#p1989421

ithinuel (approved):

PR #496, opened 2022-11-12
build: add docs.rs metadata

icedrocket:


rp2040-hal/Cargo.toml
```diff
@@ -7,6 +7,10 @@ homepage = "https://github.com/rp-rs/rp-hal"
 description = "A Rust Embeded-HAL impl for the rp2040 microcontroller"
 license = "MIT OR Apache-2.0"
 
+[package.metadata.docs.rs]
+features = ["rt"]
```
ithinuel, line 11: Would it be worth adding `rtic-monotonic` and `defmt` to have the extra types and derive show in the doc too?

jannic (approved):
I didn't know about `package.metadata.docs.rs` and had to look it up: https://docs.rs/about/metadata
Looks good. Perhaps also add `rom-v2-intrinsics` to the features list?

PR #497, opened 2022-11-12
Avoid dead_code warning if critical-section-impl is not enabled

jannic:
Also make Spinlock31 public if it's not used by the critical section impl


ithinuel (approved):
LGTM

PR #500, opened 2022-11-15
Add Waveshare RP2040 Zero board

TilCreator:
This adds the board config for the Waveshare RP2040 Zero board.
I also copied the NeoPixel example from the feather board, but didn't copy the blinky example as the board only has that single Neopixel LED.
I successfully compiled and ran the waveshare_rp2040_zero_neopixel_rainbow example on the board.

### DIFF
--- a/boards/waveshare-rp2040-zero/src/lib.rs
+++ b/boards/waveshare-rp2040-zero/src/lib.rs
@@ -1,0 +1,785 @@
+#![no_std]
+
+pub extern crate rp2040_hal as hal;
+
+#[cfg(feature = "rt")]
+extern crate cortex_m_rt;
+#[cfg(feature = "rt")]
+pub use hal::entry;
+
+/// The linker will place this boot block at the start of our program image. We
+/// need this to help the ROM bootloader get our code up and running.
+#[cfg(feature = "boot2")]
+#[link_section = ".boot2"]
+#[no_mangle]
+#[used]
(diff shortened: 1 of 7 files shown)


boards/waveshare-rp2040-zero/src/lib.rs
```diff
@@ -0,0 +1,785 @@
...
+            /// PWM Function alias for pin [crate::Pins::gpio15].
+            FunctionPwm: Gp15Pwm7B,
+            /// PIO0 Function alias for pin [crate::Pins::gpio15].
+            FunctionPio0: Gp15Pio0,
+            /// PIO1 Function alias for pin [crate::Pins::gpio15].
+            FunctionPio1: Gp15Pio1
+        }
+    },
+
+    /// GPIO 18 is connected internaly to a single Neopixel RGB LED
```
9names, line 469:
```suggestion
    /// GPIO 16 is connected internally to a single Neopixel RGB LED
```

9names (changes requested):
Looks good, just a couple of minor fixes and this will be good to go.

9names (approved):
LGTM.
Next time, can you not rebase/squash after review starts unless requested?
GitHub doesn't track changes very well when you do this which makes it harder to work out how you've addressed review feedback. We can squash as part of the merge process to clean up the history, so it's not really helpful most of the time.

PR #506, opened 2022-11-28
Use generic boot loader for arduino nano connect

### LINKED ISSUE #503: arduino_nano_connect should use the generic boot loader
According to https://github.com/rp-rs/rp2040-boot2/issues/19, there are versions of the Arduino Nano Connect with different flash chips (perhaps due to chip shortage?).

Therefore, the BSP should use the generic boot loader - which might be slower, but should work with all chips. Also add some comment about this so people depending on the faster flash speed could switch back to a boot loader specific to their flash chip.

jannic:
Fixes #503


ithinuel (approved):
LGTM

PR #510, opened 2022-12-01
Fixed: First frame is getting lost on an USB-CDC device.

MuratUrsavas:
**Issue**

While using `usb_device` and `usbd_serial` crates with `rp2040_hal`, the first frame from the host device is always getting lost. Consecutive frames work just fine.

**Root Cause**

The `LENGTH_0` region of the `EP_BUFFER_CONTROL` register is always reset (0) while waiting for the first frame. After the first frame failure, the poll operation implicitly sets it  to the necessary value (`max_packet_size`) and this clears the problem for the rest of the app execution lifetime.

The main problem is while resetting and readjustment of the bits needs a step by step operation to avoid potential issues. While doing that and setting `AVAILABLE` bit, the wrong method had been chosen and was invalidating previous changes. Hence the `LENGTH_0` param is reset, too.

**Proposed Solution**

Instead of direct register `write` operation I've used `modify` to keep current bits in the register.

### DIFF
--- a/rp2040-hal/src/usb.rs
+++ b/rp2040-hal/src/usb.rs
@@ -307,7 +307,7 @@ impl Inner {
                     w.length_0().bits(ep.max_packet_size)
                 });
                 cortex_m::asm::delay(12);
-                buf_control.write(|w| w.available_0().set_bit());
+                buf_control.modify(|_, w| w.available_0().set_bit());
             }
         }
     }


ReinoutHeeck:
Heh, 
 I just went down this same rabbit hole the past couple of days and came up with the exact same solution.
I was about to create a pull request but see you have beat me to it.

Since I independently came to the same result I believe your fix to be correct for this issue.

ithinuel (approved):
LGTM

MuratUrsavas:
@ReinoutHeeck Sorry to hear that we both worked on the same issue. I'd leave it to you if I knew :smile: But this was the only outstanding issue before releasing my new project firmware. So I had to fix it.

This is my first contribution into this repo, and hope won't be the last. 

@ithinuel Seems like I don't have necessary privileges to merge this PR into main. How do we proceed from here?

9names (approved):
@MuratUrsavas You don't have to do anything, just need to wait for one of the maintainers to find time to press the merge button :)

LGTM

PR #511, opened 2022-12-06
Pico USB to UART passthrough example

mvniekerk:
I needed a USB device to program my Pycom Lopy4 board.


mvniekerk:
Question - I've used a static Option to leak out a static reference to something. Is there a more elegant way to do it?

mvniekerk:
I've tested it on real hardware, it locks up.

PR #512, opened 2022-12-06
Pico usb to uart

mvniekerk:
A pass-through of USB bytes to UART.


jannic:
@mvniekerk, as the boards were moved to a separate repository, https://github.com/rp-rs/rp-hal-boards, could you open a new PR there?

thejpster:
I'd like to see this added here as a generic example as reading through it I found it very educational.

PR #516, opened 2022-12-10
Remove pico_i2c_pio example

jannic:
Two reasons:
- it depends on an unreleased version of i2c-pico
- i2c-pico depends on rp2040-hal = "0.6.0"

The latter causes a circular dependency between the git repos rp-hal and i2c-pico. This makes updating version numbers more difficult: It requires incremental updates, first updating the version of rp2040-hal, then updating i2c-pico to use the newer hal version, and only then updating the board support crates and their examples.

This could be reconsidered once we move the BSPs into separate repositories, https://github.com/rp-rs/rp-hal/issues/244.


thejpster:
I'm in favour of this. It can always come back later.

jannic:
> This was discussed in the past and IIRC we have a related PR to add the example in the i2c-pio repo.

Correct! I had a quick look at the repo, but didn't notice the open PR. https://github.com/rp-rs/i2c-pio-rs/pull/6

jannic:
This would be superseded by https://github.com/rp-rs/rp-hal/pull/517 in case the BSP move gets approved.

jannic:
#517 was merged, so this PR is obsolete

PR #519, opened 2022-12-11
Dynpin interrupts

### LINKED ISSUE #476: Interact with interrupts with DynPin
DynPin should have enough info to enable/check/clear interrupt status for a pin, but doesn't seem to have any way to access these functions (they're only available for Pin).

```rust
    let pin = pins.key1.into_pull_up_input();
    let status = pin.interrupt_status(EdgeLow);
    let dynpin: DynPin = pin.into();
    // this doesn't work
    let status2 = dynpin.interrupt_status(EdgeLow);
```

image

larsarv:
Added functions to interact with interrupts to DynPin.
It's a copy of the same functions found in Pin.

Resolves #476


jannic:
LGTM

jannic (approved):

jannic:
... but of course `rustfmt` is more picky than I am. @larsarv, could you fix the formatting?

Issue #522, opened 2022-12-14
We need a way to cancel an ongoing alarm


MuratUrsavas:
I've already implemented the solution and creating this ticket for better tracking.

jannic:
Fixed by #523

PR #523, opened 2022-12-14
Added Alarm cancellation feature.

MuratUrsavas:
**Issue**

#522

In current scenario there is no way to disable a triggered alarm. The only choice is to cancel the timer but it also disables other ongoing alarms. This has a negative effect if the alarm is used for detection of timeouts and should be cancelled if the current state changes.

Scheduling alarm for zero is out of option as it's simply rejected. And it wouldn't look good for readability.

**Solution**

I've added `cancel` method to the `Alarm` trait. This way only the related alarm could be cancelled without effecting other alarms and the timer.

### DIFF
--- a/rp2040-hal/src/timer.rs
+++ b/rp2040-hal/src/timer.rs
@@ -207,6 +207,9 @@ pub trait Alarm {
 
     /// Return true if this alarm is finished.
     fn finished(&self) -> bool;
+
+    /// Cancel an activated alarm.
+    fn cancel(&mut self) -> Result<(), ScheduleAlarmError>;
 }
 
 macro_rules! impl_alarm {
@@ -339,6 +342,22 @@ macro_rules! impl_alarm {
                 // safety: This is a read action and should not have any UB
                 let bits: u32 = unsafe { &*TIMER::ptr() }.armed.read().bits();
                 (bits & $armed_bit_mask) == 0
+            }
+
+            /// Cancel an activated Alarm. No negative effects if it's already disabled.
+            /// Unlike `timer::cancel` trait, this only cancels the alarm and keeps the timer running
+            /// if it's already active.
+            fn cancel(&mut self) -> Result<(), ScheduleAlarmError> {
+                unsafe {
+                    let timer = &*TIMER::ptr();
+                    timer.armed.write_with_zero(|w| w.bits($armed_bit_mask));
+                    crate::atomic_register_access::write_bitmask_clear(
+                        timer.intf.as_ptr(),
+                        $armed_bit_mask,
+                    );
+                }
+
+                Ok(())
             }
         }
 


rp2040-hal/src/timer.rs
```diff
@@ -1,8 +1,8 @@
 //! Timer Peripheral
 //!
-//! The Timer peripheral on RP2040 consists of a 64-bit counter and 4 alarms.  
-//! The Counter is incremented once per microsecond. It obtains its clock source from the watchdog peripheral, you must enable the watchdog before using this peripheral.  
-//! Since it would take thousands of years for this counter to overflow you do not need to write logic for dealing with this if using get_counter.  
+//! The Timer peripheral on RP2040 consists of a 64-bit counter and 4 alarms.
+//! The Counter is incremented once per microsecond. It obtains its clock source from the watchdog peripheral, you must enable the watchdog before using this peripheral.
+//! Since it would take thousands of years for this counter to overflow you do not need to write logic for dealing with this if using get_counter.
```
ithinuel, line 5: Please leave these spaces. In Markdown they force line feed within a paragraph (`<br/>`).

rp2040-hal/src/timer.rs
```diff
@@ -340,6 +343,16 @@ macro_rules! impl_alarm {
...
             }
+
+            /// Cancel an activated Alarm. No negative effects if it's already disabled.
+            /// Unlike `timer::cancel` trait, this only cancels the alarm and keeps the timer running
+            /// if it's already active.
+            fn cancel(&mut self) -> Result<(), ScheduleAlarmError> {
+                // safety: Only read operations are made on the timer and they should not have any UB
+                let timer = unsafe { &*TIMER::ptr() };
+
+                self.schedule_internal(timer, Instant::from_ticks(0))
```
ithinuel, line 354: I think it'd be better to directly clear the relevant flag with:

```rust
unsafe {
    timer.armed.write_with_zero(|w| w.bits($armed_bit_mask));
    crate::atomic_register_access::write_bitmask_set(
        timer.intf.as_ptr(),
        $armed_bit_mask,
    );
}
```

ithinuel (changes requested):

rp2040-hal/src/timer.rs
MuratUrsavas: Well, this should be happened due to automatic trailing white space removal of the editor. Will revert it back to the original.

rp2040-hal/src/timer.rs
MuratUrsavas: @ithinuel Thanks for the hint. Implemented similar to what you said but with a modification as your code was forcing the timer process the alarm even though it was not armed.

rp2040-hal/src/timer.rs
MuratUrsavas: Done.

jannic (approved):

rp2040-hal/src/timer.rs
jannic: @MuratUrsavas, it looks like your code does exactly what @ithinuel suggested. What's the modification you are mentioning? It looks like the current version still triggers the interrupt unconditionally.

MuratUrsavas:
```rust
unsafe {
    timer.armed.write_with_zero(|w| w.bits($armed_bit_mask));
    crate::atomic_register_access::write_bitmask_set(       <<<<<< I'm using write_bitmask_clear instead of set.
        timer.intf.as_ptr(),
        $armed_bit_mask,
    );
}
```

As I said above, I'm using `write_bitmask_clear` instead of `write_bitmask_set`. This way it prevents forcing interrupt even it's not armed. I've tried and it wasn't triggering after arming and cancelling. (Could be any source re-triggering it?) I've proved that via tracing the SFR's step by step and then with some practical tests.

On the other hand, I've seen other unintended alarm interrupts even though the cancel method was not implemented. My guess is there could be a bug in `timer.rs` but this is just a hunch. I don't have a proof, yet.

jannic: LOL, sorry, I swear I read the code 5 times and still missed that difference :-)

MuratUrsavas: @jannic No worries. That was like finding 7 differences in the picture :smile:

PR #526, opened 2022-12-18
Disambiguate a comment

jannic:
While reviewing some other merge request, I first read that comment as "in the past some value was set" instead of "the value we just set is already in the past". This is just a minor rewording to avoid that ambiguity.


9names (approved):
Yep, much more clear now

PR #527, opened 2022-12-19
example name type wrong

tianrking:
error: no example target named `pico_pwm_blink`.
Available example targets:
    gpio_irq_example
    vector_table
    adc
    watchdog
    rom_funcs
    pio_synchronized
    pio_proc_blink
    multicore_fifo_blink
    gpio_in_out
    multicore_polyblink
    pio_blink
    uart
    spi
    blinky
    lcd_display
    dht11
    i2c
    pio_side_set
    pwm_blink

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ specified 'runner'. As the 'runner' is the elf2uf2-rs tool, it will build a UF2
 file and copy it to your RP2040.
 
 ```console
-$ cargo run --release --example pico_pwm_blink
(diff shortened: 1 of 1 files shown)


jannic:
Hi @tianrking!

Your observation that the example given in the README no longer works was right. It's been caused by moving the BSP crates to a different repository.
The README should point to an example from this repo, not from the boards repo. One possible solution would be to change the line in the README to `cargo run --release --example blinky`. (The same line occurs twice. Both should be fixed.)

PR #533, opened 2022-12-22
Install cargo-udeps manually

jannic:
This avoids using cargo-udeps-action. Build times will probably increase, as  installs from source,\nbut cargo-udeps-action is very unreliable at the moment.

See https://github.com/aig787/cargo-udeps-action/issues/2

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -28,9 +28,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@nightly
         with:
           target: thumbv6m-none-eabi
+      - name: Install cargo-udeps
+        run: cargo install cargo-udeps --locked
       - name: Check unused deps
-        uses: aig787/cargo-udeps-action@v1
-        with:
-          version: latest
-          args: ${{ matrix.mode }} --workspace ${{ matrix.features }}
-
+        run: cargo udeps ${{ matrix.mode }} --workspace ${{ matrix.features }}


jannic:
Building cargo-udeps takes ~5m in github CI.

Currently, cargo-udeps is called in matrix mode:
```
  args: ${{ matrix.mode }} --workspace ${{ matrix.features }}
```

Is that really useful?

9names:
5 minutes is painful :cry: 
I wrote up an alternate version of this PR that pulls a release binary for cargo-udeps instead, which is practically instant.

Obvious downside that we now have to change the CI script in order to get new versions of cargo-udeps, and possible OS incompatibility in the future.
Would be nice if we could use something like cargo-binstall or cargo-quickinstall for it instead, but it's not supported currently.

Run job:
https://github.com/9names/rp-hal/actions/runs/3762330999/jobs/6394927100
Branch:
https://github.com/9names/rp-hal/tree/use_udeps_binary

9names:
> Currently, cargo-udeps is called in matrix mode ...  Is that really useful?

Features can enable other dependencies - if we want to exhaustively check that we're using everything I think it's necessary.

jannic:
> 5 minutes is painful

Yes, definitely. It was not meant to be a permanent solution.

But your suggestion of just downloading the binary is better, if it works reliably. It's basically what cargo-udeps-action does, just without calling out to `https://api.github.com/repos/est31/cargo-udeps/releases/latest` first to determine the latest version. As it looks like that API call is the one failing, it might be exactly what is needed to get the pipeline stable again.

9names (approved):

9names:
I'm okay either way, it's still better than what we have now.
Opened a PR with my branch but also approved this one, in case you want to go with the simple+reliable but slower option instead.

jannic:
After spending some time reading docs, I'm convinced that the issue is the API call, not the download of the release. There is an API call limit of 60 requests per hour for unauthenticated users. With the matrix jobs, a single pipeline run does at least 8 API calls. I guess the limit is per IP address, so if the same runner executes several jobs within an hour, the limit can easily be reached.

So your approach of just downloading the binary should work at least as reliable as building cargo-udeps from scratch. Let's merge #534.

PR #537, opened 2023-01-09
Add SPI slave

### LINKED ISSUE #536: SPI slave implementation not available
I am not sure that the functionality of configuring the peripheral as a slave exists yet, and I think this is an important feature. If so,
please let me know if this can be done with my own pull request, or if this will be implemented later.

zaksabeast:
Closes #536.

### DIFF
--- a/rp2040-hal/src/spi.rs
+++ b/rp2040-hal/src/spi.rs
@@ -167,6 +167,15 @@ impl<D: SpiDevice, const DS: u8> Spi<Disabled, D, DS> {
         });
     }
 
+    /// Set master/slave
+    fn set_slave(&mut self, slave: bool) {
+        if slave {
+            self.device.sspcr1.modify(|_, w| w.ms().set_bit());
+        } else {
+            self.device.sspcr1.modify(|_, w| w.ms().clear_bit());
+        }
+    }
+
     /// Initialize the SPI
     pub fn init<F: Into<HertzU32>, B: Into<HertzU32>>(
         mut self,
@@ -174,12 +183,14 @@ impl<D: SpiDevice, const DS: u8> Spi<Disabled, D, DS> {
         peri_frequency: F,
         baudrate: B,
         mode: &Mode,
+        slave: bool,
     ) -> Spi<Enabled, D, DS> {
         self.device.reset_bring_down(resets);
         self.device.reset_bring_up(resets);
 
         self.set_baudrate(peri_frequency, baudrate);
         self.set_format(DS, mode);
+        self.set_slave(slave);
         // Always enable DREQ signals -- harmless if DMA is not listening
         self.device
             .sspdmacr
--- a/rp2040-hal/src/spi.rs
+++ b/rp2040-hal/src/spi.rs
@@ -214,7 +214,10 @@ impl<D: SpiDevice, const DS: u8> Spi<Disabled, D, DS> {
 
     /// Initialize the SPI in slave mode
     pub fn init_slave(self, resets: &mut RESETS, mode: &Mode) -> Spi<Enabled, D, DS> {
-        self.init_spi(resets, 0u32.Hz(), 0u32.Hz(), mode, true)
+        // Use dummy values for frequency and baudrate.
+        // With both values 0, set_baudrate will set prescale == u8::MAX, which will break if debug assertions are enabled.
+        // u8::MAX is outside the allowed range 2..=254 for CPSDVSR, which might interfere with proper operation in slave mode.
+        self.init_spi(resets, 1000u32.Hz(), 1000u32.Hz(), mode, true)
     }
 }
 
--- a/rp2040-hal/src/spi.rs
+++ b/rp2040-hal/src/spi.rs
@@ -27,6 +27,7 @@ use eh1_0_alpha::spi as eh1;
 use embedded_hal::blocking::spi;
 use embedded_hal::spi::{FullDuplex, Mode, Phase, Polarity};
 use fugit::HertzU32;
+use fugit::RateExtU32;
 use pac::dma::ch::ch_ctrl_trig::TREQ_SEL_A;
 use pac::RESETS;
 
@@ -141,7 +142,6 @@ impl<S: State, D: SpiDevice, const DS: u8> Spi<S, D, DS> {
             .modify(|_, w| unsafe { w.scr().bits(postdiv) });
 
         // Return the frequency we were able to achieve
-        use fugit::RateExtU32;
         (freq_in / (prescale as u32 * (1 + postdiv as u32))).Hz()
     }
 }
@@ -176,8 +176,7 @@ impl<D: SpiDevice, const DS: u8> Spi<Disabled, D, DS> {
         }
     }
 
-    /// Initialize the SPI
-    pub fn init<F: Into<HertzU32>, B: Into<HertzU32>>(
+    fn init_spi<F: Into<HertzU32>, B: Into<HertzU32>>(
         mut self,
         resets: &mut RESETS,
         peri_frequency: F,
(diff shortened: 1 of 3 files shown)


DmitriLyalikov:
@zaksabeast, nice work. I will  test your implementation with a master device and slave device together and characterize what we have on the logic analyzer. will follow up

rp2040-hal/src/spi.rs
```diff
@@ -167,19 +167,30 @@ impl<D: SpiDevice, const DS: u8> Spi<Disabled, D, DS> {
...
+    }
+
     /// Initialize the SPI
     pub fn init<F: Into<HertzU32>, B: Into<HertzU32>>(
         mut self,
         resets: &mut RESETS,
         peri_frequency: F,
         baudrate: B,
         mode: &Mode,
+        slave: bool,
```
jannic, line 186: I wonder if it would be better to have a separate function to init the device in slave mode. For example, is the `baudrate`  parameter needed / meaningful if the clock pin is an input?

Also, boolean parameters are often problematic as it may not be obvious what they mean when reading the code. And having a new, separate init method would avoid a breaking change for existing callers.

zaksabeast: Yeah, good callouts.  I wasn't fond of the parameter, but figured it would be a good starting point for discussion since I'm not familiar with the style of codebase.  A new init method sounds great.

The difference between the two init methods would be pretty slim.  Is there preferred way to reduce duplication since setting baudrate and slave mode happens after the resets and before enabling spi?

kjagiello:
Just tested it out in my project where I needed this functionality and it worked just fine. ESP32 C3 acting as master and a Pico acting as slave. 

On ESP32's end:
```cpp
spi_device_handle_t spi;
esp_err_t ret;
spi_bus_config_t buscfg = {
    .mosi_io_num = PIN_NUM_MOSI,
    .miso_io_num = PIN_NUM_MISO,
    .sclk_io_num = PIN_NUM_CLK,
    .quadwp_io_num = -1,
    .quadhd_io_num = -1,
    .max_transfer_sz = (2 * 8)
};

ret = spi_bus_initialize(SPI2_HOST, &buscfg, SPI_DMA_CH_AUTO);
ESP_ERROR_CHECK(ret);

spi_device_interface_config_t devCfg={
    .mode = 0,
    .clock_speed_hz = 2 * 1000 * 1000,
    .spics_io_num=PIN_NUM_CS,
    .queue_size=3
};

ret = spi_bus_add_device(SPI2_HOST, &devCfg, &spi);
ESP_ERROR_CHECK(ret);

uint8_t data[] = {12, 34};
spi_transaction_t trx = {
    .length = 16,
    .tx_buffer = &data,
    .rx_buffer = NULL,
};

spi_device_acquire_bus(spi, portMAX_DELAY);
spi_device_transmit(spi, &trx);
spi_device_release_bus(spi);
```

On Pico's end:
```rust
let _spi_miso = pins.gpio16.into_mode::<FunctionSpi>();
let _spi_mosi = pins.gpio17.into_mode::<FunctionSpi>();
let _spi_sclk = pins.gpio18.into_mode::<FunctionSpi>();
let _spi_cs = pins.gpio19.into_mode::<FunctionSpi>();
let spi = Spi::<_, _, 8>::new(pac.SPI0);

let mut spi = spi.init(
    &mut resets,
    2.MHz(),
    &embedded_hal::spi::MODE_3,
    true,
);

loop {
    if let Ok(value) = spi.read() {
        println!("received: {}", value);
    }
}
```

It produced the expected logs:
```
INFO  received: 12
└─ dashy::__cortex_m_rt_main @ src/main.rs:134
INFO  received: 34
└─ dashy::__cortex_m_rt_main @ src/main.rs:134
```

From my logic analyzer:
image

rp2040-hal/src/spi.rs
jannic: >  Is there preferred way to reduce duplication since setting the baudrate and slave mode happens after the resets and before enabling spi?

I don't think we have defined a "best practice". For this function, I see at least two viable options:
- Don't worry, just duplicate the code. The method is too small to really matter.
- Have some internal function which is identical with what the code currently does, and make the public functions wrappers for that one. So the documented API is nice and internal code avoids duplication.

zaksabeast: Sorry for the late response.  I went with the second option, but am happy to move to the first if that works better.

DmitriLyalikov:
Hey @zaksabeast , I tested your new init_slave function, and noticed that my slave would not be responsive. I noticed here 
`self.init_spi(resets, 0u32.Hz(), 0u32.Hz(), mode, true) `

You are setting the frequency and baudrate with zero values and although they are irrelevant as a slave, i simply hard coded those to dummy values of 1000u32.Hz each, and had no issues from there. 

Let me know if you would agree with this

-Dmitri

jannic:
First, everyone involved, thanks for the great teamwork!

@DmitriLyalikov, that's an interesting observation. In the datasheet, I don't see any hint that those values are important in slave mode. But there are two things that may go wrong with setting both frequencies to 0:
With both values 0, `set_baudrate` will set `prescale == u8::MAX`. This will break at `debug_assert_ne!(prescale, u8::MAX);` in case debug assertions are enabled.
And then, the value `u8::MAX` is outside the allowed range 2..=254 for CPSDVSR. I have no idea if that invalid value might interfere with proper operation in slave mode.
Setting a dummy value like 1000 would fix both issues. Some code comment explaining that value would be helpful.

zaksabeast:
Yes, thank you everyone for your help!

Curiously I didn't run into issues when testing the 0hz values (granted, it was a release build); perhaps I tested the wrong binary by mistake.  I appreciate the testing from others and having the debug assertions pointed out.

The 1000hz values worked without issue for me too, and I added the requested comments.

jannic:
The clippy warning was not caused by this pull request.
Fix: https://github.com/rp-rs/rp-hal/pull/545

jannic (approved):

DmitriLyalikov:
Hi everyone, since this is related and I am using this environment, Id like to share an issue i am having and get your feedback. 

I am working on a creating the spi slave with RTIC that will respond via the SPI0_IRQ assigned task.
This uses the RP2040 Hal with the spi_slave addition and pi-pico BSP.

Here is my captured output.
image

My problem is that when asserting chip select from my spi master, I don't see this task invoke. Which makes me think that either I need to configure the interrupt, or I am doing something incorrectly. I am assuming that when I set the pins to FunctionSpi, the spi device takes control of them, and Chip select is left floating for the master to drive, I can test this on the hardware, but would like to understand what into_mode does to the pins when using spi. 

Here is my setup in the #init function

    let _spi_mosi = pins.gpio17.into_mode::<hal::gpio::FunctionSpi>();
    let _spi_miso = pins.gpio16.into_mode::<hal::gpio::FunctionSpi>();
    let _spi_cs = pins.gpio19.into_mode::<hal::gpio::FunctionSpi>();
    let spi = hal::Spi::<_, _, 8>::new(c.device.SPI0);
    // Exchange the uninitialized spi device for an enabled slave
    let mut spi_dev = spi.init_slave(&mut resets, &embedded_hal::spi::MODE_0);`

and here is my task bound the SPI0_IRQ

    // Task that binds to the SPI0 IRQ and handles requests. This will execute from RAM
    // This takes a mutable reference to the SPI bus writes immediately from tx_buffer while reading 
    // into the rx_buffer
    #[inline(never)]
    #[link_section = ".data.bar"] // Execute from IRAM
    #[task(binds=SPI0_IRQ, priority=4, local=[spi_dev, spi_tx_consumer], shared = [serial])]
    fn spi0_irq(cx: spi0_irq::Context) {
        let mut tx_buf = [5u8; 18];
        // Write/Read words back to slave. Received words will replace contents in tx_buf
        if let Some(mut tx_buf) = cx.local.spi_tx_consumer.dequeue(){}
        
        cx.local.spi_dev.transfer(&mut tx_buf).unwrap();
        // Received words, Now build our HostRequest
        match HostRequest::new().build_from_8bit_spi(&tx_buf) {
            Ok(hr) => { // Host Request is clean and ready to be sent out
                send_out::spawn(hr);
            }
            Err(err) => { // Print the error to serial port if Host Request is invalid
            }
        }
    } ```

9names:
Please make a new issue instead of commenting on a closed one

PR #540, opened 2023-01-18
Clarify 'finished()' behavior for alarms

nmattia:
This mentions that `finished()` will also return `false` if the alarm has _not_ been scheduled yet. The implementation of `finished()` checks for the `ARMED` bit, which (AFAICT from the datasheet, "Section 4.6.3. Alarms") is only set when the alarm is scheduled (with `schedule_at()`.

### DIFF
--- a/rp2040-hal/src/timer.rs
+++ b/rp2040-hal/src/timer.rs
@@ -205,7 +205,8 @@ pub trait Alarm {
     /// [enable_interrupt]: #method.enable_interrupt
     fn schedule_at(&mut self, timestamp: Instant) -> Result<(), ScheduleAlarmError>;
 
-    /// Return true if this alarm is finished.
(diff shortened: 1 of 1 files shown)


jannic:
I wonder if that should be documented as a fact as part of the API contract, or if it's just an implementation detail which could change with future updates.

The safe bet would be to just say "the returned value is undefined if the alarm has not been scheduled yet".

And I'd add the same sentence to the trait docs at https://github.com/rp-rs/rp-hal/blob/04ee37c2972ff85f16701be9b4e23bea032dd6fc/rp2040-hal/src/timer.rs#L209

In any case: Mentioning that the result might be true before the alarm has been scheduled is a good suggestion!

nmattia:
Yes, that makes sense! Let me change that

PR #543, opened 2023-01-25
Add set_base_1and0 used by simple_blend3 interp example

tomgilligan:
The C/C++ SDK (and datasheet) make use of `base01` (called `base_1and0` in PAC and `INTERP0_BASE_1AND0`/`INTERP1_BASE_1AND0` in datasheet)

https://github.com/raspberrypi/pico-examples/blob/master/interp/hello_interp/hello_interp.c#L132-L140

This PR adds access to this register which is needed for the `simple_blend3` example from the C/C++ SDK.
The `simple_blend3` and `linear_interpolation` examples are also currently missing from https://github.com/rp-rs/rp-hal-boards/blob/main/boards/rp-pico/examples/pico_interpolator.rs .  I'll be opening a PR to addd these shortly.

### DIFF
--- a/rp2040-hal/src/sio.rs
+++ b/rp2040-hal/src/sio.rs
@@ -816,6 +816,8 @@ pub trait Interp {
     fn set_base(&mut self, v: u32);
     ///Read the interpolator Base register (Base2 in the datasheet)
     fn get_base(&self) -> u32;
+    ///Write the lower 16 bits to BASE0 and the upper bits to BASE1 simultaneously. Each half is sign-extended to 32 bits if that lane's SIGNED flag is set
+    fn set_base_1and0(&mut self, v: u32);
 }
 
 macro_rules! interpolators {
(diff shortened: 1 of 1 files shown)


jannic:
This has endless potential for bike shedding, I fear:

The name `set_base_1and0` is close enough to the datasheet, that's nice.

But what's the meaning of that u32 value?
Perhaps it would be more convenient to specify both base values separately?
`set_base_1and0(&mut self, base1 :u16, base0 :u16)`?
But the ordering of the parameters is strange, `base1` before `base0`? However, if they are switched, the function name and the parameter orderings are using different orders:
`set_base_1and0(&mut self, base0 :u16, base1 :u16)`?

I never used the interpolator, so I can't judge if it matches common use cases. If common use cases have the combined value in a single variable anyway, it wouldn't make sense to split the u32 into two u16. And that would also avoid the parameter ordering issue.

So, if nobody chimes in with a different preference, I'll accept the pull request as proposed by @tomgilligan.

jannic (approved):

PR #532, opened 2022-12-21
Update github actions

jannic:
I noticed in https://github.com/rp-rs/rp-hal/actions/runs/3751431401 that the github workflow triggers a lot of deprecation warnings. This pull request should fix some of them.

PR #555, opened 2023-02-23
Update github actions

jannic:
PR #532 updated some github actions. A later merge of some older branch reintroduced some of the deprecated actions. This updates those instances in the same way as #532 did.


9names (approved):
Lgtm

PR #558, opened 2023-03-03
Use name TEMPERATURE_SENSOR_CHANNEL instead of numeric value

jannic:


rp2040-hal/src/adc.rs
```diff
@@ -134,7 +134,7 @@ where
     fn read(&mut self, _pin: &mut PIN) -> nb::Result<WORD, Self::Error> {
         let chan = PIN::channel();
 
-        if chan == 4 {
+        if chan == TEMPERATURE_SENSOR_CHANNEL {
             self.device.cs.modify(|_, w| w.ts_en().set_bit())
         }
```
jannic, line 139: BTW, I wonder if this block could be removed entirely.
If I'm not missing anything, `chan == 4` is only possible if `enable_temp_sensor` was called before.
https://github.com/rp-rs/rp-hal/blob/b1e63b86057c94180e6953119d4b3731423329c5/rp2040-hal/src/adc.rs#L85-L89
So `ts_en()` should already be enabled at this point.

9names: I think you are correct, this should function identically with this code removed

jannic: While writing a commit message for the change removing that code, I noticed that the following sequence is possible:
```
  let mut temperature_sensor = adc.enable_temp_sensor();
  let mut another_temperature_sensor = adc.enable_temp_sensor();
  adc.disable_temp_sensor(another_temp_sensor); // ts_en is now cleared!
  adc.read(&mut temperature_sensor).unwrap(); // trying to read temperature while ts_en is cleared...
```
Not sure what to do about that, but for now I'll keep the code, the overhead is small enough.

9names (approved):

PR #560, opened 2023-03-04
Leverage typelevel implementation in i2c, pwm (and a little bit in uart).

ithinuel:
This notably makes the documentation more readable.

### DIFF
--- a/rp2040-hal/src/uart/pins.rs
+++ b/rp2040-hal/src/uart/pins.rs
@@ -1,59 +1,69 @@
-use crate::gpio::{bank0, FunctionUart, Pin};
+use crate::gpio::{bank0, AnyPin, FunctionUart, OptionalPin};
 use crate::pac::{UART0, UART1};
-use crate::typelevel::Sealed;
+use crate::typelevel::{NoneT, Sealed};
 
 use super::UartDevice;
 
 /// Declares a valid UART pinout.
 pub trait ValidUartPinout<UART: UartDevice> {
-    /// Indicates TX should be enabled for this pinout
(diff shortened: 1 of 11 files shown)


rp2040-hal/src/uart/pins.rs
```diff
@@ -1,59 +1,69 @@
...
-    CTS: Cts<UART>,
-    RTS: Rts<UART>,
+    TX: OptionalPin,
+    RX: OptionalPin,
+    CTS: OptionalPin,
+    RTS: OptionalPin,
+    TX::Id: ValidTxPin<UART>,
+    RX::Id: ValidRxPin<UART>,
+    CTS::Id: ValidCtsPin<UART>,
+    RTS::Id: ValidRtsPin<UART>,
```
jannic, line 29: I'm currently only reading this on my phone so it's quite likely that I'm missing the big picture, literally. But I wonder: does this still, somehow, guarantee that the combination of pins comprises a valid set of pins for the UART peripheral, as the name of the struct implies?

jannic: (it probably does, because the restriction on the Id types implies that the TX/RX/... types are also valid for the UART, right?)

ithinuel: Yes. At least that is how I understand it.

jannic:
BTW, this is a breaking change as it removes some traits.

ithinuel:
Converting to draft as I'm progressing on GPIO overhaul and more breaking changes.

ithinuel:
Closing in favor of #585

PR #571, opened 2023-03-18
Upgrade syn to version 2.0

jannic:
Right now this doesn't improve anything: While other dependencies still use syn version 1, it just causes both versions to be built, increasing the build time. So we might not yet want to merge this.

Once dependencies start to migrate to version 2, we should do so as well, to complete the migration to version 2 as quickly as possible.


ithinuel (approved):

jannic:
I checked today that our dependencies still use syn version 1, so there's no reason to merge this, yet.

PR #573, opened 2023-03-27
Re-export fugit

### LINKED ISSUE #572: Suggestion: Re-export foreign types from extern crates in public APIs
## Problem
Consider re-exporting foreign types in public APIs. For example, in https://github.com/rp-rs/rp-hal/blob/715c3c9dd85f4b30d949c310b60e7621ea62c3f6/rp2040-hal/src/timer.rs#L198 there is `MicrosDurationU32` from `fugit`.

The "problem" is: To use this API, I also have to add a compatible version `fugit` to my crate manually – which is obviously possible, but feels redundant because it splits my dependency tree.
```
Clean dependency tree: my_crate ─ rp2040_hal ─ fugit

Current dependency tree: my_crate ─ rp2040_hal ─ fugit
                                  ╲
                                    fugit
```

This might become especially cumbersome if my crate – for some reason – have to use another, possibly API-incompatible version of fugit:
```
Mixed dependency tree: my_crate ─ rp2040_hal ─ fugit-x
                                 ╲
                                   fugit-x
                                 ╲
                                   fugit-y
```

## Suggestion
A possible alternative would be to publicly re-export the used type, so that I could simply `use rp2040_hal::timer::{Alarm, Alarm0, MicrosDurationU32}` without having to synchronize my crate's imports.

jannic:
Closes #572

I also had a quick look for other crates to re-export:

One candidate would be `embedded-hal`, but I think a re-export is not appropriate there: That crate is, in general, not used to access features of `rp2040-hal`, but it's the other way around: Some driver expects a type implementing a trait from `embedded-hal`, and `rp2040-hal` provides that type. In that chase, the needed version of `embedded-hal` is selected by the driver, not by `rp2040-hal`, and a re-export would not help with that.

There are probably more, but it needs to be checked manually which dependencies are actually used in exposed APIs. I don't want to re-export something unnecessarily.

On other small change: I hid the re-export of `paste` from the docs. It's only there because it's used in macros provided by rp2040-hal. No need to list it in the docs.

### DIFF
--- a/rp2040-hal/examples/adc.rs
+++ b/rp2040-hal/examples/adc.rs
@@ -20,7 +20,7 @@ use rp2040_hal as hal;
 // Some traits we need
 use core::fmt::Write;
 use embedded_hal::adc::OneShot;
-use fugit::RateExtU32;
+use hal::fugit::RateExtU32;
 use rp2040_hal::Clock;
 
 // UART related types
--- a/rp2040-hal/examples/i2c.rs
+++ b/rp2040-hal/examples/i2c.rs
@@ -13,12 +13,12 @@
 // be linked)
 use panic_halt as _;
 
-// Some traits we need
-use embedded_hal::blocking::i2c::Write;
-use fugit::RateExtU32;
-
 // Alias for our HAL crate
 use rp2040_hal as hal;
+
+// Some traits we need
+use embedded_hal::blocking::i2c::Write;
+use hal::fugit::RateExtU32;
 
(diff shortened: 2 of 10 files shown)


ithinuel (approved):
I wasn't convinced by the utility of re-export in general but the explanation here makes total sense !
This is a visible implementation detail.
Thank you :)

jannic:
(I accidentally included #709 in this pull request - fixed with the force-push a few seconds ago.)

PR #575, opened 2023-03-29
Fix probe-run command in readme

jlpettersson:
The current command gives an error,
the `--chip` and `--list-probes` flags cannot be used at the same time.

This patch fixes the command in the readme.


jannic (approved):
Thanks, looks correct!

PR #577, opened 2023-03-31
Fix a broken link in docs

jlpettersson:
The code that the link refers to has been moved to another repo.

This patch fixes the link.


jannic (approved):
Thanks!

PR #578, opened 2023-04-01
Document RAM banks in memory.x

jannic:
This defines new memory regions SRAM4, SRAM5.

Without a linker section referring to those regions, they will not actually be used.


ithinuel (approved):

thejpster:
Should we add a note somewhere on how to use these banks, and how if you use `link-section`, the default cortex-m-rt pre-main code won't initialise variables located there and so should use `MaybeUninit`?

PR #580, opened 2023-04-05
Fix DMA IRQ issue for bidirectional DMA transfer

jlpettersson:
By reading the code, this seems to be a bug, caused by a typo. The function `check_irq0()` seems to be intended to check `irq0` for both channels. But for channel 1, it is set to check `irq1`.

The corresponding function `check_irq1()` does check `irq1` for both channels.


jannic (commented):
LGTM, good catch. This could have caused hours of debugging!
@9names, you know this code better than me, I'd prefer if you acked the fix as well before merging.

9names (approved):
This change looks good to me

PR #582, opened 2023-04-13
Update embedded-hal alpha dependencies

jannic:


jannic:
>  I have an #[inline] on kind (not sure if that makes sense to do or not).

I didn't check the generated code, so I don't know the actual effect of that attribute. In general, I try to not clutter the code with `#[inline]` hints and assume that the compiler doesn't need it to infer that inlining such short functions is a good idea. This is  how I understand the documentation at https://doc.rust-lang.org/reference/attributes/codegen.html#the-inline-attribute.

rp2040-hal/src/i2c/controller.rs
```diff
@@ -284,53 +284,14 @@ impl<T: Deref<Target = Block>, PINS> eh1::I2c for I2C<T, PINS, Controller> {
...
-        let addr: u16 = address.into();
-        Self::validate(addr, Some(peekable.peek().is_none()), None)?;
-        self.setup(addr);
-
-        for tx in peekable {
-            self.write_internal(&[tx], false)?
-        }
-        self.read_internal(buffer, true, true)
-    }
-
```
ithinuel, line 325: The main argument I've seen against them being "it is hard/suboptimal to implement with a dma".
I think those method do have value, even if the embedded-hal decided to drop them.

Can we make them inherent method instead of getting rid of them?

(the kind of nice thing that can be done with such an api can be seen [here](https://github.com/ithinuel/sh1107-rs/blob/main/sh1107/src/lib.rs#L165-L225)).

jannic: That's a good idea.

Should there only be inherent methods for functions missing from E-H, or should there be an inherent impl for everything, so people who don't use drivers based on e-h don't need to bring the e-h traits into scope?

ithinuel (approved):

PR #584, opened 2023-04-19
Update usb-device implementation.

ithinuel:
Next usb-device release has some changes around isochronous endpoint type.

(only to be merged once the new version of usb-device gets published)


jannic (approved):
LGTM. Perhaps you could also add the new MSRV to the changelog, so we don't forget it later?

PR #585, opened 2023-04-19
Experiment gpio overhaul

ithinuel:
It is mostly there.

Remaining tasks:

- [ ] Spi own the pins
- [ ] Provide a special `OdPin`  that emulates open-drain behaviour.


thejpster (commented):
I started reviewing this on my phone, but I'm going to need a real computer!

In general this looks great and  thank you for doing all this work. I'm excited to see it land.

I have some minor nits around the ordering of types and impl blocks, and missing module-level docs to explain what each file is for to the casual observer (including the pub-crate modules for sealed traits). I may try and add these myself instead of leaving comments, but not right now.

rp2040-hal/src/gpio/func.rs
```diff
@@ -0,0 +1,185 @@
...
+
+// =============================
+// Pin to function mapping
+
+/// Error type for invalid function conversion.
+pub struct InvalidFunction;
+
+/// Marker of valid pin -> function combination.
+///
+/// Read as `F is a valid function implemented for the pin F`
```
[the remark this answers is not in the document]
ithinuel: oh that's a good catch thanks!

rp2040-hal/src/gpio/mod.rs
```diff
@@ -127,3 +151,1224 @@ pub enum OutputOverride {
     /// Drive output high.
     AlwaysHigh = 3,
 }
+
+/// Pin with validation of [`Function`] and [`PullType`].
```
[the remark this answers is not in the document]
ithinuel: Thank you! I changed "pin mode" to "pull type" to avoid confusion with what the current "pin mode" is (effectively a mix of function & pull type).

rp2040-hal/src/adc.rs
```diff
@@ -37,20 +37,57 @@
...
+    /// Release the pin and restore its digital circuitery's state.
+    pub fn release(self) -> P {
+        let mut p = self.pin.into();
+        p.set_output_disable(self.output_disable);
+        p.set_input_enable(self.input_enable);
+        P::from(p)
+    }
+}
+
 /// Adc
```
thejpster, line 91:
```suggestion
/// Analog to Digital Convertor (ADC).
///
/// Represents an ADC within the RP2040. Each ADC has multiple channels, and each
/// channel is either associated with a specific GPIO pin or attached to the internal
/// temperature sensor. You should put the relevant pin into ADC mode by creating an
/// [`AdcPin`] object with it, or you can put the ADC into `Temperature Sensing Mode"
/// by calling [`Adc::enable_temp_sensor()`]. Either way, the resulting objects can be
/// passed to the [`OneShot::read()`] trait method to actually do the read.
```

rp2040-hal/src/adc.rs
```diff
@@ -92,11 +129,30 @@ impl Adc {
...
+
+        while !self.device.cs.read().ready().bit_is_set() {
+            cortex_m::asm::nop();
+        }
+
+        self.device.result.read().result().bits()
+    }
 }
 
 macro_rules! channel {
```
thejpster, line 150: I think "impl Trait for Type" should be moved to be under `struct Type`. Or possibly under the `trait Trait` block, depending on which makes more sense to the casual observer. But usually the former.

rp2040-hal/src/adc.rs
```diff
@@ -124,32 +187,49 @@ impl Channel<Adc> for TempSense {
...
-            cortex_m::asm::nop();
-        }
+/// The pin was invalid for the requested operation
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct InvalidPinError;
 
-        Ok(self.device.result.read().result().bits().into())
+// Implementation for dyn-pins
+impl<WORD, F, M> OneShot<Adc, WORD, AdcPin<Pin<DynPinId, F, M>>> for Adc
```
thejpster, line 214: The ADC provides a u16, so does it make sense to be generic over any type WORD, where WORD can be created from a u16? e.g. if someone asked for a u32 from the read function, they'd get it, but it would only be scaled 0..65535 not 0..2^32. I think?

rp2040-hal/src/gpio/func.rs
```diff
@@ -0,0 +1,185 @@
...
+
+/// Type-level `enum` for pin function.
+pub trait Function: func_sealed::Function {}
+
+/// Value-level `enum` for pin function.
+#[allow(missing_docs)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum DynFunction {
+    Xip,
```
thejpster, line 25: pub enums should have comments on each variant, perhaps with a link to the datasheet?

rp2040-hal/src/gpio/func.rs
```diff
@@ -0,0 +1,185 @@
...
+    const DYN: DynSioConfig = DynSioConfig::Input;
+}
+/// Type-level `variant` for SIO configuration.
+pub enum SioOutput {}
+impl SioConfig for SioOutput {
+    #[allow(missing_docs)]
+    const DYN: DynSioConfig = DynSioConfig::Output;
+}
+
+// =============================
```
thejpster, line 122: I like these kinds of dividers, especially in large complicated files. Happy to see them made even bigger :)

rp2040-hal/src/gpio/mod.rs
```diff
@@ -127,3 +151,1225 @@ pub enum OutputOverride {
...
+pub struct Pin<I: PinId, F: func::Function, P: PullType> {
+    id: I,
+    function: F,
+    pull_type: P,
+}
+
+/// Create a new pin instance.
+///
+/// # Safety
+/// The unicity of the pin is not verified. User must make sure no other instance of that specific
```
thejpster, line 166:
```suggestion
/// The uniqueness of the pin is not verified. User must make sure no other instance of that specific
```

rp2040-hal/src/gpio/mod.rs
```diff
@@ -127,3 +151,1225 @@ pub enum OutputOverride {
...
+            pin::set_pull_type(&self.id, new_pull_type);
+        }
+
+        Pin {
+            pull_type,
+            id: self.id,
+            function: self.function,
+        }
+    }
+    /// Erase the Pin ID type check.
```
thejpster, line 291: One line of whitespace between functions, please!

thejpster:
Sorry more scattered drive-by comments as they come to me.

jannic:
We should also have a look at the generated code.
As a (very) quick check, I ran `cargo bloat --release --example blinky` on this branch and on the base commit (`22121b5078ffe75d5f2c3b49cadc067dd5d9ecda`):

This branch:
```
File  .text   Size      Crate Name
0.1%  12.4%   886B        std compiler_builtins::int::specialized_div_rem::u64_div_rem
0.1%  11.6%   828B        std core::str::count::do_count_chars
0.1%  10.2%   732B rp2040_hal rp2040_hal::clocks::ClocksManager::init_default
0.1%   9.5%   682B        std core::fmt::Formatter::pad_integral
0.1%   9.2%   656B        std core::fmt::Formatter::pad
0.1%   7.0%   500B     blinky blinky::__cortex_m_rt_main
0.0%   5.9%   420B rp2040_hal rp2040_hal::pll::setup_pll_blocking
0.0%   5.9%   420B rp2040_hal rp2040_hal::pll::setup_pll_blocking
0.0%   5.4%   388B rp2040_hal rp2040_hal::gpio::pin::set_function
0.0%   4.8%   344B rp2040_hal rp2040_hal::clocks::init_clocks_and_plls
0.0%   3.5%   248B        std core::fmt::num::imp::<impl core::fmt::Display for u32>::fmt
0.0%   2.9%   208B   cortex_m cortex_m::delay::Delay::delay_us
0.0%   1.5%   104B rp2040_hal rp2040_hal::gpio::bank0::Pins::new
0.0%   1.1%    78B  [Unknown] __aeabi_lmul
0.0%   0.9%    62B        std core::fmt::Formatter::pad_integral::write_prefix
0.0%   0.8%    60B        std core::panicking::panic_bounds_check
0.0%   0.8%    60B        std core::panicking::unreachable_display
0.0%   0.6%    44B  [Unknown] __aeabi_llsl
0.0%   0.6%    44B        std core::panicking::panic
0.0%   0.6%    40B        std core::panicking::panic_fmt
0.0%   3.0%   218B            And 23 smaller methods. Use -n N to show more.
0.8% 100.0% 7.0KiB            .text section size, the file size is 883.2KiB
```

Base commit:
```
File  .text   Size       Crate Name
0.1%  22.2%   886B         std compiler_builtins::int::specialized_div_rem::u64_div_rem
0.1%  18.3%   732B  rp2040_hal rp2040_hal::clocks::ClocksManager::init_default
0.0%  10.5%   420B  rp2040_hal rp2040_hal::pll::setup_pll_blocking
0.0%  10.5%   420B  rp2040_hal rp2040_hal::pll::setup_pll_blocking
0.0%   9.6%   384B      blinky blinky::__cortex_m_rt_main
0.0%   8.6%   344B  rp2040_hal rp2040_hal::clocks::init_clocks_and_plls
0.0%   5.2%   208B    cortex_m cortex_m::delay::Delay::delay_us
0.0%   2.2%    88B  rp2040_hal rp2040_hal::gpio::pin::bank0::Pins::new
0.0%   2.0%    78B   [Unknown] __aeabi_lmul
0.0%   1.1%    44B   [Unknown] __aeabi_llsl
0.0%   1.1%    44B         std core::panicking::panic
0.0%   1.0%    40B         std core::panicking::panic_fmt
0.0%   1.0%    40B cortex_m_rt Reset
0.0%   0.6%    22B   [Unknown] __aeabi_uldivmod
0.0%   0.5%    20B   [Unknown] HardFaultTrampoline
0.0%   0.5%    18B         std __udivmoddi4
0.0%   0.4%    16B         std <T as core::any::Any>::type_id
0.0%   0.3%    10B    cortex_m __delay
0.0%   0.3%    10B   [Unknown] main
0.0%   0.2%     6B    cortex_m __primask_r
0.0%   0.8%    30B             And 11 smaller methods. Use -n N to show more.
0.4% 100.0% 3.9KiB             .text section size, the file size is 877.4KiB
```

Most of the size change seems to come from some additional formatting code. Probably some new panic somewhere, which pulls in a lot of stuff from core. While not  perfect, I don't think this is a big issue, as it won't significantly affect a larger binary. (If at all, because with some probability a larger binary would contain that formatting code anyway.)

But I think the change in size of `blinky::__cortex_m_rt_main` from 384B to 500B is a little bit worrying. When I find some time I'll have a look at the generated code to find out what changed there.

ithinuel:
There was a few missed optimisation oportunities that a few inline fixed :)
That being said, the rp-hal workspace does not have the common recommended profile.

I get the best results with
```toml
[profile.release]
codegen-units = 1
debug = 2
debug-assertions = false
incremental = false
lto = 'fat'
opt-level = "s"
overflow-checks = false
```

and `cargo bloat --release --example blinky --features 'cortex-m/inline-asm'`

I get the exact same 

```
File  .text   Size       Crate Name
0.3%  60.7%   872B      blinky blinky::__cortex_m_rt_main
0.0%  10.3%   148B    cortex_m cortex_m::delay::Delay::delay_us
0.0%   7.5%   108B  rp2040_hal rp2040_hal::clocks::fractional_div
0.0%   5.4%    78B   [Unknown] __aeabi_lmul
0.0%   2.8%    40B cortex_m_rt Reset
0.0%   1.4%    20B   [Unknown] HardFaultTrampoline
0.0%   0.7%    10B         std core::panicking::panic
0.0%   0.7%    10B         std core::panicking::panic_fmt
0.0%   0.7%    10B   [Unknown] main
0.0%   0.1%     2B cortex_m_rt HardFault_
0.0%   0.1%     2B  panic_halt rust_begin_unwind
0.0%   0.1%     2B cortex_m_rt DefaultPreInit
0.0%   0.1%     2B cortex_m_rt DefaultHandler_
0.0%   0.0%     0B             And 0 smaller methods. Use -n N to show more.
0.5% 100.0% 1.4KiB             .text section size, the file size is 297.6KiB
```

on both `main` and this branch on commit `ea49143b899df4ee66ac71d7b50d23f26c8ed073`

rp2040-hal/src/adc.rs
ithinuel: 🤷 That's how it was when I found it 🤷 I don't have opinions about it.
An ADC output may be 12 or 14bit and return the result on a u16 as 0..(2^12) and as far as I can remember it's not usually scaled up to a u16 so I guess one should not expect to get is scaled to a u32 even if that's the type they expect.

rp2040-hal/src/gpio/func.rs
ithinuel: I couldn't find something to write on each that would add anything valuable.

This seems pretty redundant:
```rust
pub enum DynFunction {
    /// Value-level `variant` for the Xip [`Function`].
    Xip,
    /// Value-level `variant` for the Uart [`Function`].
    Uart,
    // […]
}
```

jannic:
> There was a few missed optimisation oportunities that a few inline fixed

:+1:
Indeed, with the inline annotations, I get the same code size as with the current `main` branch. I hoped that the rust compiler was better at deciding when to inline function, even without annotations.

rp2040-hal/src/gpio/mod.rs
```diff
@@ -128,3 +152,1262 @@ pub enum OutputOverride {
...
+    /// Create a new wrapper
+    pub fn new(inner: T) -> InOutPin<T>
+    where
+        T::Id: ValidFunction<FunctionSioOutput>,
+    {
+        let mut inner = inner.into();
+        inner.set_output_enable_override(OutputEnableOverride::Disable);
+
+        // into Pin<_, FunctionSioOutput, _>
+        let inner = inner.into_push_pull_output_in_state(PinState::Low);
```
jannic, line 1285: I wonder if this is the only sensible way an InOutPin could work, and therefore, if the naming is appropriate:

In theory, it would be equally valid to have pin that alternates between active drive to high and disabled driver. That could be used together with a pull-down resistor. But I have no idea if there is any practical application of this configuration, or if everything uses pull-up resistors and drive the pin low actively.

rp2040-hal/src/gpio/mod.rs
```diff
@@ -128,3 +152,1262 @@ pub enum OutputOverride {
...
+        };
+        self.id.io_ctrl().modify(|_, w| w.oeover().variant(variant));
+    }
+
+    /// Set the output override.
+    #[inline]
+    pub fn set_output_override(&mut self, override_value: OutputOverride) {
+        use pac::io_bank0::gpio::gpio_ctrl::OUTOVER_A;
+        let variant = match override_value {
+            OutputOverride::DontInvert => OUTOVER_A::NORMAL,
```
jannic, line 547: I see it's not new, but perhaps that gpio rework is a good opportunity to change this as well: The naming `DontInvert` feels wrong. It's probably because it's a kind of double negation, and because output override can not only invert the pin. While thinking about a better name, I considered `Normal` - only later seeing that the SVD uses the same term. Is there any reason why `Normal` would not be a good choice?

rp2040-hal/src/gpio/mod.rs
ithinuel: I thought about the exact same thing as I was picking this from the dht example but I couldn't think of any usecase so I figured simpler is better and we can still eventually update this later on.

jannic: I did some web searching about this question. It seems like having pull-up resistors and an output driver pulling the line down actually had significant advantages over the opposite choice back in the days of TTL logic. While this wouldn't matter much now that everybody is using CMOS technology, the convention has been kept (why change habits for no good reason?) and everyone uses it.

So I think it's fine to only provide this variant for now. I'd prefer a slightly more descriptive naming, but don't have a good idea. `OpenDrain` wouldn't be exactly true. Perhaps `OpenDrainLike`? But that's ugly. If we don't find anything better, I'm fine with keeping the `InOutPin` name.

rp2040-hal/src/gpio/pull.rs
```diff
@@ -0,0 +1,57 @@
...
+pub enum DynPullType {
+    #[allow(missing_docs)]
+    None,
+    #[allow(missing_docs)]
+    Up,
+    #[allow(missing_docs)]
+    Down,
+    /// This enables both pull resistor. This setup is described as bus-keep in the RP2040
+    /// datasheet.
+    Both,
```
jannic, line 26: The bus-keep function is very different from "enable both pull-up and pull-down resistors". Just having two resistors enabled would work as a voltage divider, stabilizing the voltage on the line to some fixed value between 0V and 3.3V, as long as nothing else is driving the line. But bus-keep will pull the line to either 0V or 3.3V, depending on the initial voltage.

Therefore I think the name "Both" as well as the description "enables both pull resistor" are misleading. I'd prefer the naming from the data sheet: Just call it `BusKeep`

rp2040-hal/src/spi.rs
```diff
@@ -88,15 +92,17 @@ impl Sealed for u8 {}
 impl Sealed for u16 {}
 
 /// Spi
-pub struct Spi<S: State, D: SpiDevice, const DS: u8> {
+pub struct Spi<S: State, D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> {
```
jannic, line 95: With the new parameter `P`, how can I translate such a type alias to the new API?
```
type SpiBus = hal::Spi<hal::spi::Enabled, pac::SPI0, 8>
```
(Taken from https://github.com/Neotron-Compute/Neotron-Pico-BIOS/blob/develop/src/main.rs#L112)

ithinuel: It depends on whether you man the pinId to be compile time checked.
If not, you'd probably do something like:

```rust
type SpiPins = (
    Pin<DynPinId, FunctionSpi, DynPullType>, // MOSI
    Pin<DynPinId, FunctionSpi, DynPullType>, // MISO
    Pin<DynPinId, FunctionSpi, DynPullType>, // SCLK
);

type SpiBus = hal::Spi<hal::spi::Enabled, pac::SPI0, SpiPins, 8>;
```

ithinuel: Which makes me realize that I should've added a `try_new` function to runtime check the valid Pin Function on serial, spi and i2c like I did for Adc 🤔 .

I'll impl that asap (probably in the next couple of days).

rp2040-hal/src/adc.rs
```diff
@@ -124,20 +134,71 @@ impl Channel<Adc> for TempSense {
...
+        self.device.result.read().result().bits()
+    }
+
+    /// Enable temperature sensor, returns a channel to use.
+    ///
+    /// This can only be done once before calling [`disable_temp_sensor`]. If the sensor has already
+    /// been enabled, this method will panic.
+    pub fn enable_temp_sensor(&mut self) -> TempSense {
+        self.take_temp_sensor()
+            .expect("Temp sensor is already enabled.")
```
jannic, line 180: I'd prefer to avoid "hidden panics", functions that panic internally, and that don't have a name making that possibility obvious.

The various '.take()' methods, which are typically called once at boot and are obviously not allowed to be called multiple times are fine.

But otherwise, I consider a panic within HAL code a smell that should be avoided.

ithinuel: We could mark this method as deprecated and remove it in a future release. Do you think that's a good solution?

We may not need to worry too much about deprecation considering the amount of breaking changes anyway 🤷

rp2040-hal/src/spi.rs
ithinuel: I updated the PR and it will now be:

```rust
type SpiPins = (
    ValidatedPinTx<Pin<DynPinId, FunctionSpi, DynPullType>, SPI0>, // MOSI
    ValidatedPinTx<Pin<DynPinId, FunctionSpi, DynPullType>, SPI0>, // MISO
    ValidatedPinTx<Pin<DynPinId, FunctionSpi, DynPullType>, SPI0>, // SCLK
);

type SpiBus = hal::Spi<hal::spi::Enabled, pac::SPI0, SpiPins, 8>;
```

I'm not sure it is the most elegant but it does the trick.

rp2040-hal/src/gpio/mod.rs
```diff
@@ -128,3 +154,1299 @@ pub enum OutputOverride {
...
+    #[inline]
+    pub fn set_slew_rate(&mut self, rate: OutputSlewRate) {
+        self.id
+            .pad_ctrl()
+            .modify(|_, w| w.slewfast().bit(OutputSlewRate::Fast == rate));
+    }
+
+    /// Get wether the schmitt trigger (hysteresis) is enabled.
+    #[inline]
+    pub fn get_schimtt_enabled(&self) -> bool {
```
jannic, line 487: Typo.
```suggestion
    pub fn get_schmitt_enabled(&self) -> bool {
```

rp2040-hal/src/adc.rs
jannic: Both deprecation or just removing the function would be fine.

What I'd really like to have is some API that can't fail at all. There's no inherent reason why calling "enable_temp_sensor" twice should not be allowed. Things don't fail if someone calls `set_high()` twice on a GPIO pin either. But of course I understand the reasoning here. Allowing for multiple calls to this function, _and_ making sure that the `TempSense` token doesn't get used after the flag has been disabled, _and_ avoiding additional bookkeeping overhead. I don't know if there is a way to do all this without having a `panic` somewhere.

ithinuel: I added a deprecation note.

rp2040-hal/src/gpio/func.rs
thejpster: Well you could write them in the case that people expect:

```rust
/// Describes the function currently assigned to a pin with a dynamic type.
///
/// A 'pin' on the RP2040 can be connected to different parts of the chip
/// internally - for example, it could be configured as a GPIO pin and connected
/// to the SIO block, or it could be configured as a UART pin and connected to
/// the UART block.
pub enum DynFunction {
    /// The 'XIP' (or Execute-in-place) function, which means talking to the QSPI Flash
    Xip,
    /// The 'UART' (or 'serial-port') function.
    Uart
}
```

rp2040-hal/src/gpio/mod.rs
```diff
@@ -128,3 +154,1299 @@ pub enum OutputOverride {
     /// Drive output high.
     AlwaysHigh = 3,
 }
+
+/// Represents a pin, with a given ID (e.g. Xx), a given function (e.g. Uart) and a given pull type
```
jannic, line 158: "Xx"?

jannic:
> I think this is a necessary evil as, unlike most platforms, the default state for pins is not `floating input` but some sort of `inactive pull down`

How do these things relate? I don't disagree with the choice of having the pull config as part of the type state, but I don't see how it is affected by the default state for pins.

ithinuel:
It is because it forces users to pay attention to the pull ups/downs. Or at least it makes it more visible.

jannic:
> It does because it forces users to pay attention to the pull ups/downs. Or at least it makes it more visible.

I'm not sure I follow the reasoning. Using wrong pull up/down settings will have the same effect regardless of the default settings. And in many cases, having a weak pull-down is safer than a floating pin. (https://electronics.stackexchange.com/questions/7179/is-it-really-a-bad-idea-to-leave-an-mcu-input-pin-floating)

Anyways, for many low-speed applications, the three settings you chose to encode in the type are exactly the three required ones. Like you wrote in one of the comments, all the other settings are only needed for (slightly) more advanced setups. So I think the choice is a good one.

rp2040-hal/src/gpio/mod.rs
```diff
@@ -233,7 +233,7 @@ impl<I: PinId, F: func::Function, P: PullType> Pin<I, F, P> {
     }
 
     /// Convert the pin from one state to the other.
-    pub fn into<F2, P2>(self) -> Pin<I, F2, P2>
+    pub fn into_typestate<F2, P2>(self) -> Pin<I, F2, P2>
```
jannic, line 236: I stumbled upon this change while porting Neotron code to the new API.
It was not immediately obvious what `into()` needs to be replaced by. Perhaps there should be some migration guide?

At first, I thought that this is a bad change, as `.into()` is a common pattern and easily discoverable. But then, I wondered if it was the correct abstraction in the first place. `.into()` is for type conversion, and usually doesn't have any side effects. Is it an anti-pattern to use `.into()` like this? If yes, then we should also avoid names like `.into_typestate()`, and prefer something like `.set_typestate()`.

If HAL users have to fix all calls to `.into()` anyway, and we already lose the nicely discoverable name, we could as well switch to some "more correct" name. That leaves the `.into_...`-pattern for pure type conversions without side effects.

jannic: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv lists `as_`, `into_` and `to_` as common prefixes for type conversion functions. There's no explicit rule that forbids to use those names for something else.

What do other HALs do?

- esp-idf-hal works differently: https://github.com/esp-rs/esp-idf-hal/blob/master/examples/button.rs#L18-L19
- stm32f1xx-hal uses the `into_` prefix: https://github.com/stm32-rs/stm32f1xx-hal/blob/master/examples/blinky.rs#L40
- embassy works more like esp-idf-hal: https://github.com/embassy-rs/embassy/blob/master/examples/rp/src/bin/blinky.rs#LL15C1-L15C53
- nrf-hal uses the `into_`-prefix, https://github.com/nrf-rs/nrf-hal/blob/master/examples/blinky-button-demo/src/main.rs#L22-L23

So using `into_` is widespread, but not universal.

ithinuel: According to the API guidelines, the prefix should probably be `to_`:

> Conversions prefixed `to_`, on the other hand, typically stay at the same level of abstraction but do some work to change from one representation to another.

In our case it definitely stays the same level of abstraction. It's the same generic type with different parameters.

jannic: The three prefixes are meant for functions which do pure type conversions (which might still be expensive, for example when converting strings), but don't perform other actions.

But the GPIO methods also configure the hardware. And that's not just some detail, I'd argue it's the main point of the call. That it also returns a different data type is nice, because it allows to detect some errors at compile time. But it's not the reason the method is called in the first place.

Therefore, IMHO, the functions should have names that don't start with one of these prefixes.

BTW, `.into_typestate()` (and `.set_typestate()`, as I suggested above) don't reflect the intention of the user at all, regardless of the prefix. The actual function is more like `.reconfigure_by_typestate()`  - but that's quite cumbersome.
What about `.reconfigure()`, `.reconfigure_pull_type()`, `.reconfigure_function()`?

Notably, `.into_dyn_pin()` is fine, or could be replaced by `.as_dyn_pin()`, as it really only changes the type, but doesn't change the pin itself.

Of course, we might still decide to keep the `into_` prefix for consistency with several other HALs. Especially as many of the functions, like `.into_pull_up_input()` are quite good at discoverability and readability. But I really worry about `.into_typestate()`. It's both difficult to find if you do not yet know the name of the function, and difficult to understand if you read the code.

ithinuel: I wouldn't consider the pin's state separate from the type.
That's why we type state it and all the effort to guaranty the unicity of the pin IDs.

So we are effectively converting (with some reconfiguration cost) a pin from one state to another. This matches the type conversion of `into_` for owned to owned (non-copy) types.

ithinuel: Leaving this here as I'm reading them:
https://willcrichton.net/rust-api-type-patterns/state_machines.html
https://docs.rust-embedded.org/book/static-guarantees/state-machines.html

ithinuel: > I stumbled upon this change while porting Neotron code to the new API.
It was not immediately obvious what into() needs to be replaced by. Perhaps there should be some migration guide?

We could keep into and mark it `#[deprecated]` to help transition to the new API.

> At first, I thought that this is a bad change, as .into() is a common pattern and easily discoverable. But then, I wondered if it was the correct abstraction in the first place. .into() is for type conversion, and usually doesn't have any side effects. Is it an anti-pattern to use .into() like this? If yes, then we should also avoid names like .into_typestate(), and prefer something like .set_typestate().

Since the type is a representation of the state of the Pin, it don't consider the configuration of the hardware on type transition as a side effect. It's the actual purpose of these type-states.

> If HAL users have to fix all calls to .into() anyway, and we already lose the nicely discoverable name, we could as well switch to some "more correct" name. That leaves the .into_...-pattern for pure type conversions without side effects.

I couldn't find a recurrent pattern for type-state transitions. They usually use verbs corresponding to the actual transition.
I couldn't think of alternative names that would express these better.
- `set_function<F>()`/`set_pull_type<P>()`/`set_state<F, P>()`
- `into_function<F>()`/`into_pull_type<P>()`/`into_state<F, P>()`
- `to_function<F>()`/`to_pull_type<P>()`/`to_state<F, P>()`
- ?

thejpster: I just wanted to jump in and say: I enjoyed reading the discussion above. Congratulations to all for Being Nice On The Internet :)

jannic (approved):
As mentioned on matrix: I didn't do a full review, but I read through most changes (multiple times), ported some code to the changed API, and didn't see any obvious issues. AFAIK others did the same. So in sum, this change should be sufficiently reviewed to be merged.

PR #587, opened 2023-04-21
Activate clippy checks on exported APIs

jannic:
Today I learned about the `avoid-breaking-exported-api = false` clippy config.

https://doc.rust-lang.org/stable/clippy/lint_configuration.html?highlight=avoid-breaking-exported-api#avoid-breaking-exported-api

While a clippy warning probably is not a good justification for a breaking change, such warnings can be useful to avoid issues in new APIs.

The number of issues clipply flags for existing APIs is managable, so I propose adding that config, and add matching allow annotations to override warnings in the current API.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,1 @@
+avoid-breaking-exported-api = false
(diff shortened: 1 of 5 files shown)


jannic:
CI checks failed because of https://github.com/lalrpop/lalrpop/issues/750

ithinuel (approved):

PR #588, opened 2023-04-27
Improve testing

ithinuel:
Use `cargo-hack` to save some maintenance effort on CI/test maintenance by auto-discovering features and optional deps.
Use `cargo-install` to benefit from its caching feature and shorten CI run time.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -13,7 +13,7 @@ jobs:
         uses: baptiste0928/cargo-install@v2
         with:
           crate: cargo-hack
-      - name: Build rp2040-hal's examples
+      - name: Build rp2040-hal's workspace (without the examples)
         run: cargo hack build --optional-deps --each-feature
   examples-builds:
     name: Build rp2040-hal's examples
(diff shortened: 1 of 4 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -2,52 +2,98 @@ on: [push, pull_request]
...
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
         with:
           target: thumbv6m-none-eabi
+      - name: Install cargo-hack
+        uses: baptiste0928/cargo-install@v2
+        with:
+          crate: cargo-hack
       - name: Build rp2040-hal's examples
```
jannic, line 16: I *think* this should be called "Build workspace". Examples are built in a different job, below.

ithinuel: Good catch, thanks!

ithinuel:
The pipeline failure is due to: https://github.com/est31/cargo-udeps/issues/154

jannic (approved):

PR #590, opened 2023-05-01
Added config options for byteswapping

Gip-Gip:


ithinuel (commented):
Thank you that is a nice addition.

I left some suggestion but feel free do disregard if you think they are not useful.

jannic:
Definitely a nice addition!

As this is a (semver-)breaking change: From a quick glance over the past commits on the main branch, we didn't have a breaking change yet, since version 0.8.0. Shall we publish a 0.8.1 before merging this?
(Of course, we could alternatively create a branch for further 0.8.x releases.)

Issue #592, opened 2023-05-04
Make `bad::Never` public


KizzyCode:
Please consider to expose [`bad::Never` from `multicore.rs`](https://github.com/rp-rs/rp-hal/blob/c6c563fc1ca261f404e2e9e4fda924f55023a95b/rp2040-hal/src/multicore.rs#L260).

## Reasoning
This type is used in a public function signature ([`Core::spawn`](https://github.com/rp-rs/rp-hal/blob/c6c563fc1ca261f404e2e9e4fda924f55023a95b/rp2040-hal/src/multicore.rs#L147)). If the function is called directly, this is not a problem. However if I – for some reason – want to write a wrapper around `Core::spawn`, my wrapper cannot enforce the same trait restrictions since the type is private and thus cannot call `Core::spawn`.

There is a workaround (reimplement `bad::Never` in your own crate), but that's not very pretty IMO 😅

ithinuel:
Since https://crates.io/crates/bad supports `no_std` couldn't we use it directly ? (and extend its use to all the `type Error` are are infallible?)

jannic:
Opposite suggestion: Assuming that there are good reasons why the never type is not yet stable, shouldn't we try to avoid using this hack at all? The spawned function could return () instead, and the spawn method could loop after calling it, for example.

ithinuel:
Haha @jannic I did this alternative PR before seen your comment :)

There's a little question left as a TODO in the PR.
When the closure returns we recover the responsibility for the interrupts etc. Should we try devise something to disable them (not sure it's possible to enforce anything reliably there), or document that the closure MUST return with the system back into a neutral state (nvic and other core specific blocks to their reset state) ?

jannic:
Also see https://github.com/rp-rs/rp-hal/pull/593#issuecomment-1534234377

(This discussion is distributed over too many tickets)

PR #593, opened 2023-05-04
use bad::Never instead of core::convert::Infallible (or re-impl of bad::Never)

ithinuel:
Addresses: #592

### DIFF
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -38,6 +38,7 @@ chrono = { version = "0.4", default-features = false, optional = true }
 defmt = { version = ">=0.2.0, <0.4", optional = true }
 
 rtic-monotonic = { version = "1.0.0", optional = true }
+bad = { version = "0.1.1", default-features = false }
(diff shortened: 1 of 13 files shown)


jannic (commented):
Please don't do that. The word `bad` is there for a reason, and using `Infallible` is a common, easy to understand convention.

ithinuel:
That was an interesting exercise, but I agree we should not merge this PR!

Should we remove the use of never in multicore though?

jannic:
> Should we remove the use of never in multicore though?
Not sure about this. If the never type was stable, it would be the correct type to use. And the reasons why the never type isn't stable yet don't apply here, afaik. So the hack is not exactly wrong. But it's still a hack, it makes the code more difficult to understand, and it can cause issues as seen in #592.
I think I wouldn't introduce bad::Never today now if it wasn't already there. But removing it is a breaking change, with little gain, isn't it?
(Or is the fact that Never can be coerced to any type sufficient to make the change non-breaking?)

PR #594, opened 2023-05-04
Remove the requirement for multicore tasks to never return

ithinuel:
Alternative solution to #592

### DIFF
--- a/rp2040-hal/src/multicore.rs
+++ b/rp2040-hal/src/multicore.rs
@@ -12,7 +12,7 @@
 //!
 //! static mut CORE1_STACK: Stack<4096> = Stack::new();
 //!
-//! fn core1_task() -> ! {
+//! fn core1_task() {
 //!     loop {}
 //! }
 //!
@@ -144,6 +144,12 @@ impl<'p> Core<'p> {
     }
 
     /// Spawn a function on this core.
+    ///
+    /// Upon exit of the entry function, the core will enter an infinite `wfe` (wait for event)
+    /// loop. In order to reduce power consumption, it is recommended that all interrupts and events
+    /// are disabled before returning.
+    ///
+    /// Core 1 will need to be reset from core 0 in order to spawn another task.
     pub fn spawn<F>(&mut self, stack: &'static mut [usize], entry: F) -> Result<(), Error>
     where
         F: FnOnce() + Send + 'static,
@@ -179,6 +185,10 @@ impl<'p> Core<'p> {
             }
 
             // Reset the core
+            // TODO: resetting without prior check that the core is actually stowed is unsafe.
+            // But there does not seem to be any obvious way to check that. A marker flag could be
+            // set from this method and cleared for the wrapper after `entry` returned. But doing
+            // so wouldn't be zero cost.
             psm.frce_off.modify(|_, w| w.proc1().set_bit());
             while !psm.frce_off.read().proc1().bit_is_set() {
                 cortex_m::asm::nop();
--- a/rp2040-hal/src/multicore.rs
+++ b/rp2040-hal/src/multicore.rs
@@ -149,6 +149,11 @@ impl<'p> Core<'p> {
     /// stable.
     ///
     /// Core 1 will be reset from core 0 in order to spawn another task.
+    ///
+    /// Resetting a single core of a running program can have undesired consequences. Deadlocks are
+    /// likely if the core being reset happens to be inside a critical section.
+    /// It may even break safety assumptions of some unsafe code. So, be careful when calling this method
+    /// more than once.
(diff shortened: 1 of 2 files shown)


jannic:
This is a breaking change, as proven by the failing doc test, https://github.com/rp-rs/rp-hal/actions/runs/4880039111/jobs/8707219403?pr=594

In the end, it's a tradeoff. 

Pro:
- simplifies the code
- voids exposing the Never type (assuming it's still unstable for good reasons)

Cons:
- breaking change
- the Never type better describes the intention

Any arguments I missed?

rp2040-hal/src/multicore.rs
```diff
@@ -170,10 +176,19 @@ impl<'p> Core<'p> {
                 let mut sio = Sio::new(peripherals.SIO);
                 sio.fifo.write_blocking(1);
 
-                entry()
+                entry();
+                // TODO: should this bring the core back to its startup state waiting for a
+                // function? and/or reset its NVIC and other core specific blocks.
```
jannic, line 181: I wouldn't add any reset logic here. I don't see a good use case.

rp2040-hal/src/multicore.rs
```diff
@@ -170,10 +176,19 @@ impl<'p> Core<'p> {
...
+                loop {
+                    cortex_m::asm::wfe()
+                }
             }
 
             // Reset the core
+            // TODO: resetting without prior check that the core is actually stowed is unsafe.
```
jannic, line 188: Is it unsafe in the sense that it breaks any of rust's guarantees? Or just a bad idea, for example because it might cause deadlocks if the other core is inside a critical section when it gets reset?

rp2040-hal/src/multicore.rs
```diff
@@ -144,15 +144,21 @@ impl<'p> Core<'p> {
     }
 
     /// Spawn a function on this core.
+    ///
+    /// Upon exit of the entry function, the core will enter an infinite `wfe` (wait for event)
+    /// loop. In order to reduce power consumption, it is recommended that all interrupts and events
+    /// are disabled before returning.
```
jannic, line 150: Alternative idea: Just say that the function shouldn't return, and it's only defined differently because the never type isn't stable yet.
(But documenting the infinite loop is fine as well.)

rp2040-hal/src/multicore.rs
ithinuel: I think it is unsafe because it can cause cases "poisoning" (of data and peripherals) without a way to detect it.
As far as I'm aware, threads in std cannot be killed nor recycle so I'm not sure about the entire scope of the consequences but I definitely "feels wrong" to just kill it.

That being said, I am not aware of a way to tell if core1 is stowed so this would require some bookkeeping in the hal to:
- avoid a new spawn while the first one hasn't gracefully shut down
- capture panics and clear/mark the book to allow graceful recovery.

I'm not even sure the latter can be done.

jannic: Isn't resetting the core very similar to a theoretical situation where the core gets stuck in some infinite loop, and a new thread gets started on another core? I don't see how this is unsafe. Of course it can cause deadlocks in case the core had locks on some resources which now wouldn't get released. But again, that's not different from an infinite loop inside a critical section. Definitely not a good idea, but still not unsafe.

ithinuel: I see what you mean. My choice of word isn't great there indeed.
It is not `unsafe` in the rust sense, but it is a `bad` thing to do.

I'm removing this comment to avoid confusion.

jannic: After further thinking about it, I must admit that reasoning about the implications of such a reset is really tricky. 

While I was not yet able to construct something which would trigger obviously undefined behavior, I'm far from confident that it's really sound.

rp2040-hal/src/multicore.rs
```diff
@@ -144,15 +144,20 @@ impl<'p> Core<'p> {
     }
 
     /// Spawn a function on this core.
+    ///
+    /// The closure should not return. It is currently defined as `-> ()` because `-> !` is not yet
+    /// stable.
+    ///
+    /// Core 1 will be reset from core 0 in order to spawn another task.
```
jannic, line 151:
```suggestion
    /// Core 1 will be reset from core 0 in order to spawn another task.
    ///
    /// Resetting a single core of a running program can have undesired consequences. Deadlocks are
    /// likely if the core being reset happens to be inside a critical section.
    /// It may even break safety assumptions of some unsafe code. So, be careful when calling this method
    /// more than once.
```

rp2040-hal/src/multicore.rs
jannic: I think this discussion is not related to the merge request itself. We don't introduce a regression. The safety of resetting a running core can be discussed separately.

jannic (approved):

PR #600, opened 2023-05-06
Ensure supported DMA word size at compile time

jannic:
Replace a runtime check by a compile time check.

The generated error message is not perfect, but should be sufficient,
given that this should be a rare case:
    
    error[E0080]: evaluation of `rp2040_hal::dma::single_channel::IsValidWordSize::<u64>::OK` failed
     --> rp-hal/rp2040-hal/src/dma/single_channel.rs:148:18
      |
      = note: the evaluated program panicked at 'Unsupported DMA word size', rp-hal/rp2040-hal/src/dma/single_channel.rs:148:18
      |
      = note: this error originates in the macro `$crate::panic::panic_2015` which comes from the expansion of the macro `panic` (in Nightly builds, run with -Z macro-backtrace for more info)
    
    note: the above error was encountered while instantiating `fn <rp2040_hal::dma::Channel<rp2040_hal::dma::CH0> as rp2040_hal::dma::single_channel::ChannelConfig>::config::<u64, &mut [u64; 10], &mut [u64; 10]>`
     --> rp-hal/rp2040-hal/src/dma/single_buffer.rs:52:9
    
    For more information about this error, try `rustc --explain E0080`.
    error: could not compile `rp2040-hal-on-target-tests` due to previous error

I'd like to have a doc-test with `compile_fail`. Unfortunately it seems like doc tests are not run for no_std crates? I didn't manage to get such a test executed.


ithinuel (approved):

jannic:
BTW, this fixed an actual bug: Before, one could start a DMA transfer for such a type:
```
 pub struct U24(pub [u8; 3]);                                                             
                                                                               
 unsafe impl embedded_dma::Word for U24 {}                                     
```
That size of 3 would completely break the calculations in `ChannelConfig::config()`

PR #603, opened 2023-05-07
Byteswapping config

Gip-Gip:
Just rearranged my fork so I can add more features through different branches, sorry about having to re-create my PR.


jannic (approved):
LGTM, and I think the review comments from #590 were included.
As we did a minor release (0.8.1) a few days ago, I'd be fine to start applying breaking changes now. In case we need to do another minor release before 0.9.0, we can always do it from a separate branch.

PR #608, opened 2023-05-14
Abort early if bank0 is in reset in rp2040-e5 workaround

jannic:
See https://github.com/rp-rs/rp-hal/pull/607#issuecomment-1546838167 for a reasoning for this patch. Only a suggestion, not ready for merging.


jannic:
Integrated into #607

PR #610, opened 2023-05-15
Implement embedded-hal's DelayUs and DelayMs traits

ithinuel:


jannic:
Would it be possible to keep a `&Timer` inside that `Delay`?
That would have two advantages:

- you could use `get_counter` and would avoid the `unsafe` code.
- it would properly reference the lifetime of `Timer`, so it wouldn't be possible to accidentally destroy the `Timer`. That doesn't make a difference right now as the `Timer` doesn't get un-initialized on drop, but perhaps we want to change that some day?

ithinuel:
IIRC, Having the timetime in the type was a burden on CountDown when used in RTIC.
Although it can be "paused" the timer cannot be stopped or configured, it is the clock block that holds all the magic, so there's nothing to un-initialize really.
So making count down take a reference to Timer might have been a mistake I made :D

jannic:
> Although it can be "paused" the timer cannot be stopped or configured, it is the clock block that holds all the magic, so there's nothing to un-initialize really.

You're probably right. While the timer is a separate peripheral with its own reset bit, I don't see a use case for actually resetting it during normal operation.

rp2040-hal/src/timer.rs
```diff
@@ -106,6 +112,63 @@ impl Timer {
...
+            // use nop instead of wfe because there is no event configured to wake us up.
+            cortex_m::asm::nop();
+        }
+    }
+
+    fn delay_from_u64(&self, us: u64) {
+        // safety: only reads are made
+        let timer = unsafe { &*pac::TIMER::PTR };
+        // wrapping_add isn't strictly necessary here, there's no way we'll reach the wrap point of
+        // that u64, right?
```
jannic, line 132: It may not be useful, but `delay_us(u64::MAX)` will probably reach that wrap point.

Which makes me wonder: Is it completely unrealistic that some code might do that? Shouldn't that just block forever (for all practical matters), instead of returning immediately?

→ perhaps it would be better to use `saturating_add`?

rp2040-hal/src/timer.rs
```diff
@@ -106,6 +112,63 @@ impl Timer {
...
 unsafe impl Sync for Timer {}
 
+/// [`Timer`] based Delay
+pub struct Delay(());
+impl Delay {
+    fn delay_from_u32(&self, us: u32) {
+        // safety: only reads are made
+        let timer = unsafe { &*pac::TIMER::PTR };
+        let eta = timer.timelr.read().bits().wrapping_add(us);
+        while eta >= timer.timelr.read().bits() {
```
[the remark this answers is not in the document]
ithinuel: Ah that's right! I always get caught by this 🤦 
This newer commit should fix the issue.

rp2040-hal/src/timer.rs
ithinuel: using the `wrapping_sub` method should avoid the issue all together :D

jannic:
Let me suggest an alternative: https://github.com/rp-rs/rp-hal/pull/614

The separate `Timer` and `Delay` types feel artificial and don't provide any additional function.

rp2040-hal/src/timer.rs
```diff
@@ -106,6 +112,61 @@ impl Timer {
...
+    }
+}
+impl embedded_hal::blocking::delay::DelayMs<u32> for Delay {
+    fn delay_ms(&mut self, ms: u32) {
+        for _ in 0..ms {
+            self.delay_from_u32(1000);
+        }
+    }
+}
+impl embedded_hal::blocking::delay::DelayMs<u64> for Delay {
```
jannic, line 155: Implementing the trait for both u32 and u64 can cause a very strange error message:
```
error[E0277]: the trait bound `Delay: DelayMs<i32>` is not satisfied
  --> rp2040-hal/examples/blinky.rs:85:24
   |
85 |         delay.delay_ms(500);
   |               -------- ^^^ the trait `DelayMs<i32>` is not implemented for `Delay`
   |               |
   |               required by a bound introduced by this call
   |
   = help: the following other types implement trait `DelayMs<UXX>`:
             <Delay as DelayMs<u32>>
             <Delay as DelayMs<u64>>
```
This happens because rust doesn't know the exact type of the literal number `500`: All it knows it that it must be a number. Usually, that gets resolved automatically by type inference. But in this case, inference would be ambiguous: Either `u32` and `u64` could be valid. In that case, rust falls back to its default type: `i32`. Unfortunately, that's not a valid type for `DelayMs`, leading to the error message.

Of course, this can easily be fixed by calling `delay_ms(500u32)`. But that's a little bit ugly, and, more importantly, difficult to discover from the error message.

ithinuel: Is there any recommendations on embedded-hal 0.2.7 w.r.t. this ? (I'll do an extra search for this)

jannic: I don't know if there are recommendations, but I'm sure there are reasons e-h 1.0 went with only u32.
However, a quick search didn't bring up much besides "less code duplication" and "a blocking delay should never be >1h, so why care about u64?".

jannic: This is what cortex-m does to avoid this paper cut: https://docs.rs/cortex-m/0.7.7/src/cortex_m/delay.rs.html#85

jannic: Uh, and another issue: While playing with https://github.com/rust-embedded-community/embedded-sdmmc-rs/pull/80 I found this usage of `DelayUs<u8>`: https://github.com/rust-embedded-community/embedded-sdmmc-rs/pull/80/files#diff-cb7954c75c0cda9068329334e554ea5ce46a5ec8a40692a40be0e760f8c31e9fR37

If such things are common, we should implement the traits for all reasonable types, even if it causes a lot of duplication. Like cortex-m does.

jannic: A code search on github actually found many instances of `DelayUs<u8>` and `DelayUs<u16>` that were not just implementations of those traits. Interestingly, nobody uses `DelayUs<u64>`.

ithinuel: I added support for u8 to u64 and the i32 work around.
Do you think we should also cover for all the other signed integers ?

jannic: No, signed types do not make sense here. Even i32 is very questionable, but it's necessary because rusts fallback to that type if it can't infer the exact type of an integer.

rp2040-hal/src/timer.rs
```diff
@@ -106,6 +112,75 @@ impl Timer {
...
+        })*
+    };
+}
+
+impl_delay!(u32: u8, u16, u32);
+impl embedded_hal::blocking::delay::DelayUs<u64> for Delay {
+    fn delay_us(&mut self, us: u64) {
+        use embedded_hal::blocking::delay::DelayMs;
+
+        let (ms, us) = (us / 1_000, us % 1000);
```
jannic, line 150: According to figure 2 on page 26 of https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf, a 64 bit division using the hardware divider takes about 100 cycles. (Assuming that's the unit used in the graph.)
That's 0.8µs at 125MHz. Even if the compiler is smart enough to merge the division and the modulo operation into a single division, the CPU actually runs at 125MHz, and the hardware divider is used, this means that a 1µs delay becomes significantly longer than needed.
Given that the timer has a 64bit counter, why do you do this division at all? Is it just to avoid overflow issues in the calculations?

jannic: Good points! I actually completely missed that his implementation is only used for u64. Should be more careful during code review.
While it's theoretically possible to use the u64 version to do a 1µs delay, performance-sensitive callers would probably use a more suitable type and won't be affected by the division.

jannic: I had a look at the generated assembly. While it doesn't matter at all whether the parameter of the inner function is an u32 or an u64 (the call gets inlined anyway), doing the subtraction and comparison math in u32 saves several instructions.

rp2040-hal/src/timer.rs
```diff
@@ -106,6 +112,75 @@ impl Timer {
...
 unsafe impl Sync for Timer {}
 
+/// [`Timer`] based Delay
+pub struct Delay(());
+impl Delay {
+    fn delay_from_u32(&self, us: u32) {
+        // safety: only reads are made
+        let timer = unsafe { &*pac::TIMER::PTR };
+        let start = timer.timelr.read().bits();
+        while timer.timelr.read().bits().wrapping_sub(start) <= us {
```
jannic, line 122: There is a slight risk that this misses the deadline and waits for too long:
If `us` is `u32::MAX`, the condition `now.wrapping_sub(start) <= us` is true for only 1µs. It is not unthinkable that the loop gets slowed down by an interrupt sufficiently to miss that moment, leading to another hour of sleep.

So code like this might break:
```
  initialize_interrups(); // interrupts will do all the heavy lifting
  loop {
    delay_us(u32::MAX); // sleep ~1hour
    do_some_hourly_maintenance();
  }
```

jannic: I implemented a version that shouldn't have this issue in https://github.com/rp-rs/rp-hal/pull/614/commits/6aef40d957c6925e8039a97c50c7d0e0868297c1

While it looks a little bit ugly, it get optimized very well. For the common case of a constant delay, it gets compiled like this:
```
fn do_delay(delay: &mut dyn DelayUs<u32>) {
    delay.delay_us(500000);
}
```
⇓
```
rp2040_project_template::do_delay:
        ldr r0, .LCPI1_0 # address of the timelr register
        ldr r2, [r0]     # start = timelr
        ldr r1, .LCPI1_1 # delay value (500000)
.LBB1_1:
        ldr r3, [r0]     # now = timelr
        subs r2, r3, r2  # waited = now - start
        subs r1, r1, r2  # delay = delay - waited
        mov r2, r3       # started = now
        bhi .LBB1_1      # loop if subs resulted in a positive value
        bx lr
```

ithinuel:
Closing in favour of #614

PR #612, opened 2023-05-16
multicore: Fix undefined behaviour in multicore

ithinuel:
See these embassy PR for more explanation:
- https://github.com/embassy-rs/embassy/pull/1459
- https://github.com/embassy-rs/embassy/pull/1460

### DIFF
--- a/rp2040-hal/src/multicore.rs
+++ b/rp2040-hal/src/multicore.rs
@@ -155,12 +155,15 @@ impl<'p> Core<'p> {
             extern "C" fn core1_startup<F: FnOnce() -> bad::Never>(
                 _: u64,
                 _: u64,
-                entry: &mut ManuallyDrop<F>,
+                entry: *mut ManuallyDrop<F>,
                 stack_bottom: *mut usize,
             ) -> ! {
                 core1_setup(stack_bottom);
 
-                let entry = unsafe { ManuallyDrop::take(entry) };
+                let entry = unsafe { ManuallyDrop::take(&mut *entry) };
+
+                // make sure the preceding read doesn't get reordered past the following fifo write
+                compiler_fence(Ordering::SeqCst);
 
                 // Signal that it's safe for core 0 to get rid of the original value now.
                 //
@@ -181,20 +184,28 @@ impl<'p> Core<'p> {
             psm.frce_off.modify(|_, w| w.proc1().clear_bit());
 
             // Set up the stack
-            let mut stack_ptr = unsafe { stack.as_mut_ptr().add(stack.len()) };
+            // AAPCS requires in 6.2.1.2 that the stack is 8bytes aligned., we may need to trim the
+            // array size to guaranty that the bottom of the stack (the end of the array) meets that requirement.
+            // The start of the array does not need to be aligned.
+
+            let mut stack_ptr = stack.as_mut_ptr_range().end;
+            // on rp2040, usize are 4 bytes, so align_offset(8) on a *mut usize returns either 0 or 1.
+            let miss_alignement_offset = stack_ptr.align_offset(8);
 
             // We don't want to drop this, since it's getting moved to the other core.
             let mut entry = ManuallyDrop::new(entry);
 
             // Push the arguments to `core1_startup` onto the stack.
             unsafe {
+                stack_ptr = stack_ptr.sub(miss_alignement_offset);
+
                 // Push `stack_bottom`.
                 stack_ptr = stack_ptr.sub(1);
                 stack_ptr.cast::<*mut usize>().write(stack.as_mut_ptr());
 
                 // Push `entry`.
                 stack_ptr = stack_ptr.sub(1);
-                stack_ptr.cast::<&mut ManuallyDrop<F>>().write(&mut entry);
+                stack_ptr.cast::<*mut ManuallyDrop<F>>().write(&mut entry);
             }
 
             // Make sure the compiler does not reorder the stack writes after to after the
--- a/rp2040-hal/src/multicore.rs
+++ b/rp2040-hal/src/multicore.rs
@@ -52,7 +52,7 @@ pub enum Error {
 }
 
 #[inline(always)]
-fn install_stack_guard(stack_bottom: *mut usize) {
+fn install_stack_guard(stack_limit: *mut usize) {
     let core = unsafe { pac::CorePeripherals::steal() };
 
     // Trap if MPU is already configured
@@ -62,7 +62,7 @@ fn install_stack_guard(stack_bottom: *mut usize) {
 
     // The minimum we can protect is 32 bytes on a 32 byte boundary, so round up which will
     // just shorten the valid stack range a tad.
-    let addr = (stack_bottom as u32 + 31) & !31;
+    let addr = (stack_limit as u32 + 31) & !31;
     // Mask is 1 bit per 32 bytes of the 256 byte range... clear the bit for the segment we want
     let subregion_select = 0xff ^ (1 << ((addr >> 5) & 7));
     unsafe {
@@ -78,8 +78,8 @@ fn install_stack_guard(stack_bottom: *mut usize) {
 }
 
 #[inline(always)]
-fn core1_setup(stack_bottom: *mut usize) {
-    install_stack_guard(stack_bottom);
+fn core1_setup(stack_limit: *mut usize) {
+    install_stack_guard(stack_limit);
     // TODO: irq priorities
 }
 
@@ -156,9 +156,9 @@ impl<'p> Core<'p> {
                 _: u64,
                 _: u64,
                 entry: *mut ManuallyDrop<F>,
-                stack_bottom: *mut usize,
+                stack_limit: *mut usize,
             ) -> ! {
-                core1_setup(stack_bottom);
+                core1_setup(stack_limit);
 
                 let entry = unsafe { ManuallyDrop::take(&mut *entry) };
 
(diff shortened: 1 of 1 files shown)


jannic:
Just copying this from the embassy matrix room, https://matrix.to/#/!YoLPkieCYHGzdjUhOK:matrix.org/$nUUDUP2f7deY8ZfCuRl6zWal4A37A1T9qhxxP5WWtxg?via=matrix.org&via=grusbv.com&via=mozilla.org:

I think the issue is that the copy is just optimized away. So there's nothing left to be reordered.
Given this line (added pseudo-code type-annotations for clarity):

```
let entry: F = unsafe { ManuallyDrop::take(entry: &mut ManuallyDrop<F>) };
```

Semantically, the compiler needs to copy the value F from within the ManuallyDrop. But the compiler is smart. It knows that the original entry, inside the ManuallyDrop, can't go away, because it has a &mut, and &mut guarantees `noalias`. So why do the copy at all? This line is guaranteed to have no visible effect. So the compiler just optimizes the copy away entirely. At that point, no fence can safe you.
Why do I care? Well even if this is technically UB and one could stop thinking at that point, I wanted an explanation on how this happens. What optimization can make the code fail? And the compiler doesn't actively scan the code for UB to make us suffer. So it doesn't "know" that there's another reference to that memory in another function. Especially as the interface between the two functions is an `extern "C"`, and there is no actual function call that could be inlined. The compiler just doesn't know that `core1_startup` and `spawn_core1` run at the same time. So how can the UB trigger a miscompilation? I'm not satisfied with "well it was UB to begin with".
But this optimization is independent of any UB. The compiler is just allowed to optimize the copy away, as it doesn't have any (legally) observable effect. It would do the same optimization if `spawn_core1` didn't have another copy of the reference. The issue is not that there are aliasing mutable references. The actual issue is just that the compiler is allowed to assume that the original `entry` stays valid until the end of `core1_startup`, because it is `noalias`. But this assumption is wrong: `entry` will become invalid after `fifo_write(1);`. And that's the actual root cause of the bug.

Now, does it make a difference? I think, theoretically, yes: `let entry = unsafe { ManuallyDrop::take(&mut *entry) };`  still creates a mutable reference, even if only temporarily, and the compiler might still assume that it must not alias and that the copy could be elided. But in practice, your change works: In this case, the compiler doesn't take advantage of the non-aliasing guarantee. There's actually a ticket asking for that optimization: https://github.com/rust-lang/rust/issues/38941 . Given that the ticket is from 2017, it doesn't seem likely to change anytime soon.

rp2040-hal/src/multicore.rs
```diff
@@ -155,12 +155,15 @@ impl<'p> Core<'p> {
...
                 _: u64,
                 _: u64,
-                entry: &mut ManuallyDrop<F>,
+                entry: *mut ManuallyDrop<F>,
                 stack_bottom: *mut usize,
             ) -> ! {
                 core1_setup(stack_bottom);
 
-                let entry = unsafe { ManuallyDrop::take(entry) };
+                let entry = unsafe { ManuallyDrop::take(&mut *entry) };
```
jannic, line 163: I'd prefer to explicitly read the pointer value, instead of just dereferencing it.
`&mut *entry` still points to memory on the core0 stack, and `ManuallyDrop::take` is not guaranteed to copy the contents somewhere else.

```suggestion
                let entry = unsafe { ManuallyDrop::take(&mut core::ptr::read(entry)) };
```

Dirbaio: I think it's sound either with or without the `ptr::read()`. `ptr::read()` can still be optimized away, it's logically the same as a regular dereference. Only `ptr::read_volatile()` is guaranteed to not be optimized away.

rp2040-hal/src/multicore.rs
```diff
@@ -181,20 +184,28 @@ impl<'p> Core<'p> {
...
 
             // Set up the stack
-            let mut stack_ptr = unsafe { stack.as_mut_ptr().add(stack.len()) };
+            // AAPCS requires in 6.2.1.2 that the stack is 8bytes aligned., we may need to trim the
+            // array size to guaranty that the bottom of the stack (the end of the array) meets that requirement.
+            // The start of the array does not need to be aligned.
+
+            let mut stack_ptr = stack.as_mut_ptr_range().end;
+            // on rp2040, usize are 4 bytes, so align_offset(8) on a *mut usize returns either 0 or 1.
+            let miss_alignement_offset = stack_ptr.align_offset(8);
```
Dirbaio, line 193: misalignment

rp2040-hal/src/multicore.rs
Dirbaio: > But if the value is used, the read must happen before the fence. That's guaranteed by the ptr::read.

it is also guaranteed by the dereference. It is also a memory read (from the pointer to a local variable).

> we can't be sure that the original location doesn't alias. In fact we know that it does.

The `*mut` pointer is borrowed from the original `entry` in core0 stack, so we are allowed to access it even with `&mut`. Only when core0 touches `entry` again, the `*mut` is invalidated (see Stacked Borrows / Tree Borrows). We know that doesn't happen because core0 doesn't do so until we tell it it's fine to do. And when we do, the `&mut` already no longer exists.

ithinuel: `ManuallyDrop::take`  does the `ptr::read`. That's actually the only thing it does.

I am wondering why don't we pass in the `ManuallyDrop<F>` directly rather than an `*mut` to it?
As a far as I can see from a little godbolt experiment, the only difference is that when using `ManuallyDrop::into_inner` it does not seem able to lower into a memcpy while this is what is done when using `ManuallyDrop::take`.
IMHO this is really an issue only if F is large, and F must not be large.

jannic: > F must not be large

Why?

ithinuel: Because a large object on a stack that has no enforced limit is unsound and likely to (unexpectedly?) overflow.

ithinuel: My Extremely limited understanding is that &mut has a lifetime attached while *mut does not. And that might not even be true 😅

jannic: > I am wondering why don't we pass in the `ManuallyDrop<F>` directly rather than an `*mut` to it?

We place the parameters on the stack manually, and need to follow the AAPCS32 calling convention standards. For the pointer/reference, that's easy. But a `ManuallyDrop<F>` has the same memory layout as `F`, ie. it depends on what `F` is. That _could_ become difficult. For example, what happens if `F` contains a field with unusual alignment requirements?

jannic:
Just in case someone else is as curious, I played around a bit with compiler explorer.
The optimization in question that did cause a bad compilation result happens only with current nightly / beta, but not with stable as of rust 1.69.

I'm compiling this function:
```
pub extern "C" fn testf(p: &mut ManuallyDrop<[u8; 1024]>) {
   let q = unsafe { ManuallyDrop::take(p) };
   inner(&q);
}
```

In one of the `InstCombinePass` optimization steps, this:
```
define dso_local void @example::testf(ptr noalias nocapture noundef readonly align 1 dereferenceable(1024) %p) unnamed_addr {
start:
  %q = alloca [1024 x i8], align 1
  call void @llvm.lifetime.start.p0(i64 1024, ptr nonnull %q)
  call void @llvm.memcpy.p0.p0.i32(ptr noundef nonnull align 1 dereferenceable(1024) %q, ptr noundef nonnull align 1 dereferenceable(1024) %p, i32 1024, i1 false)
  call fastcc void @example::inner(ptr noalias noundef nonnull readonly align 1 dereferenceable(1024) %q)
  unreachable
}
```
gets optimized to this:
```
define dso_local void @example::testf(ptr noalias nocapture noundef readonly align 1 dereferenceable(1024) %p) unnamed_addr {
start:
  call fastcc void @example::inner(ptr noalias noundef nonnull readonly align 1 dereferenceable(1024) %p)
  unreachable
}
```

It's obvious that the `memcpy` gets optimized away. This doesn't change if I place fences in front of the call to `inner`. 

In comparison, this function doesn't get optimized in the same way:
```
pub extern "C" fn testg(p: *mut ManuallyDrop<[u8; 1024]>) {
   let q = unsafe { ManuallyDrop::take(&mut *p) };
   inner(&q);
}
```

The step mentioned above doesn't change the code:
```
define dso_local void @example::testg(ptr nocapture noundef readonly %p) unnamed_addr {
start:
  %q = alloca [1024 x i8], align 1
  call void @llvm.lifetime.start.p0(i64 1024, ptr nonnull %q)
  call void @llvm.memcpy.p0.p0.i32(ptr noundef nonnull align 1 dereferenceable(1024) %q, ptr noundef nonnull align 1 dereferenceable(1024) %p, i32 1024, i1 false)
  call fastcc void @example::inner(ptr noalias noundef nonnull readonly align 1 dereferenceable(1024) %q)
  unreachable
}
```
→
```
define dso_local void @example::testg(ptr nocapture noundef readonly %p) unnamed_addr {
start:
  %q = alloca [1024 x i8], align 1
  call void @llvm.lifetime.start.p0(i64 1024, ptr nonnull %q)
  call void @llvm.memcpy.p0.p0.i32(ptr noundef nonnull align 1 dereferenceable(1024) %q, ptr noundef nonnull align 1 dereferenceable(1024) %p, i32 1024, i1 false)
  call fastcc void @example::inner(ptr noalias noundef nonnull readonly align 1 dereferenceable(1024) %q)
  unreachable
}
```

What does this tell us?

- It's probably a new optimization, so it's no surprise that we didn't see such miscompilations earlier
  (Thanks to embassy for using the nightly compiler, and catching the bug before it affects stable users :smile: )
- As expected, the difference between the original and the fixed version is that the pointer passed in has different attributes. `noalias` is a likely reason.

This doesn't help answering the question if the new code is 100% correct. But that was expected: Looking at the compiler steps can't tell us what optimizations a future compiler might apply.

https://rust.godbolt.org/z/TxxdTfrdG

Issue #564, opened 2023-03-06
Provide functions to safely halt or reset the RP2040


jannic:
A discussion on https://matrix.to/#/#rust-embedded:matrix.org showed that it's difficult to safely halt or reset a multicore MCU.

It would be nice if the HAL provided simple to use methods to halt / reset the full chip. Those could also be used from panic / HardFault handlers.

For reset, this seems to work:
```
        (*pac::PSM::PTR).wdsel.write(|w| w.bits(0x0000ffff));    
        (*pac::WATCHDOG::PTR).ctrl.write(|w| w.trigger().set_bit());    
```
However, it has the disadvantage that it also breaks the debugger connection.
Alternatively, this will first turn core1 off and then reset core0:
```
        (*pac::PSM::PTR).frce_off.write(|w| w.proc1().set_bit());                                                                                                                                               
        pac::SCB::sys_reset();                                                                                                                                                                                  
```
This doesn't break the debugger connection. But it won't work when called from core1.

Similarly, for halting the MCU from core0, this works:
```
        (*pac::PSM::PTR).frce_off.write(|w| w.proc1().set_bit());                                                                                                                                               
        loop {}
```
(Don't forget to also disable interrupts, though.)
Some method that works from both cores the same way would be nicer.

thejpster:
Is there a register we can read to know which core we are? Then we can do the right thing no matter.

9names:
Yes, the sio module has a register for this:
https://docs.rs/rp2040/latest/rp2040/sio/cpuid/index.html

PR #613, opened 2023-05-16
Implement reset() and halt() functions

jannic:
Possible implementation of #564

### DIFF
--- a/rp2040-hal/src/lib.rs
+++ b/rp2040-hal/src/lib.rs
@@ -88,6 +88,7 @@ pub use i2c::I2C;
 /// of `main`. As spinlocks are not automatically unlocked on software resets,
 /// this can prevent unexpected deadlocks when running from a debugger.
 pub use rp2040_hal_macros::entry;
+use sio::CoreId;
 pub use sio::Sio;
 pub use spi::Spi;
 pub use timer::Timer;
@@ -104,7 +105,7 @@ pub use watchdog::Watchdog;
 /// which breaks a running debug connection.
 pub fn reset() -> ! {
     unsafe {
-        if crate::Sio::core() == 0 {
+        if crate::Sio::core() == CoreId::Core0 {
             (*pac::PSM::PTR).frce_off.write(|w| w.proc1().set_bit());
             pac::SCB::sys_reset();
         } else {
@@ -126,7 +127,7 @@ pub fn halt() -> ! {
     unsafe {
         cortex_m::interrupt::disable();
         // Stop other core
-        if crate::Sio::core() == 0 {
+        if crate::Sio::core() == CoreId::Core0 {
             (*pac::PSM::PTR).frce_off.write(|w| w.proc1().set_bit());
         } else {
             (*pac::PSM::PTR).frce_off.write(|w| w.proc0().set_bit());
--- a/rp2040-hal/src/lib.rs
+++ b/rp2040-hal/src/lib.rs
@@ -117,10 +117,9 @@ pub fn halt() -> ! {
     unsafe {
         cortex_m::interrupt::disable();
         // Stop other core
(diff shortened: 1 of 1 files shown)


rp2040-hal/src/lib.rs
```diff
@@ -90,3 +90,48 @@ pub use sio::Sio;
...
+/// then call `SCB::sys_reset()`, which keeps the debug
+/// connection active.
+///
+/// This is not possible when called from core1. It will
+/// trigger a watchdog reset of the whole system instead,
+/// which breaks a running debug connection.
+pub fn reset() -> ! {
+    unsafe {
+        if crate::Sio::core() == 0 {
+            (*pac::PSM::PTR).frce_off.write(|w| w.proc1().set_bit());
```
jannic, line 106: This should probably reset other parts of the MCU as well. At least the SIO to clear the spinlocks.

rp2040-hal/src/lib.rs
```diff
@@ -90,3 +90,48 @@ pub use sio::Sio;
...
+        loop {}
+    }
+}
+
+/// Halt the RP2040.
+///
+/// Completely disables the other core, and parks the current core in an
+/// infinite loop with interrupts disabled.
+///
+/// Doesn't stop other subsystems.
```
jannic, line 122: Is that a good choice, or should it disable as much as possible?
With DMA and PIO still running, the RP2040 can do a lot even with both cores halted.

jannic:
Unfortunately, I'm yet to find a good way to reset the rp2040 without losing the debug connection.

The current implementation, which just stops core1 an then calls `sys_reset` does work, but doesn't fully reset the rp2040. Notably, the spinlocks aren't reset.

However, any attempt to also reset the SIO (which hosts the spinlocks) also breaks the debug connection.

I tend to drop the requirement that the debug connection should stay alive, and instead just trigger a full watchdog reset. Any opinions?

jannic:
For now, I settled on just doing a full reset, even if it breaks the debug connection. I consider a reliable reset more important than the ability to debug through a reset sequence.

thejpster:
Neotron Pico BIOS uses scratchpad registers to detect a core0 reset and forces a SoC reset with the watchdog. probe-rs gets kicked out, but the workaround I found is to flash it from within the USB bootloader - none of my hardware is configured, and so my crowbar routine is neither triggered nor required.

rp2040-hal/src/lib.rs
```diff
@@ -91,7 +91,43 @@ pub use i2c::I2C;
...
+///
+/// Completely disables the other core, and parks the current core in an
+/// infinite loop with interrupts disabled.
+///
+/// Doesn't stop other subsystems, like the DMA controller.
+pub fn halt() -> ! {
+    unsafe {
+        cortex_m::interrupt::disable();
+        // Stop other core
+        if crate::Sio::core() == CoreId::Core0 {
```
thejpster, line 123: I know there's only two cores, but I think a match might be better here.

jannic:
> Neotron Pico BIOS uses scratchpad registers to detect a core0 reset and forces a SoC reset with the watchdog. probe-rs gets kicked out, but the workaround I found is to flash it from within the USB bootloader - none of my hardware is configured, and so my crowbar routine is neither triggered nor required.

That's a nice solution, but I don't think it's suitable for the HAL: The firmware might need the scratchpad registers for some other purpose.

Also, what this PR implements is mainly what you do in `watchdog_reboot`, just without the 10ms delay. If the firmware author makes sure to use it everywhere instead of calling `sys_reset`, no boot-time workaround is needed. (Of course, the workaround would still be useful in case the incomplete reset was done by the debug probe.)

thejpster:
rebase on 674 I guess?

thejpster (approved):

PR #614, opened 2023-05-18
Cloneable timer

jannic:
Make `Timer` a cloneable type:

We don't support de-initializing a `Timer` anyways. And once it's initialized, all operations on it are either read-only, or already wrapped in the separate alarm structures.

Therefore, we can safely allow for `Timer` to be cloned, and implement the delay functions directly on it. That makes the API more ergonomic.

Co-authored-by: Wilfried Chauveau <wilfried.chauveau@ithinuel.me>

### DIFF
--- a/rp2040-hal/src/timer.rs
+++ b/rp2040-hal/src/timer.rs
@@ -141,13 +141,9 @@ macro_rules! impl_delay_traits {
     ($($t:ty),+) => {
         $(
         impl embedded_hal::blocking::delay::DelayUs<$t> for Timer {
-            fn delay_us(&mut self, mut us: $t) {
+            fn delay_us(&mut self, us: $t) {
                 #![allow(unused_comparisons)]
                 assert!(us >= 0); // Only meaningful for i32
-                while <$t>::MAX as u64 > u32::MAX as u64 && us > u32::MAX as $t {
-                    self.delay_us_internal(u32::MAX);
-                    us -= u32::MAX as $t;
-                }
                 self.delay_us_internal(us as u32)
             }
         }
@@ -165,7 +161,7 @@ macro_rules! impl_delay_traits {
 }
 
 // The implementation for i32 is a workaround to allow `delay_ms(42)` construction without specifying a type.
-impl_delay_traits!(u8, u16, u32, u64, i32);
+impl_delay_traits!(u8, u16, u32, i32);
 
 #[cfg(feature = "eh1_0_alpha")]
 impl eh1_0_alpha::delay::DelayUs for Timer {
--- a/rp2040-hal/src/timer.rs
+++ b/rp2040-hal/src/timer.rs
@@ -113,7 +113,7 @@ impl Timer {
     }
 
     /// Pauses execution for at minimum `us` microseconds.
-    fn delay_us(&self, mut us: u32) {
+    fn delay_us_internal(&self, mut us: u32) {
         let mut start = self.get_counter_low();
         // If we knew that the loop ran at least once per timer tick,
         // this could be simplified to:
@@ -145,10 +145,10 @@ macro_rules! impl_delay_traits {
                 #![allow(unused_comparisons)]
                 assert!(us >= 0); // Only meaningful for i32
                 while <$t>::MAX as u64 > u32::MAX as u64 && us > u32::MAX as $t {
-                    (*self).delay_us(u32::MAX);
+                    self.delay_us_internal(u32::MAX);
                     us -= u32::MAX as $t;
                 }
-                (*self).delay_us(us as u32)
+                self.delay_us_internal(us as u32)
             }
         }
         impl embedded_hal::blocking::delay::DelayMs<$t> for Timer {
@@ -156,7 +156,7 @@ macro_rules! impl_delay_traits {
                 #![allow(unused_comparisons)]
                 assert!(ms >= 0); // Only meaningful for i32
                 for _ in 0..ms {
-                    self.delay_us(1000);
+                    self.delay_us_internal(1000);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


rp2040-hal/src/timer.rs
```diff
@@ -61,17 +59,32 @@ impl Timer {
...
         timer.reset_bring_down(resets);
         timer.reset_bring_up(resets);
-        Self { timer }
+        Self { _private: () }
+    }
+
+    // Safety: Must only be used for read-only operations
+    fn timer(&self) -> &crate::pac::timer::RegisterBlock {
+        unsafe { &*pac::TIMER::PTR }
     }
```
ithinuel, line 68: Could we keep this as an unsafe function to force point of use to check on these requirements?

It seems too easy to forget in a subsequent one-liner PR.

jannic: I inlined the 3 accesses instead. The difference between `unsafe { &*pac::TIMER::PTR }` and `unsafe { self.timer() }` is not big enough to justify a separate function.

rp2040-hal/src/timer.rs
```diff
@@ -35,21 +34,20 @@ fn release_alarm(mask: u8) {
...
+///
+/// This struct logically wraps a `pac::TIMER`, but doesn't actually store it:
+/// As after initialization all accesses are read-only anyways, the `pac::TIMER` can
+/// be summoned unsafely instead. This allows timer to be cloned.
+///
+/// (Alarms do use write operations, but they are local to the respective alarm, and
+/// those are still owned singletons.)
+///
+/// As the timer peripheral needs to be started first, this struct can only be
+/// constructed by calling `Timer::new(...)`.
```
ithinuel, line 47: While I see the value of this comment, I am not sure we should let implementation details transpire in the html doc.

jannic: Oh yes, sorry - I didn't even notice that this is a doc comment.

rp2040-hal/src/timer.rs
```diff
@@ -84,27 +94,85 @@ impl Timer {
...
 
-// safety: all write operations are synchronised and all reads are atomic
-unsafe impl Sync for Timer {}
+macro_rules! impl_delay_traits {
+    ($($t:ty),+) => {
+        $(
+        impl embedded_hal::blocking::delay::DelayUs<$t> for Timer {
+            fn delay_us(&mut self, mut us: $t) {
+                #![allow(unused_comparisons)]
+                assert!(us >= 0); // Only meaningful for i32
```
ithinuel, line 146: Is this completely elliminated when the comparison is unused ?

rp2040-hal/src/timer.rs
```diff
@@ -84,27 +94,85 @@ impl Timer {
...
     }
 
     /// Retrieve a reference to alarm 3. Will only return a value the first time this is called
     pub fn alarm_3(&mut self) -> Option<Alarm3> {
-        take_alarm(1 << 3).then_some(Alarm3(PhantomData))
+        take_alarm(1 << 3).then_some(Alarm3(*self))
+    }
+
+    /// Pauses execution for at minimum `us` microseconds.
+    fn delay_us(&self, mut us: u32) {
```
ithinuel, line 116: I think this should be named `delay_us_internal` or something similar to avoid the odd reborrows `(*self).delay_us(…)`.

rp2040-hal/src/timer.rs
jannic: I think so, yes. I checked the generated assembly (at optimizer level "z") for u16, u32 and u64, as well as i32 with a non-negative constant value. In all those cases the comparison with 0 and the branch with the panic call get removed by the compiler.

rp2040-hal/src/timer.rs
```diff
@@ -84,27 +94,85 @@ impl Timer {
...
+    ($($t:ty),+) => {
+        $(
+        impl embedded_hal::blocking::delay::DelayUs<$t> for Timer {
+            fn delay_us(&mut self, mut us: $t) {
+                #![allow(unused_comparisons)]
+                assert!(us >= 0); // Only meaningful for i32
+                while <$t>::MAX as u64 > u32::MAX as u64 && us > u32::MAX as $t {
+                    self.delay_us_internal(u32::MAX);
+                    us -= u32::MAX as $t;
+                }
```
ithinuel, line 150: I really find this convoluted for a loop that's only present when we're implementing support for u64.
I think it would be more readable and less error-prone to have this in its own impl block.

ithinuel (changes requested):

rp2040-hal/src/timer.rs
jannic: Fair. It's one thing that the optimizer removes the loop, but that doesn't help readability. I'll implement the u64 version separately.

jannic: Or should we skip the u64 implementation entirely? A blocking delay of more than an hour is usually just wrong. And if someone really needs it for some reason, it's not difficult to implement it locally:
```
 use embedded_hal::blocking::delay::DelayUs;                                                                                                                                                                     
                                                                                                                                                                                                                 
 pub struct DelayU64<T>(T);                                                                                                                                                                                      
                                                                                                                                                                                                                 
 impl<T> DelayU64<T> {                                                                                                                                                                                           
     pub fn new(delay: T) -> Self {                                                                                                                                                                              
         Self(delay)                                                                                                                                                                                             
     }                                                                                                                                                                                                           
 }                                                                                                                                                 
                                                                                                                                                                                                                 
 impl<T> DelayUs<u64> for DelayU64<T>                                                                                                                                                                            
 where                                                                                                                                                                                                           
     T: DelayUs<u32>,                                                                                                                                                                                            
 {                                                                                                                                                                                                               
     fn delay_us(&mut self, mut us: u64) {                                                                                                                                                                 
         while us > u32::MAX as u64 {                                                                                                                                                                
             self.0.delay_us(u32::MAX);                                                                                                                                                                          
             us -= u32::MAX as u64;                                                                                             
         }                                                                                                                      
         self.0.delay_us(us as u32);                                                                                            
     }                                                                                                                          
 } 
 ```
 
 And in case anyone really misses these functions, we can always add them later.

ithinuel (approved):

PR #618, opened 2023-05-20
Make sure clocks are initialized before creating a Timer

### LINKED ISSUE #568: Should Timer::new() make sure clocks are initialized?
Currently, `Timer::new(...)` can be called without initializing clocks and watchdog. This will most probably cause the timer to return nonsensical values.

One way to avoid this would be to change the signature of the function:
```
   pub fn new(_clocks: ClocksManager, timer: TIMER, resets: &mut RESETS) -> Self
```
While this still doesn't guarantee that the clocks are configured to correct values, it at least proves that initializing the clocks wasn't simply forgotten.

jannic:
Closes #568


ithinuel (approved):

jannic:
I missed some examples that need to be adapted. I'll try to update them tomorrow.

PR #619, opened 2023-05-21
bump hal's version

ithinuel:
We already introduced some breaking changes so we might as well already increase the version number on git a while before the release.

That may help users using the git repo be warned about more upcoming breakage.


jannic (approved):

jannic:
I pushed a branch of rp-hal-boards using this version to https://github.com/rp-rs/rp-hal-boards/tree/hal-0.9

PR #626, opened 2023-05-31
ADC free-running mode & FIFO

nilclass:
An attempt to come up with an API for the ADC in free-running mode.

Inspired by #326.

Overview of changes:
- Adds two new structs to the `adc` module: `FreeRunning` and `Fifo` (EDIT: naming changed to `AdcFifoBuilder` and `AdcFifo`)
- `FreeRunning` represents the `Adc` configuration for free-running mode:
  - it is obtained by calling `adc.free_running()`
  - it provides methods to configure things that only make sense in this mode: set clock_divider (DIV.INT, DIV.FRAC), set round_robin (CS.RROBIN), ...
  - and methods to start enable the FIFO and start the capture (CS.START_MANY): `start_fifo`, `start_fifo_with_dma(thresh)`
- the `Fifo` struct represents the ADC Fifo while capture is running:
  - it is returned by the `start_*` methods from `FreeRunning`
  - `fifo.len()` and `fifo.read()` give access to the FCS.LEVEL and FIFO register values
  - `fifo.is_over()` and `fifo.is_under()` expose the FCS.OVER and FCS.UNDER flags (and reset them on read)
  - `fifo.stop()` stops the capture and disables the fifo

Brief example:
```
    let mut adc = Adc::new(dp.ADC, &mut resets);
    let mut adc_pin_0 = pins.gpio26.into_floating_input();
    let mut adc_pin_1 = pins.gpio27.into_floating_input();

    let mut fifo = adc.free_running()
        .clock_divider(0, 0)
        // first conversion is from pin 26
        .initial_input(&mut adc_pin_0)
        // cycle between pin 26 and 27
        .round_robin((&mut adc_pin_0, &mut adc_pin_1))
        // enable fifo and starts capture
        .start_fifo();

    // alternative to .start_fifo(), which also enables FCS.DREQ (with given FCS.THRESH of 1):
    //  .start_fifo_with_dma(1)

    let mut i = 0;

    while i < 100 {
        if fifo.len() >= 2 {
            // a comes from pin 26, b from pin 27
            let (a, b) = (fifo.read(), fifo.read());

            // ... do something with `a` and `b`

            i += 1;
        }
    }

    // Stop capturing (CS.START_MANY = 0).
    // Returns the Adc, so it can be reused for single capture or another free-running capture.
    let adc = fifo.stop();
```

### DIFF
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -44,6 +44,7 @@ frunk = { version = "0.4.1", default-features = false }
 
 [dev-dependencies]
 cortex-m-rt = "0.7"
+cortex-m-rtic = "1.1.4"
 panic-halt = "0.2.0"
 rp2040-boot2 = "0.3.0"
 hd44780-driver = "0.4.0"
@@ -90,4 +91,9 @@ required-features = ["rt", "critical-section-impl"]
 [[example]]
 # vector_table example uses cortex-m-rt::interrupt, need rt feature for that
 name = "vector_table"
+required-features = ["rt", "critical-section-impl"]
+
+[[example]]
+# adc_fifo_irq example uses cortex-m-rt::interrupt, need rt feature for that
+name = "adc_fifo_irq"
 required-features = ["rt", "critical-section-impl"]
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -39,6 +39,7 @@ chrono = { version = "0.4", default-features = false, optional = true }
 defmt = { version = ">=0.2.0, <0.4", optional = true }
 
 rtic-monotonic = { version = "1.0.0", optional = true }
+num-traits = { version = "0.2.15", default-features = false }
 
 frunk = { version = "0.4.1", default-features = false }
 
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -39,7 +39,6 @@ chrono = { version = "0.4", default-features = false, optional = true }
 defmt = { version = ">=0.2.0, <0.4", optional = true }
 
 rtic-monotonic = { version = "1.0.0", optional = true }
-num-traits = { version = "0.2.15", default-features = false }
 
 frunk = { version = "0.4.1", default-features = false }
 
--- a/rp2040-hal/src/adc.rs
+++ b/rp2040-hal/src/adc.rs
@@ -243,7 +243,7 @@ impl Adc {
     ///
     /// Capturing is started by calling [`AdcFifoBuilder::start`], which
     /// returns an [`AdcFifo`] to read from.
-    pub fn build_fifo<'a>(&'a mut self) -> AdcFifoBuilder<'a> {
+    pub fn build_fifo<'a>(&'a mut self) -> AdcFifoBuilder<'a, false> {
         AdcFifoBuilder { adc: self }
     }
 
@@ -306,11 +306,11 @@ where
 /// Used to configure & build an [`AdcFifo`]
 ///
 /// See [`Adc::build_fifo`] for details, as well as the `adc_fifo_*` [examples](https://github.com/rp-rs/rp-hal/tree/main/rp2040-hal/examples).
-pub struct AdcFifoBuilder<'a> {
+pub struct AdcFifoBuilder<'a, const SHIFTED: bool> {
     adc: &'a mut Adc,
 }
 
-impl<'a> AdcFifoBuilder<'a> {
+impl<'a, const SHIFTED: bool> AdcFifoBuilder<'a, SHIFTED> {
     /// Manually set clock divider integral and fractional parts
     ///
     /// Default: 0, 0
@@ -352,12 +352,23 @@ impl<'a> AdcFifoBuilder<'a> {
         self
     }
 
+    /// Shift values to produce 8 bit samples (discarding the lower 4 bits).
+    ///
+    /// Normally the ADC uses 12 bits of precision, packed into a u16.
+    /// Shifting the values loses some precision, but produces smaller samples.
+    ///
+    /// When this method has been called, the resulting fifo's `read` method returns u8.
+    pub fn shift_8bit(self) -> AdcFifoBuilder<'a, true> {
+        self.adc.device.fcs.modify(|_, w| w.shift().set_bit());
+        AdcFifoBuilder { adc: self.adc }
+    }
+
     /// Enable ADC FIFO and start free-running conversion
     ///
     /// Use the returned [`AdcFifo`] instance to access the captured data.
     ///
     /// To stop capturing, call [`AdcFifo::stop`].
-    pub fn start(self) -> AdcFifo<'a> {
+    pub fn start(self) -> AdcFifo<'a, SHIFTED> {
         self.adc.device.fcs.modify(|_, w| w.en().set_bit());
         self.adc.device.cs.modify(|_, w| w.start_many().set_bit());
         AdcFifo { adc: self.adc }
@@ -368,11 +379,11 @@ impl<'a> AdcFifoBuilder<'a> {
 ///
 /// Constructed by [`AdcFifoBuilder::start`], which is accessible through [`Adc::build_fifo`].
 ///
-pub struct AdcFifo<'a> {
+pub struct AdcFifo<'a, const SHIFTED: bool> {
     adc: &'a mut Adc,
 }
 
-impl<'a> AdcFifo<'a> {
+impl<'a, const SHIFTED: bool> AdcFifo<'a, SHIFTED> {
     /// Returns the number of elements currently in the fifo
     pub fn len(&mut self) -> u8 {
         self.adc.device.fcs.read().level().bits()
@@ -404,11 +415,6 @@ impl<'a> AdcFifo<'a> {
         under
     }
 
-    /// Read a single value from the fifo
-    pub fn read(&mut self) -> u16 {
-        self.adc.device.fifo.read().val().bits()
-    }
-
     /// Stop capturing in free running mode.
     ///
     /// Resets all capture options that can be set via [`AdcFifoBuilder`] to
@@ -431,7 +437,7 @@ impl<'a> AdcFifo<'a> {
         // The only way to clear `INTS.FIFO` is for `FCS.LEVEL` to go
         // below `FCS.THRESH`, which requires `FCS.THRESH` not to be 0.
         while self.len() > 0 {
-            self.read();
+            self.read_from_fifo();
         }
         // disable fifo and reset threshold to 0
         self.adc.device.fcs.modify(|_, w| unsafe {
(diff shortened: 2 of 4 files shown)


rp2040-hal/Cargo.toml
```diff
@@ -39,6 +39,7 @@ chrono = { version = "0.4", default-features = false, optional = true }
 defmt = { version = ">=0.2.0, <0.4", optional = true }
 
 rtic-monotonic = { version = "1.0.0", optional = true }
+num-traits = { version = "0.2.15", default-features = false }
```
jannic, line 42: I think this isn't used at all. At least clippy marks the only obvious reference to this crate, the import of `FloatCore`, as unused.

rp2040-hal/src/adc.rs
```diff
@@ -153,3 +154,210 @@ where
...
+    /// Stop capturing in free running mode.
+    ///
+    /// Disables the ADC fifo and stops ADC capture.
+    /// Returns the underlying `Adc` instance.
+    pub fn stop(self) -> &'a mut Adc {
+        self.adc.device.cs.modify(|_, w| w.start_many().clear_bit());
+        self.adc.device.fcs.modify(|_, w| w.en().clear_bit());
+        self.adc
+    }
+}
```
jannic, line 291: Just an observation, not sure if it's good or bad: This doesn't disable DREQ. I _think_ it doesn't do harm to keep it enabled, and I also _think_ it would be fine to disable it here. Anyone with deeper knowledge of the DMA engine around who could tell if one or the other would be better?

rp2040-hal/src/adc.rs
```diff
@@ -153,3 +154,210 @@ where
...
+        self.adc.device.fcs.modify(|_, w| w.en().set_bit());
+        self.adc.device.cs.modify(|_, w| w.start_many().set_bit());
+        Fifo { adc: self.adc }
+    }
+
+    /// Enable ADC fifo with DMA and start free-running conversion
+    ///
+    /// In addition to being usable the same way as a `Fifo` returned from `start_fifo`,
+    /// the Fifo returned by this function can also be used as a source for DMA transfers.
+    pub fn start_fifo_with_dma(self, thresh: u8) -> Fifo<'a, true> {
```
jannic, line 210: The datasheet says: "The threshold for DREQ assertion (FCS.THRESH) should be set to 1, so that the DMA transfers as soon as a single sample is present in the FIFO. Note this is also the threshold used for IRQ assertion, so non-DMA use cases might prefer a higher value for less frequent interrupts."

Therefore I'd say for the DMA use case, the threshold should not be configurable, but fixed to 1.
Also, it probably (but see my other comment) doesn't harm to always enable DREQ. So `start_fifo_with_dma` and `start_fifo` may not need to be separate functions? And the DmaEnabled flag could be removed from the Fifo type signature?

rp2040-hal/src/adc.rs
```diff
@@ -153,3 +154,210 @@ where
...
+    D: Channel<Adc, ID = u8>,
+    E: Channel<Adc, ID = u8>,
+{
+    fn from(_: (&mut A, &mut B, &mut C, &mut D, &mut E)) -> Self {
+        Self(1 << A::channel() | 1 << B::channel() | 1 << C::channel() | 1 << D::channel() | 1 << E::channel())
+    }
+}
+
+impl crate::dma::ReadTarget for Fifo<'_, true> {
+    type ReceivedWord = u8;
```
jannic, line 348: From the datasheet: "If the DMA transfer size is set to 8 bits, so that the DMA transfers to a byte array in memory, FCS.SHIFT must also be set, to pre-shift the FIFO samples to 8 bits of significance."

rp2040-hal/src/adc.rs
```diff
@@ -153,3 +154,210 @@ where
...
+    ///
+    /// In addition to being usable the same way as a `Fifo` returned from `start_fifo`,
+    /// the Fifo returned by this function can also be used as a source for DMA transfers.
+    pub fn start_fifo_with_dma(self, thresh: u8) -> Fifo<'a, true> {
+        self.adc.device.fcs.modify(|_, w| unsafe {
+            w.en().set_bit()
+                .dreq_en().set_bit()
+                .thresh().bits(thresh)
+        });
+        self.adc.device.cs.modify(|_, w| w.start_many().set_bit());
```
jannic, line 216: This doesn't adhere to the documented sequence for starting DMA on the ADC:

> Once the ADC is suitably configured, the DMA channel should be started first, and the ADC conversion should be started second, via CS.START_MANY. Once the DMA completes, the ADC can be halted, or a new DMA transfer promptly started. After clearing CS.START_MANY to halt the ADC, software should also poll CS.READY to make sure the last conversion has finished, and then drain any stray samples from the FIFO.

This is something we probably can't do properly with the current DMA API.
Not sure what to do about it. If you tested your code and it works reliably, perhaps we can just add a doc comment that this isn't properly following the spec for now?

jannic (changes requested):
Hi @nilclass, thanks for this pull request. Overall I like the API, but there are a few details left to do.

nilclass:
@jannic thanks for taking the time to review this.

I realize this PR is not in a good state to merge, it's more of a draft, to see if there is interest in such an API :slightly_smiling_face: 

I'll respond to your feedback inline.

rp2040-hal/src/adc.rs
nilclass: hm, yeah, those are some good points.

If DREQ is always enabled, we can't enforce FCS.THRESH=1 for the DMA usecase (or else not have the threshold configurable).

We could move this issue to the user (at least until it's clear how to handle it nicely API-wise), by having builder methods `set_threshold(...)`, `enable_dma(true | false)`, `enable_interrupt(true | false)`, and have a single `start_fifo` function as you suggest.
That would make these features usable without messing with registers manually, but it would leave some caveats:
- either there needs to be some note, that when enabling DMA, the threshold should be set to 1
- or `enable_dma` would force the threshold to 1, possibly overriding the user choice (or not, if the user calls `set_threshold` after `enable_dma`)

Not the best options, imho.

So, also in light of your other comments: maybe the DMA part can be skipped for now, and handled in a future PR, when more is known about the usecases.
Then my goal would be to make the API nice for the interrupt usecase.

Which makes me wonder: do you think there is a usecase where neither the interrupt, nor DMA is used?
From what I can tell when capturing in free-running mode, CS.READY does not change, so even though the samples can be read from CS.RESULT there is no way to tell when a new sample was captured.

rp2040-hal/src/adc.rs
nilclass: True. I don't know what's the best practice here.

We could store the register values when the `Fifo` is constructed, and reset them to those exact values when `stop`ping.

rp2040-hal/src/adc.rs
nilclass: tbh, I know very little about the DMA part, and I have not tested it thoroughly yet. As mentioned above, maybe it's better to drop it for now.

rp2040-hal/src/adc.rs
nilclass: Good point, this would not work properly then (or capture the wrong values). I don't know if it's possible to use DMA transfers with 12 bit precision (seems a shame to throw away those bits). In that case we would need two different `ReadTarget` implementations (one with `ReceivedWord = u8`, one with `u16`), depending on if the user decided to FCS.SHIFT or not.

rp2040-hal/src/adc.rs
nilclass: Update: turns out we also need to drain the FIFO here. Otherwise if there is still a pending interrupt, the ISR will still be called, even if interrupts are already disabled at that point.

jannic:
I'll try to do the review soon but I'm quite busy atm so I can't promise much. May take a few days, sorry!

nilclass:
@jannic did you get a chance to take a look? 🫣

Is there anything I can do to make things easier to review?

In case there is no capacity currently for this review, or there are general concerns about the API, I could look into putting this functionality in a separate crate for now. That way it can be used, without locking `rp-hal` into this particular approach. Wdyt?

rp2040-hal/src/adc.rs
```diff
@@ -259,3 +302,234 @@ where
...
+/// See [`Adc::build_fifo`] for details, as well as the `adc_fifo_*` [examples](https://github.com/rp-rs/rp-hal/tree/main/rp2040-hal/examples).
+pub struct AdcFifoBuilder<'a, const SHIFTED: bool> {
+    adc: &'a mut Adc,
+}
+
+impl<'a, const SHIFTED: bool> AdcFifoBuilder<'a, SHIFTED> {
+    /// Manually set clock divider integral and fractional parts
+    ///
+    /// Default: 0, 0
+    pub fn clock_divider(self, int: u16, frac: u8) -> Self {
```
jannic, line 317: You could add an explanation to those parameters.

Also, in the examples (eg. https://github.com/rp-rs/rp-hal/pull/626/files#diff-54b5bbce718eaba891383cd18b03aa2b4533ef7b314c7584ec31bf880712a18bR114) it is difficult to understand why one needs such an odd divisor value.

jannic:
> @jannic did you get a chance to take a look? 🫣
> 
> Is there anything I can do to make things easier to review?
> 
> In case there is no capacity currently for this review, or there are general concerns about the API, I could look into putting this functionality in a separate crate for now. That way it can be used, without locking `rp-hal` into this particular approach. Wdyt?

Sorry, the delay was entirely caused by my lack of time, and not by anything wrong with your work. The code is nicely readable and any difficulty reviewing it is caused by the inherent complexity of the hardware.

I added a few minor remarks, but in general the pull request looks good. I won't worry too much about locking `rp-hal` into one specific API. After all, the top-level readme still states "These packages are under active development. As such, it is likely to remain volatile until a 1.0.0 release."

rp2040-hal/src/adc.rs
nilclass: I agree, that sounds reasonable.

Since `AdcFifo` currently holds a `&mut Adc` calling `adc.read_single()` directly wouldn't work though (unless the `AdcFifo` is dropped first, which makes it impossible to stop the conversion later on).
So I've added a `read_single` method to `AdcFifo` now, which just calls the underlying `adc.read_single` to enable this use case.

rp2040-hal/src/adc.rs
```diff
@@ -259,3 +302,234 @@ where
...
+            w.start_many().clear_bit()
+                .rrobin().bits(0)
+                .ainsel().bits(0)
+        });
+        // disable fifo interrupt
+        self.adc.device.inte.modify(|_, w| w.fifo().clear_bit());
+        // drain remaining values from fifo.
+        // This MUST happen *after* the interrupt is disabled, but
+        // *before* `thresh` is modified. Otherwise if `INTS.FIFO = 1`,
+        // the interrupt fill be fired one more time.
```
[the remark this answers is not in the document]
nilclass: fixed

rp2040-hal/src/adc.rs
```diff
@@ -259,3 +302,234 @@ where
...
+        // The only way to clear `INTS.FIFO` is for `FCS.LEVEL` to go
+        // below `FCS.THRESH`, which requires `FCS.THRESH` not to be 0.
+        while self.len() > 0 {
```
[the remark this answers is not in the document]
nilclass: good catch, fixed

rp2040-hal/src/adc.rs
nilclass: Fair enough :)

Added a bit of explanation, and a few example values

jannic: "A bit"? That doc comment is great!

jannic (approved):
Both code and comments are of high quality. Very readable, and correct as far as I can tell. While I didn't have time to try it out on real hardware, I think it's ready to be merged.

Thanks, @nilclass, for this great contribution, and sorry again for the delay reviewing it.

nilclass:
Thanks for the review @jannic, I'm happy to hear the changes are appreciated.

I believe the pipeline will be green now. I've just run the two examples one more time on real hardware, still working fine :)

PR #636, opened 2023-06-24
ADC fifo DMA

nilclass:
DMA support for free-running ADC capture.

The basic API looks like this:
```
// without DMA:
let adc_fifo = adc.build_fifo()
    .something_something()
    .start();

do_something_with_fifo(adc_fifo);

// with DMA:
let adc_fifo = adc.build_fifo()
    .something_something()
    .enable_dma()
    .prepare();

let transfer = dma::single_buffer::Config::new(dma.ch0, adc_fifo.dma_read_target(), some_write_target).start();
adc_fifo.resume();

do_something_else_while_dma_does_all_the_work();
```

To enable this functionality, the following methods were added:
- `AdcFifoBuilder::enable_dma`: sets `FCS.DREQ_EN = 1` and `FCS.THRESH = 1`
- `AdcFifoBuilder::prepare`: same as `start`, but does not set `CS.START_MANY` just yet
- `AdcFifo::resume` / `AdcFifo::pause`: control if capturing is active (`CS.START_MANY`) without reconfiguring the FIFO in any other way
- `AdcFifo::is_paused`: can be used to check the current state
- `AdcFifo::clear`: reads the FIFO until empty, discarding values
- `AdcFifo::dma_read_target`: returns a struct that implements `dma::ReadTarget`, and can be used to construct the transfer. (If `dma::ReadTarget` were implemented on `AdcFifo` directly, it wouldn't be possible to `resume()` the FIFO after starting the DMA transfer, because the transfer would take ownership of the `AdcFifo`)

### DIFF
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -97,3 +97,8 @@ required-features = ["rt", "critical-section-impl"]
 # adc_fifo_irq example uses cortex-m-rt::interrupt, need rt feature for that
 name = "adc_fifo_irq"
 required-features = ["rt", "critical-section-impl"]
+
+[[example]]
+# adc_fifo_dma example uses cortex-m-rt::interrupt, need rt feature for that
+name = "adc_fifo_dma"
+required-features = ["rt", "critical-section-impl"]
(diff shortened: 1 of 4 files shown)


jannic (approved):
I finally found the time to actually try it out on hardware.
LGTM, and thanks for contributing!

PR #640, opened 2023-06-28
Adds the ability to initialise the ring oscillator with a known frequ…

hardiesoft:
…ency.  This PR is necessary because freq_hz is a private member of the `RingOscillator<Enabled>` struct.

The ROSC must be brought to the desired frequency before using this method.

The rp2040 datasheet has instructions on how to measure and modify the ROSC frequency:

- 2.15.6.2. Using the frequency counter
- 2.17.3.   Modifying the frequency

### DIFF
--- a/rp2040-hal/src/rosc.rs
+++ b/rp2040-hal/src/rosc.rs
@@ -65,6 +65,18 @@ impl RingOscillator<Disabled> {
             freq_hz: 6_500_000u32.Hz(),
         })
     }
+
+    /// Initializes the ROSC with a known frequency.
+    /// See sections 2.17.3. "Modifying the frequency", and 2.15.6.2. "Using the frequency counter"
+    /// in the rp2040 datasheet for guidance on how to do this before initialising the ROSC.
+    pub fn initialize_with_freq(self, known_freq_hz: u32) -> RingOscillator<Enabled> {
+        self.device.ctrl.write(|w| w.enable().enable());
+
+        use fugit::RateExtU32;
+        self.transition(Enabled {
+            freq_hz: known_freq_hz.Hz(),
+        })
+    }
 }
 
 impl RingOscillator<Enabled> {
--- a/rp2040-hal/src/rosc.rs
+++ b/rp2040-hal/src/rosc.rs
@@ -70,10 +70,10 @@ impl RingOscillator<Disabled> {
     /// See sections 2.17.3. "Modifying the frequency", and 2.15.6.2. "Using the frequency counter"
     /// in the rp2040 datasheet for guidance on how to do this before initialising the ROSC.
     /// Also see `rosc_as_system_clock` example for usage.
-    pub fn initialize_with_freq(self, known_freq_hz: HertzU32) -> RingOscillator<Enabled> {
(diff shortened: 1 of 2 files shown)


jannic:
I don't fully understand how this is meant to be used. Don't you need to first enable the ROSC to be able to measure its frequency? So, how is the caller supposed to know the frequency before calling `initialize_with_freq`?

Also, there is some potential for misunderstanding: One could think that `initialize_with_freq` actually changes the ROSC frequency to the desired value. But in fact it doesn't change the frequency at all, but only makes the frequency which is somehow known to the caller available to the HAL.

hardiesoft:
ROSC is enabled on startup, before being initialized or known to the HAL.
At that point, with the XOSC also enabled and XOSC set as the system clock, it's possible to measure the actual ROSC frequency using the fc0 frequency counter.
At that point you can initialise the ROSC using that known frequency, and optionally disable the XOSC and use the ROSC as the system clock for power savings.  Optionally, prior to initialising the ROSC, you can iterate adding strength to the ROSC stages and measuring via the frequency counter until a desired speed is reached.

I wanted to make this as minimal a PR as possible, since it's not clear to me that all of those operations should be the responsibility of helper functions in the HAL (at least the appropriate registers are exposed to achieve setting and measuring ROSC frequency already).  

Would it help to call the method `initialize_with_known_freq` or similar?

jannic:
Yes, a better name and some more documentation would be great. Perhaps even an example?

After reading your comments and the datasheet, I understand why having this function is useful. But with the current docs, a reader without that prior knowledge would have a hard time understanding what the function actually does.

rp2040-hal/src/rosc.rs
```diff
@@ -65,6 +65,18 @@ impl RingOscillator<Disabled> {
             freq_hz: 6_500_000u32.Hz(),
         })
     }
+
+    /// Initializes the ROSC with a known frequency.
+    /// See sections 2.17.3. "Modifying the frequency", and 2.15.6.2. "Using the frequency counter"
+    /// in the rp2040 datasheet for guidance on how to do this before initialising the ROSC.
+    pub fn initialize_with_freq(self, known_freq_hz: u32) -> RingOscillator<Enabled> {
```
nilclass, line 72: The parameter could be a `HertzU32` instead of a `u32`:
```suggestion
    pub fn initialize_with_freq(self, known_freq: HertzU32) -> RingOscillator<Enabled> {
```

That way the caller does not have to check the docs or the parameter name to find out what the number means, but can instead pass something like `1_234_567.Hz()` or `1_234.kHz()`.

hardiesoft: Thank you both for the feedback!  I've now made changes to the function signature as requested, and added an example `rosc_as_system_clock`, which shows working code setting the ROSC to a desired target frequency.

jannic:
Thanks for adding the example. It really helps to understand how the added function can be used in practice.
Can you fix the clippy / formatting warnings?

hardiesoft:
Done, thanks.

jannic (approved):

PR #643, opened 2023-07-05
Add specific build-dependency on is-terminal = "=0.4.7"

jannic:
With is-terminal 0.4.8, the MSRV was changed from 1.48 to 1.63. This implicitly affects the MSRV of lalrpop, breaking existing builds.

As is-terminal has a policy of changing MSRV in a minor release, and actively suggests to pin a specific version to avoid that, I think lalrpop should do that and use exactly version 0.4.7. (So I opened https://github.com/lalrpop/lalrpop/pull/805, but it's not clear if that's a good solution either.)

To work around this issue, add a specific build-dependency to rp2040-hal. This should be removed with the next major release of rp2040-hal.

### DIFF
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -47,6 +47,9 @@ hd44780-driver = "0.4.0"
 pio-proc = "0.2.0"
 dht-sensor = "0.2.1"
 
+[build-dependencies]
+is-terminal = "=0.4.7"
+
 [features]
 # Minimal startup / runtime for Cortex-M microcontrollers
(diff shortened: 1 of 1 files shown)


9names (approved):
LGTM

hlavaatch:
Didn't they add terminal detection into the std lately (1.70.0)? `std::io::TsTerminal` trait on stuff... You may be able to drop the dependency alltogether

jannic:
Yes, but this PR is all about keeping the MSRV at 1.62.
With 1.70 we (or, for that matter, lalrpop) may indeed be able to drop
the dependency.

PR #645, opened 2023-07-08
Use fixed version nightly-2023-06-27 for cargo udeps

jannic:
This should avoid random breakage with newer nightly versions

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -52,9 +52,9 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
-      - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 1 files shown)


jannic:
The previously failing build of cargo-udeps was not actually executed in https://github.com/rp-rs/rp-hal/actions/runs/5493763368/jobs/10012154038?pr=645 because it used a cached version.
See https://github.com/jannic/rp-hal/actions/runs/5493762713/jobs/10012153213 for a successful run, and https://github.com/jannic/rp-hal/actions/runs/5493635352/jobs/10011948303 for a previously failing one.

jannic:
Wait - I'm not sure this is enough. Once a new version of cargo-udeps is released, it will probably rely on newer nightly features. So we need to specify a certain version of cargo-udeps, as well.

ithinuel (approved):

PR #621, opened 2023-05-23
Mark ReadTarget and WriteTarget as unsafe

jannic:
This implements the "mark traits as unsafe" part of https://github.com/rp-rs/rp-hal/issues/620#issuecomment-1558767880

PR #646, opened 2023-07-10
Implement WriteTarget for PWM top and cc registers.

mBornand:
Hello,
Here is an implementation of WriteTarget for PWM slices.

In my opinion this respects the safety requirements described in #621.

The implementation uses `u16` as type for the writes to the top register, which will cause the reserved upper bits to be written to by the same value. Is this problematic? I thought of using a type that has the upper 16 bits set to zero, but this would contradict the safety requirements of the `Word` trait.

### DIFF
--- a/rp2040-hal/src/pwm/mod.rs
+++ b/rp2040-hal/src/pwm/mod.rs
@@ -79,12 +79,14 @@
 use core::marker::PhantomData;
 
 use crate::{
+    dma::{EndlessWriteTarget, WriteTarget},
     gpio::{bank0::*, AnyPin, FunctionPwm, Pin, ValidFunction},
     resets::SubsystemReset,
     typelevel::{Is, Sealed},
 };
+use embedded_dma::Word;
 use embedded_hal::PwmPin;
-use pac::PWM;
+use pac::{dma::ch::ch_al1_ctrl::TREQ_SEL_A, PWM};
 
 use crate::atomic_register_access::{write_bitmask_clear, write_bitmask_set};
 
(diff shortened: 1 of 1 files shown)


rp2040-hal/src/pwm/mod.rs
```diff
@@ -764,3 +769,135 @@ impl<S: SliceId, M: ValidSliceInputMode<S>> Slice<S, M> {
...
+        Self {
+            slice: value,
+            top: SliceDmaWriteTop {
+                slice: PhantomData::<S> {},
+                mode: PhantomData::<M> {},
+            },
+            cc: SliceDmaWriteCc {
+                slice: PhantomData::<S> {},
+                mode: PhantomData::<M> {},
+            },
```
jannic, line 881: I think this can be simplified:
```suggestion
            top: SliceDmaWriteTop {
                slice: PhantomData,
                mode: PhantomData,
            },
            cc: SliceDmaWriteCc {
                slice: PhantomData,
                mode: PhantomData,
            },
```

jannic:
> I thought of using a type that has the upper 16 bits set to zero, but this would contradict the safety requirements of the `Word` trait.

That's an interesting dilemma. The root cause seems to be that the `Word` trait is used both for reading and for writing. The safety requirement for the word type is only relevant when DMA writes to a variable implementing that trait.
When reading from a variable, restricting the possible values wouldn't hurt. So if embedded-dma used different word types for reading and writing, you could make one restricted and the other one unrestricted.

One possible approach is using a restricted type which is represented internally as an u32 (so, from rust's point of view, all bit patterns are valid), but which can only be constructed from a u16. So it wouldn't be possible to construct a value with the high bits set from pure rust code.

It would still be technically possible to first write arbitrary `u32` values into such a variable by using a DMA transfer from a different memory location, and then use a second DMA transfer to write those invalid values into the TOP register. But that would be unlikely to happen by accident, and it could be forbidden by an appropriate note in the doc comment.

(I have no idea if writing the reserved bits of the TOP register has any adverse side effect. A short web search didn't result in any definite answer.)

jannic (approved):
Yes, looks good, thank you!

jannic:
While I think the current approach is fine, I was curious and asked the question in the raspberry forum at https://forums.raspberrypi.com/viewtopic.php?t=353551.

PR #647, opened 2023-07-11
Avoid missing docs warning on pub extern crate

jannic:
Due to https://github.com/rust-lang/rust/issues/112308, the public re-export of the pac will trigger a missing docs warning in rust 1.71.0.

This is bogus, as the doc string would not be shown in the rendered docs anyway. However, the warning would be annoying, so add a short doc string to the item.

### DIFF
--- a/rp2040-hal/src/lib.rs
+++ b/rp2040-hal/src/lib.rs
@@ -43,6 +43,7 @@ extern crate embedded_hal as hal;
 extern crate nb;
 pub use paste;
 
+/// Re-export of the PAC
 pub extern crate rp2040_pac as pac;
 
 #[macro_use]


9names (approved):

jannic:
> Does the issue arise if we use edition 2021's `pub use rp2040_pac as pac;` ?

It doesn't seem so, but it leads to lots of other errors:
```
    Checking rp2040-hal v0.9.0-alpha.1 (/home/jan/rp2040/rp-rs/rp-hal/rp2040-hal)
error[E0433]: failed to resolve: use of undeclared crate or module `pac`
 --> rp2040-hal/src/i2c/controller.rs:6:5
  |
6 | use pac::{i2c0::RegisterBlock as Block, RESETS};
  |     ^^^ use of undeclared crate or module `pac`

error[E0433]: failed to resolve: use of undeclared crate or module `pac`
 --> rp2040-hal/src/pwm/reg.rs:2:5
  |
2 | use pac::pwm::CH;
  |     ^^^ use of undeclared crate or module `pac`

[...]
error: could not compile `rp2040-hal` (lib) due to 255 previous errors; 4 warnings emitted
```

Probably all fixable by adjusting those `use` statements, but is it worth the effort?

ithinuel:
Since our MSRV has edition 2021 support anyway, do we have any benefits from sticking with 2018?

jannic:
I don't think so. It's probably just that nobody did the conversion, because it doesn't give us significant benefits either.

Maybe https://doc.rust-lang.org/stable/edition-guide/rust-2021/panic-macro-consistency.html could add some additional overhead because `panic!` now always involves string formatting. But my gut feeling is that, as we already have panic statements which use string formatting, the overhead will be very small, if any.

PR #648, opened 2023-07-12
Rename `dma::SingleChannel::listen_irq*` to `enable_irq*`

nilclass:
My motivation for this change is that the first time I encountered the `listen_irq*` methods, the name invoked a misleading picture in my mind:
`channel.listen_irq0()` sounds like the channel is "listening" for the IRQ, and will do something once it triggers.

That's not the case though: calling this method will cause the channel to *trigger* the interrupt -- some other part of the program must do the listening.

Reasons to call it `enable_irq*` / `disable_irq*` instead:
- The corresponding function in pico-sdk is called `dma_channel_set_irqX_enabled`
- The INTEx register documentation also uses this terminology: "Interrupt Enables for IRQ X"


jannic (approved):
LGTM, and the rationale is convincing.

PR #651, opened 2023-07-16
Migrate rp2040-hal to edition 2021

ithinuel:
In addition to the fairly trivial migration to edition 2021, this commit provides some sanitation of macro hygiene and brings some consistencies to rp2040_pac references.


jannic:
Just curious: does this change bring any visible advantage for users of the crate, or is it "only" improving internal consistency?

ithinuel:
I am not aware of any user visible impact.

PR #654, opened 2023-07-21
[spi] Make SPI set_format accept frame format

### LINKED ISSUE #241: SPI formats
As an electronics beginner, I wanted to drive some ws2812b's with an SPI pin, not realizing that the only supported format right now is the Motorola one.

I was able to modify the set_format function myself to (hardcode) the TI format in the FRF register.

I would really like a way to change it with the set_format function directly, so that my christmas tree can shine more engonomically :smile:

NelsonAPenn:
Allow selecting Texas Instruments SSI and National Semiconductor Microwire frame formats. Modify interface to only configure clock phase and polarity when applicable (if frame format is Motorola SPI).

Additional notes:

- resolves #241 
- contains breaking changes to both `set_format` and SPI init functions
- removes generic <M: Into<Mode>>
    - I could add a generic Mode to the `FrameFormat` enum, but I believe it adds more complexity than it is worth (`FrameFormat::MotorolaSpi(mode.into)` is concise enough

### DIFF
--- a/rp2040-hal/src/spi.rs
+++ b/rp2040-hal/src/spi.rs
@@ -38,7 +38,6 @@ use crate::{resets::SubsystemReset, typelevel::Sealed};
 mod pins;
 pub use pins::*;
 
-/// Spi mode
 
 impl From<embedded_hal::spi::Mode> for FrameFormat {
     fn from(f: embedded_hal::spi::Mode) -> Self {
@@ -53,6 +52,7 @@ impl From<&embedded_hal::spi::Mode> for FrameFormat {
 }
 
 /// SPI frame format
+#[derive(Clone, Copy, PartialEq, Eq)]
 pub enum FrameFormat
 {
     /// Motorola SPI format. See section 4.4.3.9 of RP2040 datasheet.
--- a/rp2040-hal/src/spi.rs
+++ b/rp2040-hal/src/spi.rs
@@ -53,6 +53,17 @@ impl From<&embedded_hal::spi::Mode> for Mode {
     }
 }
 
+/// SPI frame format
+pub enum FrameFormat
+{
+    /// Motorola SPI format. See section 4.4.3.9 of RP2040 datasheet.
+    MotorolaSpi(Mode),
+    /// Texas Instruments synchronous serial frame format. See section 4.4.3.8 of RP2040 datasheet.
+    TexasInstrumentsSynchronousSerial,
+    /// National Semiconductor Microwire frame format. See section 4.4.3.14 of RP2040 datasheet.
+    NationalSemiconductorMicrowire,
+}
+
 #[cfg(feature = "eh1_0_alpha")]
 impl From<eh1_0_alpha::spi::Mode> for Mode {
     fn from(f: eh1_0_alpha::spi::Mode) -> Self {
@@ -217,14 +228,29 @@ impl<D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<Disabled, D, P, DS> {
     }
 
     /// Set format and datasize
-    fn set_format(&mut self, data_bits: u8, mode: Mode) {
+    fn set_format(&mut self, data_bits: u8, frame_format: FrameFormat) {
         self.device.sspcr0.modify(|_, w| unsafe {
             w.dss()
                 .bits(data_bits - 1)
-                .spo()
-                .bit(mode.0.polarity == Polarity::IdleHigh)
-                .sph()
-                .bit(mode.0.phase == Phase::CaptureOnSecondTransition)
+                .frf()
+                .bits(match &frame_format {
+                    FrameFormat::MotorolaSpi(_) => 0x00,
+                    FrameFormat::TexasInstrumentsSynchronousSerial => 0x01,
+                    FrameFormat::NationalSemiconductorMicrowire => 0x10,
+                });
+
+            /*
+             * Clock polarity (SPO) and clock phase (SPH) are only applicable to
+             * the Motorola SPI frame format.
+             */
+            if let FrameFormat::MotorolaSpi(ref mode) = frame_format
+            {
+                w.spo()
+                    .bit(mode.0.polarity == Polarity::IdleHigh)
+                    .sph()
+                    .bit(mode.0.phase == Phase::CaptureOnSecondTransition);
+            }
+            w
         });
     }
 
@@ -242,14 +268,14 @@ impl<D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<Disabled, D, P, DS> {
         resets: &mut RESETS,
         peri_frequency: F,
         baudrate: B,
-        mode: Mode,
+        frame_format: FrameFormat,
         slave: bool,
     ) -> Spi<Enabled, D, P, DS> {
         self.device.reset_bring_down(resets);
         self.device.reset_bring_up(resets);
 
         self.set_baudrate(peri_frequency, baudrate);
-        self.set_format(DS, mode);
+        self.set_format(DS, frame_format);
         self.set_slave(slave);
         // Always enable DREQ signals -- harmless if DMA is not listening
         self.device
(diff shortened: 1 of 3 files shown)


rp2040-hal/src/spi.rs
```diff
@@ -263,22 +285,26 @@ impl<D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<Disabled, D, P, DS> {
     }
 
     /// Initialize the SPI in master mode
-    pub fn init<F: Into<HertzU32>, B: Into<HertzU32>, M: Into<Mode>>(
+    pub fn init<F: Into<HertzU32>, B: Into<HertzU32>>(
```
ithinuel, line 288: Why removing the generic over Mode instead of updating it?
```suggestion
    pub fn init<F: Into<HertzU32>, B: Into<HertzU32, M: Into<FrameFormat>>>(
```

rp2040-hal/src/spi.rs
```diff
@@ -53,6 +54,16 @@ impl From<&embedded_hal::spi::Mode> for Mode {
     }
 }
 
+/// SPI frame format
+pub enum FrameFormat {
```
ithinuel, line 58: Can we make it derive a few useful things like
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
#[cfg_attr(feature = "defmt", derive(defmt::Format))]
pub enum FrameFormat {
```

rp2040-hal/src/spi.rs
```diff
@@ -53,6 +54,16 @@ impl From<&embedded_hal::spi::Mode> for Mode {
...
+    /// Motorola SPI format. See section 4.4.3.9 of RP2040 datasheet.
+    MotorolaSpi(Mode),
```
ithinuel, line 60: `Mode` is a _newtype_ around `embedded_hal::spi::Mode`.

Can we make `FrameFormat` implement `From<_>` for the embedded_hal's `spi::Mode` and ditch our _newtype_?

```rust
impl From<embedded_hal::spi::Mode> for FrameFormat {
    // …
}

#[cfg(feature = "eh1_0_alpha")]
impl From<eh1_0_alpha::spi::Mode> for FrameFormat {
    // …
}
```

This would make the changes to the doc and examples unnecessary.

rp2040-hal/src/spi.rs
```diff
@@ -263,22 +285,26 @@ impl<D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<Disabled, D, P, DS> {
...
+        self.init_spi(resets, peri_frequency, baudrate, frame_format, false)
     }
 
     /// Initialize the SPI in slave mode
-    pub fn init_slave<M: Into<Mode>>(self, resets: &mut RESETS, mode: M) -> Spi<Enabled, D, P, DS> {
+    pub fn init_slave(
+        self,
+        resets: &mut RESETS,
+        frame_format: FrameFormat,
+    ) -> Spi<Enabled, D, P, DS> {
```
ithinuel, line 303: Similarly here
```suggestion
    pub fn init_slave<M: Into<FrameFormat>>(
        self,
        resets: &mut RESETS,
        frame_format: M,
    ) -> Spi<Enabled, D, P, DS> {
```

rp2040-hal/src/spi.rs
```diff
@@ -263,22 +285,26 @@ impl<D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<Disabled, D, P, DS> {
...
         // Use dummy values for frequency and baudrate.
         // With both values 0, set_baudrate will set prescale == u8::MAX, which will break if debug assertions are enabled.
         // u8::MAX is outside the allowed range 2..=254 for CPSDVSR, which might interfere with proper operation in slave mode.
-        self.init_spi(resets, 1000u32.Hz(), 1000u32.Hz(), mode.into(), true)
+        self.init_spi(resets, 1000u32.Hz(), 1000u32.Hz(), frame_format, true)
```
ithinuel, line 307:
```suggestion
        self.init_spi(resets, 1000u32.Hz(), 1000u32.Hz(), frame_format.into(), true)
```

rp2040-hal/src/spi.rs
```diff
@@ -263,22 +285,26 @@ impl<D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<Disabled, D, P, DS> {
...
 
     /// Initialize the SPI in master mode
-    pub fn init<F: Into<HertzU32>, B: Into<HertzU32>, M: Into<Mode>>(
+    pub fn init<F: Into<HertzU32>, B: Into<HertzU32>>(
         self,
         resets: &mut RESETS,
         peri_frequency: F,
         baudrate: B,
-        mode: M,
+        frame_format: FrameFormat,
```
ithinuel, line 293:
```suggestion
        frame_format: M,
```

rp2040-hal/src/spi.rs
```diff
@@ -263,22 +285,26 @@ impl<D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<Disabled, D, P, DS> {
...
     ) -> Spi<Enabled, D, P, DS> {
-        self.init_spi(resets, peri_frequency, baudrate, mode.into(), false)
+        self.init_spi(resets, peri_frequency, baudrate, frame_format, false)
```
ithinuel, line 295:
```suggestion
        self.init_spi(resets, peri_frequency, baudrate, frame_format.into(), false)
```

rp2040-hal/src/spi.rs
```diff
@@ -217,14 +228,25 @@ impl<D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<Disabled, D, P, DS> {
...
-            w.dss()
-                .bits(data_bits - 1)
-                .spo()
-                .bit(mode.0.polarity == Polarity::IdleHigh)
-                .sph()
-                .bit(mode.0.phase == Phase::CaptureOnSecondTransition)
+            w.dss().bits(data_bits - 1).frf().bits(match &frame_format {
+                FrameFormat::MotorolaSpi(_) => 0x00,
+                FrameFormat::TexasInstrumentsSynchronousSerial => 0x01,
+                FrameFormat::NationalSemiconductorMicrowire => 0x10,
```
ithinuel, line 236: There seem to be a missing enum in the PAC for `FRF`.
https://github.com/rp-rs/rp2040-pac/issues/74 created to track that.

Not blocking here but worth updating in the future.

ithinuel (changes requested):

rp2040-hal/src/spi.rs
jannic: Is this API really better? Personally, I don't like those `.into()` conversions.

Sure, they are sometimes necessary, because rust prefers to be explicit about type conversions, and that's good. But if we could avoid them in the common case, it'd be even better.

Or, from a different perspective: Are there other usages of the API that become nicer with your suggestion?

ithinuel: Well, my suggestion removes the manual wrapping and the `.into()` that this PR introduces.

So, IMHO:
```rust
    let mut spi = spi.init(
        &mut pac.RESETS,
        16.MHz(),
        embedded_hal::spi::MODE_0,
    );
```
is better/more readable than
```rust
    let mut spi = spi.init(
        &mut pac.RESETS,
        16.MHz(),
        hal::spi::FrameFormat::MotorolaSpi(embedded_hal::spi::MODE_0.into()),
    );
```

ithinuel:
Also, I always forget to add that but:

**Thank you for your contribution! Your work is greatly appreciated** 😃

rp2040-hal/src/spi.rs
jannic: Good point, the modes from EH are probably much more commonly used than other two, so being able to use them without explicitly specifying Motorola frame format is nice.

NelsonAPenn: I think it could make sense to implement the conversions from the mode structs embedded_hal crate, but only because they do not apply to the other frame formats. It might not be clear from a usage standpoint why this is the case.

Perhaps the inconvenience factor might instead be begging the question of why we are intent on instantiating this by converting from embedded_hal::spi::MODE_0 every time.

I'm wondering if an alternate approach of making it easier to get the default value by providing a constant / implementing Default would be acceptable. Users could very easily see that there is a default format and that that format is MotorolaSpi with the usual values of phase and polarity, as opposed to having to understand that there is conversion magic from the embedded_hal structs and why it ends up being MotorolaSpi format when it does.

```rust
// not necessarily what I'm saying it should be named, but something like 
    let mut spi = spi.init(
        &mut pac.RESETS,
        16.MHz(),
        hal::spi::DEFAULT_FRAME_FORMAT,
    );
    // or
   let mut spi = spi.init(
        &mut pac.RESETS,
        16.MHz(),
        hal::spi::FrameFormat::default(),
    );
    // or
    let mut spi = spi.init(
        &mut pac.RESETS,
        16.MHz(),
        Default::default(),
    );
```

What do you think?

jannic: I don't think there is one mode which could be considered "usual values of phase and polarity". And getting this wrong can be difficult to debug. So I'd prefer to select the desired mode explicitly.

rp2040-hal/src/spi.rs
NelsonAPenn: I added most of these, but if you want Debug and defmt::Format, I'll have to bring the newtype back.

ithinuel (approved):

PR #656, opened 2023-07-23
bump to rp2040-pac next gen

ithinuel:


jannic:
As mentioned on matrix, I already tested a version of rp-hal ported to the updated pac.
It's available at https://github.com/jannic/rp-hal/tree/experimental-update-pac
Not yet a PR because I need to find a good solution for enabling critical-section-impl for all the examples

ithinuel:
IMHO, `rp2040-pac/critical-section` should be a regular feature dependency of `rp2040-hal`(not optional) since we use `critical-section` internally anyways.

jannic:
That's true for critical-section. But we still need a solution for crtitical-section-impl, which is needed for all our examples, but shouldn't be enabled by default.

jannic:
Opened https://github.com/rp-rs/rp-hal/issues/657 to track the critical-section issue.

jannic:
I think this was all included in #662

PR #660, opened 2023-08-01
USB host support (experimental)

nilclass:
As mentioned on matrix, I've been looking into USB host support.
While it is by no means finished or even usable, I want to share my progress and findings.

My initial draft (now removed from this branch) can be found in this commit: https://github.com/rp-rs/rp-hal/commit/0cf375f787c83a6d355dadb50e6eff5d087a5d30
It works to the point where a keyboard can be attached and the LEDs can be controlled.

Since then I've tried to translate that draft into a more reasonable, interrupt driven implementation.

It's tricky, since I'm not certain yet which parts of the USB host stack should go into the HAL and which parts should go into a more generic package.
I'm thinking that embedded USB host support in general could mirror the way things work on the device side:
For devices there is the `usb-device` package, which provides the glue between different USB functions (usbd-serial, usbd-audio, ...) on one side and the specific USB device hardware on the other side.
For hosts, a similar package would provide the glue between function drivers and the specific USB host hardware.
There is some prior art in that direction: the aptly named [usb-host crate](https://crates.io/crates/usb-host) defines such an interface, although it is in a very early state.
For now I decided not to target that crate though, but instead do my own experiments here: [usbh](https://github.com/nilclass/usbh).

One difference between the two is that the `usb-host` [API for control transfers](https://docs.rs/usb-host/0.1.3/usb_host/trait.USBHost.html#tymethod.control_transfer) is blocking, while my attempt is using only interrupts and callbacks on the driver side to.
In my version also the state machine for control transfers would not be part of the HAL, but part of the common usbh package. That means the HAL only provides methods to issue `SETUP`, `DATA IN` and `DATA OUT` transactions on the control endpoint. The usbh package then calls these methods as needed for a control in / control out transfer.

### How to try it out?

This takes a couple of pieces:
1. Changes to the PAC: I'm currently using [this branch](https://github.com/nilclass/rp2040-pac/tree/usb-host-support-old), which is based on an older version of the PAC. The changes there will need some adjustments to work with the latest PAC. However, the latest PAC did not work for me, since it requires a bunch of other changes to the HAL as well (I tried basing my work on the [experimental-update-pac](https://github.com/jannic/rp-hal/tree/experimental-update-pac) branch by @jannic but ran into a bunch of errors which I did not want to figure out at the time)
2. The `usbh` package https://github.com/nilclass/usbh (it's not published yet, so you'll have to use it from git or from a local clone)
3. The example application, which can be found here: https://github.com/nilclass/rp-hal-usb-host-example It reads the device descriptor and the first configuration descriptor and logs them, then goes on assuming the attached device is a keyboard (it doesn't actually interpret the descriptor yet to check) and tries to poll for keypresses (not working yet)

You'll likely have to adjust some paths in various Cargo.toml files to get this to work. The example also requires `rp2040-monotonic`, I'm using a local copy of that as well, with the Cargo.toml adjusted so that `rp2040-hal` and `rp2040-monotonic` both depend on the same local `rp2040-pac` copy.

### What is working, what is missing?

A bunch of things are working, but nothing completely useful yet :stuck_out_tongue: 
- control transfers: as mentioned above, the usbh package facilitates control transfers ([code](https://github.com/nilclass/usbh/blob/f3d4f7f13c00c56ce5a2bdcc707d7c8a168e5e52/src/transfer.rs))
- enumeration: when a device is detected it issues a reset, attempts a GetDescriptor control transfer on address 0, then resets again and assigns an address to the device. This functionality is also part of the usbh package, not HAL specific ([code](https://github.com/nilclass/usbh/blob/f3d4f7f13c00c56ce5a2bdcc707d7c8a168e5e52/src/enumeration.rs))

Currently I am trying to get interrupt endpoints to work, but it is a bit tricky.

Besides defining a driver interface, another big thing that would be very useful is hub support.

This is a big topic, and I am not sure I can get it to a satisfactory state on my own, so if somebody wants to collaborate, please reply here or ping me on matrix :)


nilclass:
@tommy-gilligan thanks, I've picked the commit over here.
Unfortunately I'm having some trouble with the example currently that I need to figure out: it works when I first connect a keyboard, but when I remove it and reattach it, it's no longer recognized. I haven't had the time to debug this yet (without logging it's a bit more tricky :stuck_out_tongue:).
Does it work for you?

Alternatively the example could also go into the [rp-pico examples in the rp-hal-boards repo](https://github.com/rp-rs/rp-hal-boards/tree/main/boards/rp-pico/examples), some of those already depend on defmt, so in that case we could keep the logging.

thejpster:
I've been using cotton-usb (https://github.com/pdh11/cotton) with some success.

PR #667, opened 2023-08-08
Add function to enable multiple PWM slices at the same time

jlpettersson:
This is needed to be able to have e.g. 3 PWM slices that are in phase, but may use different duty cycles.

Example usage: `pwm_slices.enable_simultaneous(0b0000_0111);`

Demo
----

Phases between slices before this fix, using
```
    pwm0.enable();
    pwm1.enable();
```

IMG-6377

Phases between slices after this fix, using
```
pwm_slices.enable_simultaneous(0b0000_0111);
```

IMG-6378


rp2040-hal/src/pwm/mod.rs
```diff
@@ -501,17 +501,17 @@ impl Slices {
...
-    //     }
+    /// Enable multiple slices at the same time to make their counters sync up.
+    ///
+    /// You still need to call `slice` to get an actual slice
+    pub fn enable_simultaneous(&mut self, bits: u8) {
+        // Enable all slices at the same time
+        unsafe {
+            let _ = &(*pac::PWM::ptr())
+                .en
+                .modify(|r, w| w.bits(((r.bits() as u8) | bits) as u32));
```
jannic, line 512: To both simplify the code and make it atomic, I'd suggest the following alternative:
```suggestion
            let reg = self._pwm.en.as_ptr();                                                  
            write_bitmask_set(reg, bits as u32);           
```

jlpettersson: Sounds good to me. I mostly uncommented the function that was in the comment :)

jlpettersson:
@jannic I made the mistake to use GitHub GUI to apply your suggestions. That did not end well. So I ended up with creating a new PR for this change, https://github.com/rp-rs/rp-hal/pull/667

PR #670, opened 2023-08-16
Add function to clear PIO fifos

ThadHouse:
The C SDK has pio_sm_clear_fifos, which clears the FIFOs without disrupting the internal state machine state. This function is needed for porting one of my projects from C to Rust.


jannic:
The CI error doesn't look related to your changes. Probably some dependency bumped its MSRV in a minor release.

jannic (approved):

PR #674, opened 2023-08-22
Bump MSRV to 1.64

### LINKED ISSUE #673: MSRV check fails
From https://github.com/rp-rs/rp-hal/actions/runs/5933769438/job/16092767596?pr=670:
```
info: running `cargo build --examples --no-default-features --features critical-section-impl` on rp2040-hal (2/13)
error: package `indexmap v2.0.0` cannot be built because it requires rustc 1.64 or newer, while the currently active rustc version is 1.63.0
error: process didn't exit successfully: `/home/runner/.rustup/toolchains/1.63-x86_64-unknown-linux-gnu/bin/cargo build --examples --manifest-path rp2040-hal/Cargo.toml --no-default-features --features critical-section-impl` (exit status: 101)
Error: Process completed with exit code 1.
```

jannic:
Closes #673

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -82,7 +82,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: 1.63
+          toolchain: 1.64
           target: thumbv6m-none-eabi
       - name: Install cargo-hack
         uses: baptiste0928/cargo-install@v2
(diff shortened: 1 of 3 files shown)


thejpster (approved):

jannic:
> What kind of version bump does this warrant?

There's no consensus: https://github.com/rust-lang/api-guidelines/discussions/231

Personally, I think the MSRV is not very meaningful if a MSRV bump isn't considered semver breaking. Otherwise, a library would need to pin all its dependencies to specific versions to be able to specify a MSRV itself. Otherwise, a dependency might do a MSRV bump at any time, breaking the MSRV guarantee of the depenent crate.

But at the same time, I agree with https://github.com/rust-lang/api-guidelines/discussions/231#discussioncomment-6361907.

Fortunately, for this MSRV bump it doesn't matter much, as the first release to include it will be a semver breaking version, 0.9.0, anyway.

Issue #659, opened 2023-07-29
Going into dormant state


0xa10:
Hi, in my application I'd like the RP2040 to completely power off in certain cases, so it consumes minimal current. 
According to several sources its possible to go into a dormant state in which the power consumption is <1mA.

Ive looked around and it seems this involves disabling interrupts and then disabling several of the system clocks. 
I also saw this PR https://github.com/rp-rs/rp-hal/pull/613 which seems to have a halt function, but that doesnt put the main core to sleep as far as I could tell. Using WFE/WFI even with interrupts disabled still puts my Pico at ~20mA of draw. A timer sleep puts it at around 9mA. 

Is there any way using the existing HAL code to go into deep sleep?

ithinuel:
Hello,
Out of curiosity, have you tried the c-sdk's [`hello_dormant`](https://github.com/raspberrypi/pico-playground/blob/master/sleep/hello_dormant/hello_dormant.c) to verify what could be achieved with your setup?

20mA seem to be quite a lot as the bootsel mode typically runs around 14mA.
This makes me think that your MCU may not be sleeping at all.

From reading the rp2040 datasheet, it seems to me that core1 enters deepsleep at boot (WFE + SCR.SLEEPDEEP set).
The first core on the other hand needs to be configured (setting SCR.SLEEPDEEP) to behave like so.

Based on what I can see from the listed source, the dormant mode is not linked to deepsleep/WFI/WFE.
It's literally stopping the clock.
https://github.com/raspberrypi/pico-extras/blob/master/src/rp2_common/pico_sleep/sleep.c
https://github.com/raspberrypi/pico-sdk/blob/6a7db34ff63345a7badec79ebea3aaef1712f374/src/rp2_common/hardware_xosc/xosc.c#L52

Anyways, I should have started with this:

The `rp2040-hal` implements [`CrystalOscillator::dormant()`](https://docs.rs/rp2040-hal/latest/rp2040_hal/xosc/struct.CrystalOscillator.html#method.dormant) which should (partially) achieve what you are looking for.
One could argue this is incomplete as there does not seem to be a way to recover a `CrystalOscillator<Stable>` from a `CrystalOscillator<Dormant>` to which I'd reply that all contributions are more than welcome 😃.

0xa10:
Thank you. I will look into that and make sure to submit a PR if I end up writing recovery logic 🙂

0xa10:
[...]

PR #676, opened 2023-08-28
Add RTC examples & expand RealTimeClock & ClockManager capabilities.

ithinuel:


rp2040-hal/examples/rtc_irq_example.rs
```diff
@@ -0,0 +1,176 @@
...
+            *LED_AND_RTC = GLOBAL_SHARED.borrow(cs).take();
+        });
+    }
+
+    // Need to check if our Option<LedAndButtonPins> contains our pins
+    if let Some(led_and_rtc) = LED_AND_RTC {
+        // borrow led and rtc by *destructuring* the tuple
+        // these will be of type `&mut LedPin` and `&mut RealTimeClock`, so we don't have
+        // to move them back into the static after we use them
+        let (led, rtc) = led_and_rtc;
```
jannic, line 166: I had to look up why this works. Even if I knew it in the past.
The answer is "default binding modes", https://doc.rust-lang.org/reference/patterns.html#binding-modes

Would it be useful to mention that here, or would it distract too much from the point of this example?

rp2040-hal/examples/rtc_sleep_example.rs
```diff
@@ -0,0 +1,169 @@
...
+    // will jump to the interrupt function when the interrupt occurs.
+    // We do this last so that the interrupt can't go off while
+    // it is in the middle of being configured
+    unsafe {
+        pac::NVIC::unmask(pac::Interrupt::RTC_IRQ);
+    }
+
+    cortex_m::asm::delay(XTAL_FREQ_HZ.to_Hz() / 2);
+    loop {
+        // interrupts handle everything else in this example.
```
jannic, line 149: This comment is a little bit misleading because in this example, the actual work (toggling the LED) is not done from the interrupt handler.

jannic:
I like those examples. Properly configuring the dormant state is non-trivial, so having a working example is really useful.

Perhaps you could add a comment with the results of the measurements you mentioned in https://github.com/rp-rs/rp-hal/issues/659#issuecomment-1694745509 ?

rp2040-hal/src/clocks/mod.rs
```diff
@@ -301,6 +302,22 @@ impl ClocksManager {
             .configure_clock(&self.system_clock, self.system_clock.freq())
     }
 
+    /// Configure the clocks staying ON during deep-sleep.
+    pub fn configure_sleep_en<F0, F1>(&mut self, sleep_en0: F0, sleep_en1: F1)
```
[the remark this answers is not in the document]
ithinuel: I'm not fond of it either.
I prefer this to taking a u32 bit pattern because in its use place it is self explanatory.

I think the "real" good solution would be to implement a bit field with accessors similar to what svd2rust generates.
I was being lazy yesterday :)

rp2040-hal/examples/rtc_irq_example.rs
ithinuel: I'll add a comment above the if-let

rp2040-hal/src/rtc/mod.rs
```diff
@@ -29,8 +29,6 @@ mod filter;
 
 pub use self::filter::DateTimeFilter;
 
-#[cfg_attr(feature = "chrono", path = "datetime_chrono.rs")]
-#[cfg_attr(not(feature = "chrono"), path = "datetime_no_deps.rs")]
```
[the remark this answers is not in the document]
ithinuel: For the motivation of that change in this PR, this feature currently changes the shape of the API.

This makes the rtc examples require conditional compilation to accommodate with the different DateTime creation style.

Following the philosophy of features being additive, it seems more natural that enabling the `chrono` feature _adds_ chrono support rather than _replace_ the local types.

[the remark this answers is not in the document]
jannic: That's a very good point for making the no-deps implementation available unconditionally. But in it's current form, this PR completely removes chrono support. (It leaves the `datetime_chrono.rs` file in place, but there's no longer any code referencing it.)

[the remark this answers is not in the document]
ithinuel: [This block](https://github.com/rp-rs/rp-hal/pull/676/files#diff-c111db841a5d3011d4c1f34c3ab3b11cb01d38f58a5d9484807a8a80d7949d5bR136-R207) should be enough to allow easy conversion between chrono's `NaiveDateTime` and our own `DateTime` type.

rp2040-hal/src/clocks/mod.rs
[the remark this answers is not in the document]
ithinuel: The latest commit should fix this.
It introduces a new dependency which has itself none and is fairly stable.

rp2040-hal/src/rtc/mod.rs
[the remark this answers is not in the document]
VictorKoenders: @jonathanpallant recommended using `chrono` in the previous PR #207, which is why it was in #213

rp2040-hal/examples/rtc_sleep_example.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    // Unmask the RTC IRQ so that the NVIC interrupt controller
+    // will jump to the interrupt function when the interrupt occurs.
+    // We do this last so that the interrupt can't go off while
+    // it is in the middle of being configured
+    unsafe {
+        pac::NVIC::unmask(pac::Interrupt::RTC_IRQ);
+    }
+
+    cortex_m::asm::delay(XTAL_FREQ_HZ.to_Hz() / 2);
```
jonathanpallant, line 143: What does this delay do?

rp2040-hal/src/rtc/mod.rs
```diff
@@ -179,19 +173,26 @@ impl RealTimeClock {
...
+
+    /// Disable the propagation of the alarm to the NVIC.
+    pub fn disable_interrupt(&mut self) {
+        self.rtc.inte.modify(|_, w| w.rtc().clear_bit());
+    }
+
+    /// Clear the interrupt.
     ///
-    /// [`schedule_alarm`]: #method.schedule_alarm
+    /// This should be called every time the `RTC_IRQ` interrupt is triggered of the interrupt will
```
jonathanpallant, line 191:
```suggestion
    /// This should be called every time the `RTC_IRQ` interrupt is triggered or the interrupt will
```

rp2040-hal/examples/rtc_sleep_example.rs
ithinuel: Oh, indeed, that's a debugging left over.

rp2040-hal/src/rtc/mod.rs
ithinuel: Thank you!

ithinuel:
@jonathanpallant do you think the approach proposed here is acceptable?

jonathanpallant (approved):

jannic (approved):

PR #677, opened 2023-08-29
Disable clippy lint needless_pass_by_ref_mut

jannic:
We sometimes use &mut to guarantee exclusive access to some ressource, where the compiler can't see that such a guarantee is needed for soundness. Therefore, warnings about needles &mut references should not be generated.

Without this change, starting from rust 1.73 (currently in beta), the following clippy warnings will be generated:
```
warning: this argument is a mutable reference, but not used mutably 
  --> rp2040-hal/src/rtc/mod.rs:57:17 
   | 
57 |         resets: &mut RESETS, 
   |                 ^^^^^^^^^^^ help: consider changing to: `&RESETS` 
   | 
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut 
   = note: `#[warn(clippy::needless_pass_by_ref_mut)]` on by default 
 
warning: this argument is a mutable reference, but not used mutably 
   --> rp2040-hal/src/uart/peripheral.rs:278:13 
    | 
278 |     device: &mut U, 
    |             ^^^^^^ help: consider changing to: `&U` 
    | 
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut 
 
warning: this argument is a mutable reference, but not used mutably 
  --> rp2040-hal/src/vector_table.rs:43:33 
   | 
43 |     pub fn init(&mut self, ppb: &mut crate::pac::PPB) { 
   |                                 ^^^^^^^^^^^^^^^^^^^^ help: consider changing to: `&crate::pac::PPB` 
   | 
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut 
 
warning: this argument is a mutable reference, but not used mutably 
  --> rp2040-hal/src/vector_table.rs:80:44 
   | 
80 |     pub unsafe fn activate(&mut self, ppb: &mut crate::pac::PPB) { 
   |                                            ^^^^^^^^^^^^^^^^^^^^ help: consider changing to: `&crate::pac::PPB` 
   | 
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut 
 
warning: `rp2040-hal` (lib) generated 4 warnings
```

### DIFF
--- a/rp2040-hal/src/rtc/mod.rs
+++ b/rp2040-hal/src/rtc/mod.rs
@@ -51,6 +51,8 @@ impl RealTimeClock {
     ///
     /// [`ClocksManager`]: ../clocks/struct.ClocksManager.html
     /// [`clocks`]: ../clocks/index.html
+    #[allow(unknown_lints)]
+    #[allow(clippy::needless_pass_by_ref_mut)]
     pub fn new(
         rtc: RTC,
         clock: RtcClock,
--- a/rp2040-hal/src/uart/peripheral.rs
+++ b/rp2040-hal/src/uart/peripheral.rs
@@ -272,6 +272,8 @@ fn calculate_baudrate_dividers(
 }
 
 /// Baudrate configuration. Code loosely inspired from the C SDK.
+#[allow(unknown_lints)]
+#[allow(clippy::needless_pass_by_ref_mut)]
 fn configure_baudrate<U: UartDevice>(
     device: &mut U,
     wanted_baudrate: HertzU32,
--- a/rp2040-hal/src/vector_table.rs
+++ b/rp2040-hal/src/vector_table.rs
@@ -40,6 +40,8 @@ impl VectorTable {
     }
 
     /// Initialise our vector table by copying the current table on top of it
+    #[allow(unknown_lints)]
+    #[allow(clippy::needless_pass_by_ref_mut)]
     pub fn init(&mut self, ppb: &mut crate::pac::PPB) {
         let vector_table = ppb.vtor.read().bits();
         unsafe {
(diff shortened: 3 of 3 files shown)


jannic:
Well that leads to ``error: unknown lint: `clippy::needless_pass_by_ref_mut` `` with the compiler version used by CI.
Therefore, I'll make it a draft PR to be merged later, when we bump our MSRV.

ithinuel:
Having this as a blanket allow on the whole crate seems a bit heavy handed.
Would it add a lot of uglyness to keep it where it is really needed?
I'd expect there not to be a lot of functions taking &mut refs to pac::RESETS or other peripherals.

thejpster:
Do you have an example of where clippy doesn't know we need the mut ref? And can we do something with the value to make it realise we do?

ithinuel:
Any can where we use the mut ref to guaranty exclusive access but in reality we access register with write/modify which only requires a n & to the register

jannic:
Interestingly, only the four locations listed in the ticket trigger this clippy warning. So ithinuel's suggestion to apply the annotation locally instead of globally makes a lot of sense.

Regarding `RESETS`, most peripherals call the macro-generated `reset_bring_up` / `reset_bring_down` functions, which take a `&mut` parameter. So at the caller site, not warning is generated as the `&mut` is obviously needed.
I'm not sure why no warning is triggered in `src/resets.rs`. Maybe it's because the functions implement `trait SubsystemReset`, and functions implementing a trait are exempt from this warning? That would make sense, as the trait could demand a `&mut` because _some_ implementation may need it, and the individual implementation just implements the trait, so its fine if it gets a `&mut` even if it doesn't need it.

9names (approved):

PR #685, opened 2023-09-05
Provide functions to access watchdog scratch registers

jannic:


rp2040-hal/src/watchdog.rs
```diff
@@ -93,6 +97,37 @@ impl Watchdog {
         self.watchdog.ctrl.write(|w| w.enable().bit(bit))
     }
 
+    /// Read a scratch register
+    pub fn read_scratch(&self, idx: u8) -> Result<u32, NoSuchRegister> {
```
jannic, line 101: Request for feedback: Is the use of `idx: u8` with the possibility of a `NoSuchRegister` error a good choice?
Or should the parameter be an `enum` with the possible values, so an error becomes impossible?

ithinuel: Would it make sense to expose them as a sized array? (May require a change in the PAC)

jannic: I considered that, but without a good use case I didn't want to do the PAC modification.
However I tried the `enum` variant and I think it's better than the `idx: u8` one. I'll push the update as soon as I find the time to actually test it.

thejpster (approved):

PR #691, opened 2023-09-16
Removed second division by `refdiv`

vinsynth:
While setting up a custom clock with `refdiv = 2`, I noticed it's post division frequency was half it's expected value; it seems that when making a new `PhaseLockedLoopDevice`, division of the reference frequency by `refdiv` occurs twice, which I assume is unintentional.

### DIFF
--- a/rp2040-hal/src/pll.rs
+++ b/rp2040-hal/src/pll.rs
@@ -183,8 +183,8 @@ impl<D: PhaseLockedLoopDevice> PhaseLockedLoop<Disabled, D> {
         let refdiv = config.refdiv;
         let post_div1 = config.post_div1;
         let post_div2 = config.post_div2;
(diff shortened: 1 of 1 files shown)


9names:
It's not obvious to me where this second division by refdiv occurs, would you mind pointing it out?

jannic:
@9names: A few lines above, `ref_freq_hz` is calculated as:
```
        let ref_freq_hz: HertzU32 = xosc_frequency
            .to_Hz()
            .checked_div(u32::from(config.refdiv))
            .ok_or(Error::BadArgument)?
            .Hz();
```
So that value is already `xosc_frequency / refdiv`. Dividing it again looks wrong.

The patch LGTM.

jannic (approved):

PR #694, opened 2023-09-20
Implement InputPin for all pin configurations via .as_input() function

### LINKED ISSUE #689: Allow reading the input state of pins in any function mode
On the actual chip, all input pins are always wired to the SIO block for inputs, no matter what mode the device is in. This means that no matter the function selected, reading the input state would be valid. It would be very nice if the InputPin trait could be implemented for all function types, as this doesn't break any of rust's safety rules.

I actually ran into a case where I needed to work around this. I was using a PWM Input, and needed to know when the falling edge was triggered. However, for the pin to be readable by PWM, it has to be in FunctionPwm mode, which doesn't allow reading the input. I actually had to use into_unchecked to turn the pin back into an input just so I could read it.

jannic:
Closes #689

### DIFF
--- a/rp2040-hal/src/gpio/mod.rs
+++ b/rp2040-hal/src/gpio/mod.rs
@@ -746,9 +746,9 @@ impl<I: PinId, F: func::Function, P: PullType> Pin<I, F, P> {
 
     /// Return a wrapper that implements InputPin.
     ///
-    /// This allows to read from the pin independen from the selected function.
+    /// This allows to read from the pin independent of the selected function.
     /// Depending on the pad configuration, reading from the pin may not return a
-    /// meaninful result. Check the datasheet.
+    /// meaningful result.
     ///
     /// Calling this function does not set the pad's input enable bit.
     pub fn as_input(&self) -> AsInputPin<I, F, P> {
--- a/rp2040-hal/src/gpio/mod.rs
+++ b/rp2040-hal/src/gpio/mod.rs
@@ -1429,20 +1429,6 @@ mod eh1 {
         }
     }
 
-    impl<I, P> InputPin for Pin<I, FunctionSio<SioOutput>, P>
-    where
-        I: PinId,
-        P: PullType,
-    {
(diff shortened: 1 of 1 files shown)


rp2040-hal/src/gpio/mod.rs
```diff
@@ -743,6 +743,17 @@ impl<I: PinId, F: func::Function, P: PullType> Pin<I, F, P> {
             }
         }
     }
+
+    /// Return a wrapper that implements InputPin.
+    ///
+    /// This allows to read from the pin independen from the selected function.
+    /// Depending on the pad configuration, reading from the pin may not return a
+    /// meaninful result. Check the datasheet.
```
9names, line 751: Nitpick:
```suggestion
    /// This allows to read from the pin independent of the selected function.
    /// Depending on the pad configuration, reading from the pin may not return a
    /// meaningful result.
```
I removed "Check the datasheet" because it's  vague and not actionable - what are they checking for? where should they check?
I couldn't find anything obvious in the datasheet that could tell me when a particular mode would yield a meaningful result, so I would opt to just not say anything about it here.

9names (approved):
Looks good to me, except for the docs on the as_input pin.
See comments for details.

rp2040-hal/src/gpio/mod.rs
jannic: The most important information, IMHO, is this:
"Software control of GPIO, from the single-cycle IO (SIO) block. The SIO function (F5)
must be selected for the processors to drive a GPIO, but the input is always connected,
so software can check the state of GPIOs at any time."
(section 1.4.3, table 3)
So in general, the inputs can be read independent from the configured pin function.

Then there is the input enable flag in the pad control register. You are right, I didn't find any explicit description what happens when that flag isn't set. But from what's written there, I'd assume that the bit must be set to be able to read a meaningful input value.

In any case, I agree with removing "Check the datasheet": We already mention the pad configuration, and these words don't add any useful information.

PR #695, opened 2023-09-26
Rename private function Adc::read to Adc::inner_read

jannic:
The name conflicts with the public trait method OneShot::read. This isn't an issue with correct programs, as rustc just uses the visible function when called from other crates.

But if you forget importing OneShot, you get a confusing error message:

```
error[E0624]: method `read` is private
   --> rp2040-hal/examples/adc.rs:116:45
    |
116 |         let temp_sens_adc_counts: u16 = adc.read(&mut temperature_sensor).unwrap();
    |                                             ^^^^ private method
    |
   ::: /home/jan/rp2040/rp-rs/rp-hal/rp2040-hal/src/adc.rs:302:5
    |
302 |     fn read(&mut self, chan: u8) -> u16 {
    |     ----------------------------------- private method defined here
    |
    = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
15  + use cortex_m::prelude::_embedded_hal_adc_OneShot;
    |
```

Note that the referenced method in line 302 is not the one that should be called, and also has different parameters.

(This was noticed in https://github.com/rp-rs/rp-hal-boards/issues/42)

### DIFF
--- a/rp2040-hal/src/adc.rs
+++ b/rp2040-hal/src/adc.rs
@@ -299,7 +299,7 @@ impl Adc {
         }
     }
 
-    fn read(&mut self, chan: u8) -> u16 {
+    fn inner_read(&mut self, chan: u8) -> u16 {
         while !self.device.cs.read().ready().bit_is_set() {
             cortex_m::asm::nop();
         }
@@ -327,7 +327,7 @@ where
     fn read(&mut self, _pin: &mut SRC) -> nb::Result<WORD, Self::Error> {
         let chan = SRC::channel();
 
-        Ok(self.read(chan).into())
+        Ok(self.inner_read(chan).into())
     }
 }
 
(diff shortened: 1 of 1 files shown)


jannic:
For comparison, with this change, the error message becomes:
```
error[E0599]: no method named `read` found for struct `Adc` in the current scope
   --> rp2040-hal/examples/adc.rs:116:45
    |
116 |         let temp_sens_adc_counts: u16 = adc.read(&mut temperature_sensor).unwrap();
    |                                             ^^^^ method not found in `Adc`
    |
   ::: /home/jan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/embedded-hal-0.2.7/src/adc.rs:97:8
    |
97  |     fn read(&mut self, pin: &mut Pin) -> nb::Result<Word, Self::Error>;
    |        ---- the method is available for `Adc` here
    |
    = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
15  + use cortex_m::prelude::_embedded_hal_adc_OneShot;
    |
```

ithinuel (approved):

PR #696, opened 2023-09-28
Add example using GPIO DynPin

9names:
#585 changed how GPIO worked, and the changelog hinted that it made it easier to use Pins in arrays but did not include any examples, doc-examples, or code snippets that tell you how to do it.
This example should at least provide some hints to new embedded rustaceans on how to use this feature.

### DIFF
--- a/rp2040-hal/examples/gpio_dyn_pin_array.rs
+++ b/rp2040-hal/examples/gpio_dyn_pin_array.rs
@@ -15,10 +15,7 @@
 #![no_std]
 #![no_main]
 
-use hal::gpio::DynPinId;
-use hal::gpio::FunctionSioOutput;
-use hal::gpio::Pin;
-use hal::gpio::PullNone;
+use hal::gpio::{DynPinId, FunctionSioOutput, Pin, PullNone, PullUp};
 // Ensure we halt the program on panic (if we don't mention this crate it won't
 // be linked)
 use panic_halt as _;
--- a/rp2040-hal/examples/gpio_dyn_pin_array.rs
+++ b/rp2040-hal/examples/gpio_dyn_pin_array.rs
@@ -18,7 +18,7 @@
 use hal::gpio::DynPinId;
 use hal::gpio::FunctionSioOutput;
 use hal::gpio::Pin;
-use hal::gpio::PullDown;
+use hal::gpio::PullNone;
 // Ensure we halt the program on panic (if we don't mention this crate it won't
 // be linked)
 use panic_halt as _;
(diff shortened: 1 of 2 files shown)


jannic:
This needs an entry in Cargo.toml:
```
[[example]]
name = "gpio_dyn_pin_array"
required-features = ["critical-section-impl"]
```

rp2040-hal/examples/gpio_dyn_pin_array.rs
```diff
@@ -0,0 +1,120 @@
...
+        pac.IO_BANK0,
+        pac.PADS_BANK0,
+        sio.gpio_bank0,
+        &mut pac.RESETS,
+    );
+
+    // To put pins into an array we have to convert them to Dynamically Typed pins.
+    // This means they'll carry their pin and bank numbers around with them at run time,
+    // rather than relying on the Type of the pin to track that.
+    let mut pinarray: [Pin<DynPinId, FunctionSioOutput, PullDown>; 4] = [
```
jannic, line 94: `PullNone` would be better for push-pull outputs.

9names: Agreed. But that's actually the value of what it's turned into by calling into_dyn_pin - not sure if that is unintentional, but it sure feels weird to have to change the pulls again after converting it.

jannic: `PullDown` is the default pull mode, that what `pins.gpio2` starts with. Neither `into_push_pull_output()` nor `into_dyn_pin()` changes that.

Possible alternatives are:
- `pins.gpio2.into_push_pull_output().into_pull_type().into_dyn_pin()`
- `pins.gpio2.reconfigure().into_dyn_pin()`

9names: `.into_pull_type()` seems to have done the trick, but, once again, feels weird to have to say it .

jannic: Why does it feel weird? Maybe unnecessarily verbose. But it is a nicely orthogonal API, where each of the `_into*` functions changes one of the generic parameters. If you want to change multiple of them at once, there's `reconfigure()`.

jannic (approved):

rp2040-hal/examples/gpio_dyn_pin_array.rs
9names: > PullNone would be better for push-pull outputs.
> That's usually the best choice

This, basically, is what makes me feel a bit strange about it.
As you say, it's not like the Dyn behaviour is different. It's just that the base behaviour was unexpected.

> there's reconfigure().

True. I guess I could write it like
```rust
let mut pinarray: [Pin<DynPinId, FunctionSioOutput, PullNone>; 4] = [
        pins.gpio2.reconfigure().into_dyn_pin(),
        pins.gpio3.reconfigure().into_dyn_pin(),
        pins.gpio4.reconfigure().into_dyn_pin(),
        pins.gpio5.reconfigure().into_dyn_pin(),
    ];
```
or
```rust
let mut pinarray: [Pin<_, FunctionSioOutput, PullNone>; 4] = [
        pins.gpio2.reconfigure().into_dyn_pin(),
        pins.gpio3.reconfigure().into_dyn_pin(),
        pins.gpio4.reconfigure().into_dyn_pin(),
        pins.gpio5.reconfigure().into_dyn_pin(),
    ];
```

but while that makes it neater, it's now more magic (at least for someone new to the chip, the hal, and rust).

9names: I added another line showing use of reconfigure. I'm happy enough with this now.

PR #699, opened 2023-10-18
Improve Spi device/peripheral docs

fu5ha:
Documents the types of the `Spi` struct and brushes up a few other docs.

### DIFF
--- a/rp2040-hal/src/spi.rs
+++ b/rp2040-hal/src/spi.rs
@@ -1,5 +1,8 @@
 //! Serial Peripheral Interface (SPI)
 //!
+//! [`Spi`] is the main struct exported by this module, representing a configured Spi bus. See its
+//! docs for more information on its type parameters.
+//!
 //! See [Chapter 4 Section 4](https://datasheets.raspberrypi.org/rp2040/rp2040_datasheet.pdf) for more details
 //!
 //! ## Usage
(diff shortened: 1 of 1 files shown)


ithinuel (approved):

jannic:
Very nice docs, thanks a lot!

rp2040-hal/src/spi.rs
```diff
@@ -16,7 +19,11 @@
 //! let sclk = pins.gpio2.into_function::<FunctionSpi>();
 //! let mosi = pins.gpio3.into_function::<FunctionSpi>();
 //!
-//! let spi = Spi::<_, _, _, 8>::new(peripherals.SPI0, (mosi, sclk)).init(&mut peripherals.RESETS, 125_000_000u32.Hz(), 16_000_000u32.Hz(), MODE_0);
+//! let spi_device = peripherals.SPI0;
+//! let spi_pin_layout = (mosi, sclk);
+//!
+//! let spi = Spi::new(spi_device, spi_pin_layout)
```
jannic, line 25: This still fails to compile:
```
help: consider giving `spi` an explicit type, where the value of const parameter `DS` is specified
```
That's a little bit unfortunate, as the type annotation clutters the code. Any suggestion on how to express this in a more readable way would be welcome!
```suggestion
//! let spi = Spi::<_, _, _, 8>::new(spi_device, spi_pin_layout)
```

jannic:
Don't worry about the CI error in "Verify build on MSRV". That's known, not caused by this change, and already fixed in #698.
But the failing doc comment in "Build and Test check" should be repaired before merging this.

thejpster:
#698 is now merged, so this just needs a rebase to fix that test, right?

rp2040-hal/src/spi.rs
fu5ha: I made https://github.com/fu5ha/rp-hal/pull/1 on top of this which is unfortunately a breaking change but would improve the type inference ergonomics for this. It's a bit unfortunate Rust can't infer the type default as-is, but yeah.

PR #701, opened 2023-10-19
Full support for entering and exiting DORMANT mode

andrewh42:
I noticed that there were the beginnings of some support for the RP2040 dormant lowest-power sleep but that several loose ends remained.

I've taken care not to make any breaking changes in this pull request to maximise its prospects. However, I think an interesting future change would be to add an addiitional `pll::State` to represent a powered down PLL. This would streamline the process of restarting the PLLs following dormancy.

Looking forward to your feedback and hopefully an approved pull request. I'm keen to hear of any areas that don't meet rp-rs standards or where I've done something that isn't idiomatic - apologies in advance for any crimes against these!

Andrew


ithinuel (changes requested):
Thank you!

I had started doing things in that direction but never got to complete it 🤦. 
Because the Dormant mode doesn't really need to be represented (the device does not run while in dormant mode).

I think `Initialized` and `Dormant` can be merged into a `Unstable` state.

You can find my WIP here: https://github.com/ithinuel/rp-hal/tree/clock_dormant_recovery

andrewh42:
Thanks @ithinuel, that's a good suggestion. I've replaced the XOSC `Initialized` and `Dormant` states with an `Unstable` state. And ROSC no longer changes state on after waking.

I'm not convinced of the benefit of adding `StableOscillatorTokenAlreadyIssuedError` however: it seems overly picky to return that error given we can only redeem the token once via `get_stable()`, regardless of how many times we request a token. Am I missing something here?

andrewh42:
Hi @ithinuel, is there anything I can do to help progress this PR?

rp2040-hal/src/xosc.rs
```diff
@@ -184,6 +181,7 @@ impl CrystalOscillator<Stable> {
             w
         });
 
-        self.transition(Dormant)
+        let freq_hz = self.state.freq_hz;
+        self.transition(Unstable { freq_hz })
```
jannic, line 185: In case the system clock is currently running from the ROSC, transitioning XOSC to dormant will not stop the processor. Am I missing something, or would this function just return `CrystalOscillator<Unstable>` even though the oscillator is dormant?

I'm not sure if there are valid use cases for calling this function in such a way. If not, perhaps just mention this behavior in the safety comment? ("This should only be called if this oscillator is the clock source for the system clock" or similar)?

jannic:
I never used the dormant state and didn't think about it enough to decide if these changes are sufficient to make dormant mode usable. From reading the diffs I get the impression that the changes make sense, and I don't see anything obviously wrong. @ithinuel, as you already worked on your own implementation of dormant support, you are probably more qualified to review this merge request. But if you don't have time to do a review, I could plug together some test setup on a breadboard to try out the example, and approve the PR if everything works as expected.

@andrewh42, you wrote that you avoided breaking changes to make merging easier. As the next release probably will include breaking changes anyway, are there any improvements you'd want to add?

rp2040-hal/src/xosc.rs
andrewh42: Yes, good point @jannic. That's my understanding too.

I've added comments to that effect for the rosc and xosc `dormant()` methods.

andrewh42:
> @andrewh42, you wrote that you avoided breaking changes to make merging easier. As the next release probably will include breaking changes anyway, are there any improvements you'd want to add?

Thanks for asking @jannic. I was thinking it could be useful to add an addiitional pll::State to represent a powered down PLL. At present the PLL's `disable()` reconstructs the `Disabled` state by reading the PLL's registers, so that `initialize()` can be used to restart the disabled PLL. The alternative I have in mind is to replace `disable()` with, say, `stop()` and return a new `Stopped` state that  then requires less work to restart. `initialize()` might then return a `Stopped` PLL, in order to keep the code DRY. Giving us the following state transitions when starting up:

    Disabled -> Stopped -> Locking -> Locked

And these transitions when stopping and then restarting (e.g. as part of going dormant):

    Locked -> Stopped -> Locking -> Locked

Does that sound reasonable? Either way, I'd prefer to do this in a separate PR so this one doesn't stick around for too long and gather more complexity. Does that sound OK?

rp2040-hal/src/xosc.rs
jannic: Great!
If it turns out that switching the non-active oscillator to dormant is needed for some use case, we probably need to bring the `Dormant` state back. But I'm fine with merging this as it is. In any useful configuration I see at the moment, the other oscillator can just be disabled, like you do in the example.

rp2040-hal/examples/dormant_sleep.rs
```diff
@@ -0,0 +1,295 @@
...
+);
+
+type RestartedClockAndPlls = (
+    CrystalOscillator<Stable>,
+    PhaseLockedLoop<Locked, PLL_SYS>,
+    PhaseLockedLoop<Locked, PLL_USB>,
+);
+
+/// The button input.
+type ButtonPin = gpio::Pin<gpio::bank0::Gpio14, gpio::FunctionSioInput, gpio::PullNone>;
```
jannic, line 63: I needed to activate a pull-up on gpio14 to make the example work:
```suggestion
type ButtonPin = gpio::Pin<gpio::bank0::Gpio14, gpio::FunctionSioInput, gpio::PullUp>;
```

With that, it works as expected, and I see the current drawn by the pico to alternate between 0.03A and 0.05A (using a cheap USB power monitor with 10mA resolution). Nice!

rp2040-hal/examples/dormant_sleep.rs
```diff
@@ -0,0 +1,295 @@
+//! # DORMANT low-power mode example
+//!
+//! This application demonstrates how to enter and exit the RP2040's lowest-power DORMANT mode
+//! where all clocks and PLLs are stopped.
+//!
+//! Pulling GPIO 14 low (e.g. via a debounced momentary-contact button) alternately wakes the
+//! RP2040 from DORMANT mode and a regular WFI sleep. A LED attached to GPIO 25 (the onboard LED
+//! on the Raspberry Pi Pico) pulses once before entering DORMANT mode and twice before entering WFI sleep.
```
jannic, line 8:
```suggestion
//! on the Raspberry Pi Pico) pulses once before entering DORMANT mode and twice before entering WFI sleep.
//!
//! Note: DORMANT mode breaks the debug connection. You may need to power cycle while pressing the
//! BOOTSEL button to regain debug access to the pico.
```

rp2040-hal/src/xosc.rs
ithinuel: That's right, I didn't think about this case 🤔.
When applied to the core's source, the state transitions visible from the core are: `Running -> Unstable -> Stable`
And from the other source it's `Running -> Dormant -> Unstable -> Stable`.

Would it make sense to have something like:
```rust
fn dormant(self) -> Either<CrystalOscillator<Unstable>, CrystalOscillator<Dormant>>
```

rp2040-hal/src/rosc.rs
```diff
@@ -97,20 +92,21 @@ impl RingOscillator<Enabled> {
...
+    /// After waking up from the DORMANT state, ROSC restarts in approximately 1µs.
     ///
     /// # Safety
     /// This method is marked unsafe because prior to switch the ROSC into DORMANT state,
     /// PLLs must be stopped and IRQs have to be properly configured.
     /// This method does not do any of that, it merely switches the ROSC to DORMANT state.
+    /// It should only be called if this oscillator is the clock source for the system clock.
     /// See Chapter 2, Section 16, §5) for details.
-    pub unsafe fn dormant(self) -> RingOscillator<Dormant> {
+    pub unsafe fn dormant(self) {
```
ithinuel, line 105: This consumes the `RingOscillator<_>` without returning anything.
Is this intended?

rp2040-hal/src/xosc.rs
jannic: Yes, such a return type would be exactly what I'd suggest if we want to support setting the non-active oscillator to dormant.
As it would require additional code to determine the actual mode, and complicates the return type, I hesitate to add it just for completeness. That's why I think about actual use cases.

andrewh42: Yes, I like the idea of keeping the implementation as ergonomic as possible for likely use cases.

rp2040-hal/src/rosc.rs
andrewh42: Oops, well spotted. I didn't have my rust brain fully switched on when removing the `Dormant` as you suggested earlier @ithinuel. :-} I'll change it to borrow as intended.

jannic:
> Hi @ithinuel and @jannic, is there anything I can do to help complete this PR?

Sorry for not answering sooner. Currently, my focus is on getting 0.10.0 out the door, so I neglected some older pull requests.

Currently, there are some merge conflicts. Could you try to resolve them?

andrewh42:
> > Hi @ithinuel and @jannic, is there anything I can do to help complete this PR?
> 
> Sorry for not answering sooner. Currently, my focus is on getting 0.10.0 out the door, so I neglected some older pull requests.
> 
> Currently, there are some merge conflicts. Could you try to resolve them?

No worries, done.

jannic (approved):
Thanks, Andrew! I think the pull request can be merged now. @ithinuel, any objections?

jannic:
BTW I also verified that the dormant mode actually makes a difference. Even with a cheap USB power meter, I see 0.02A in wfi mode, while in dormant mode, it shows 0.00A. The absolute numbers might be very inaccurate, but it's obvious that the dormant mode is different from the wfi sleep.

ithinuel (approved):

ithinuel:
I'm happy with this. We can always come back later to this if we need to make improvements.
I have a ppk2 at home so I could take some samples and post screenshots here.

PR #702, opened 2023-10-21
Improve `Spi::new` documentation

ithinuel:
cc @thejpster


thejpster:
Perhaps we could add an example, or add a link to an example?

thejpster:
See also #699.

jannic (approved):

PR #704, opened 2023-10-21
Document what's happening with derives on i2c::Error

ithinuel:


thejpster:
Thanks! That will help next time I stumble across this and wonder why the derive is cfg'd out.

thejpster (approved):

PR #706, opened 2023-10-24
Reconfigure Spi new functions for better type inference

fu5ha:
Does what it says on the tin :) unfortunately a breaking change but it means we can get rid of the `Spi::<_, _, _, 8>::` paradigm.

### DIFF
--- a/rp2040-hal/src/spi.rs
+++ b/rp2040-hal/src/spi.rs
@@ -22,7 +22,7 @@
 //! let spi_device = peripherals.SPI0;
 //! let spi_pin_layout = (mosi, sclk);
 //!
-//! let spi = Spi::<_, _, _, 8>::new(spi_device, spi_pin_layout)
+//! let spi = Spi::new(spi_device, spi_pin_layout)
 //!     .init(&mut peripherals.RESETS, 125_000_000u32.Hz(), 16_000_000u32.Hz(), MODE_0);
 //! ```
 
@@ -174,11 +174,12 @@ impl Sealed for u16 {}
 /// - `DS`: The "data size", i.e. the number of bits transferred per data frame. Defaults to 8.
 ///
 /// In most cases you won't have to specify these types manually and can let the compiler infer
-/// them for you based on the values you pass in to `new`. If you want to select a different
-/// data frame size, you'll need to do that by specifying the `DS` parameter manually.
+/// them for you based on the values you pass in to [`new`][Self::new], which assumes the default
+/// data frame size of 8 bits. If you want to select a  different data frame size, you'll need to
+/// use [`new_with_data_size`][Self::new_with_data_size].
 ///
 /// See [the module level docs][self] for an example.
-pub struct Spi<S: State, D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8 = 8u8> {
+pub struct Spi<S: State, D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> {
     device: D,
     pins: P,
     state: PhantomData<S>,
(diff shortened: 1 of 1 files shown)


rp2040-hal/src/spi.rs
```diff
@@ -251,17 +252,33 @@ impl<S: State, D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<S, D, P, DS
...
             pins,
             state: PhantomData,
         }
     }
 
+    /// Create new (not initialized) Spi bus with a non-default data size.
+    ///
+    /// Initialize it with [`.init`][Self::init]
+    /// or [`.init_slave`][Self::init_slave].
+    pub fn new_with_data_size<const DS: u8>(device: D, pins: P) -> Spi<Disabled, D, P, DS> {
```
jannic, line 272: I wonder if this method should be moved down to the `impl<D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<Disabled, D, P, DS>`? It's a little bit strange to have the constructor for the bit-generic version in the 8-bit impl.

fu5ha: Done. The only awkward thing is that now there is already `DS` const generic from the impl block in scope, so I had to rename it on the function level parameter. But not that big of a deal.

jannic: I don't see any update on the branch on github. Did you perhaps forget to push it?

Regarding the function level parameter: I think you don't need it at all, you can use the parameter from the impl block:
```
impl<D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<Disabled, D, P, DS> {
  [...]
  pub fn new_with_data_size(device: D, pins: P) -> Spi<Disabled, D, P, DS> {
    [...]
  }
  [...]
}
```

Which makes me wonder: Is the `new_with_data_size` name confusing, if it's often possible to call it without specifying the data size at the call location? It's enough if the compiler can somehow infer the correct value for DS, eg. if you later store the value in some struct where the type is specified.

rp2040-hal/src/spi.rs
```diff
@@ -251,15 +252,32 @@ impl<S: State, D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<S, D, P, DS
...
+    ///
+    /// Initialize it with [`.init`][Self::init]
     /// or [`.init_slave`][Self::init_slave].
     ///
     /// Valid pin sets are in the form of `(Tx, Sck)` or `(Tx, Rx, Sck)`
     ///
     /// If you pins are dynamically identified (`Pin<DynPinId, _, _>`) they will first need to pass
     /// validation using their corresponding [`ValidatedPinXX`](ValidatedPinTx).
-    pub fn new(device: D, pins: P) -> Spi<Disabled, D, P, DS> {
+    pub fn new(device: D, pins: P) -> Spi<Disabled, D, P, 8> {
```
ithinuel, line 265: It seems to me that choosing 8bits for the `fn new` a bit arbitrary.
I understand 8 bits were used as the default for the type before, but I don't think it makes it more correct 😛

fu5ha: To be fair I don't have that much embedded experience, but it seems like 8 bits as word size is extremely common in general and would at least not be particularly surprising

rp2040-hal/src/spi.rs
fu5ha: True, but I feel like storing this in a wider struct or somewhere it could be inferred would be relatively uncommon? Maybe I'm mistaken though.

fu5ha:
Maybe a way to solve both comments @ithinuel and @jannic would be to just make `fn new<const DS: usize>` so you always have to specify `DS` on the function (if it can't be inferred), but at least now you wouldn't have to `_` the other three type parameters. Thoughts?

ithinuel:
I need to try a few thing, to make sure I'm not missing some inference subtleties.

rp2040-hal/src/spi.rs
jannic: That was my assumption as well, but to be honest my experience with embedded peripherals is also quite limited.

jannic:
> EDIT: I made this: [ithinuel@90c6c82](https://github.com/ithinuel/rp-hal/commit/90c6c820090c796a3e6551c03acfcc5f10f9ad34) I only managed to get rid of the turbofish by introducing a sorta builder type.
> 
> Let me know your thoughts.

I'm not convinced.
This is all about discoverability and ease of writing code. The current incantation, `let spi = hal::spi::Spi::<_, _, _, 8>::new(pac.SPI0, (spi_mosi, spi_miso, spi_sclk));`, isn't unreadable. It's just difficult to get right if you have to write it.

This is not really getting better with `let spi = hal::spi::SpiBuilder::build::<8>(pac.SPI0, (spi_mosi, spi_miso, spi_sclk));`.
In some cases (like in the on target tests) it even gets worse, as previously the parameters were inferred from the `struct State` definition, so the actual code was just `let spi = hal::spi::Spi::new(pac.SPI0, (spi_mosi, spi_miso, spi_sclk));`

I wonder if we could remove the `DS` parameter entirely? Sure, we'd lose a little bit of type safety. But not much, I don't think this prevents actual errors. However, removing it would prevent us from easily implementing the DMA traits with the correct word width.

rp2040-hal/src/spi.rs
jannic: I had a look at embassy-rp to see how they handle the word width. They don't. It's hard coded to 8 bits. :smile: 
https://github.com/embassy-rs/embassy/blob/main/embassy-rp/src/spi.rs#L87

jannic:
Another option could be to make DS a conventional generic parameter, with different types for the available  bit sized. This would have the additional advantage that it would become impossible to set invalid bit sizes. The current API would allow to configure a device with a bit size of 0 or more than 16 bits. (That device would be unusable though, as the read and write methods are only defined for valid bit sizes.)

ithinuel:
I'm not too attached to that parameter TBH. On a setup with multiple spi peripheral with the rp2040 as one controller on the bus, you may need to change the word size between different peripheral and this generic parameter does not allow for an easy implementation of that usecase.

fu5ha:
I will close this for now, it may still be relevant but not so much work to revive or redo if we want it...

PR #708, opened 2023-10-25
Ensure that i2c pins have PullUp activated

### LINKED ISSUE #690: Infinite loop in i2c example
In my environment, I2C example does not reach to [loop](https://github.com/rp-rs/rp-hal/blob/78f71ae2f12d67af07c79277461ec6e7606e3267/rp2040-hal/examples/i2c.rs#L99) .
It stucks at
https://github.com/rp-rs/rp-hal/blob/0f5bc072cdb6d51bf27d6fd236fe2faf894124f9/rp2040-hal/src/i2c/controller.rs#L219
that causes infinite loop.
I simply did `git clone` and `cargo run --example i2c --features="critical-section-impl"`
Am I missing some hidden settings?

I'm working on Windows10, and using Raspberry Pi Pico.

jannic:
It is possible to skip this check by using `hal::I2C::i2c0_unchecked(...)` or `hal::I2C::i2c1_unchecked(...)`, but the default constructor functions now only accept pins with activated PullUp.

Fixes #690.

### DIFF
--- a/rp2040-hal/src/i2c.rs
+++ b/rp2040-hal/src/i2c.rs
@@ -340,8 +340,10 @@ macro_rules! hal {
 
                 $crate::paste::paste! {
                     /// Configures the I2C peripheral to work in master mode
-                    /// This function can be called without the pull-ups on the I2C pins.
-                    pub fn [<$i2cX _unchecked>]<F, SystemF>(
+                    ///
+                    /// This function can be called without activating internal pull-ups on the I2C pins.
+                    /// It should only be used if external pull-ups are provided.
+                    pub fn [<$i2cX _with_external_pull_up>]<F, SystemF>(
                         i2c: $I2CX,
                         sda_pin: Sda,
                         scl_pin: Scl,
--- a/rp2040-hal/src/i2c.rs
+++ b/rp2040-hal/src/i2c.rs
@@ -47,7 +47,7 @@ use core::{marker::PhantomData, ops::Deref};
 use fugit::HertzU32;
 
 use crate::{
-    gpio::{bank0::*, pin::pin_sealed::TypeLevelPinId, AnyPin, FunctionI2c},
+    gpio::{bank0::*, pin::pin_sealed::TypeLevelPinId, AnyPin, FunctionI2c, PullUp},
     pac::{self, i2c0::RegisterBlock as I2CBlock, I2C0, I2C1, RESETS},
     resets::SubsystemReset,
     typelevel::Sealed,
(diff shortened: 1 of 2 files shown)


rp2040-hal/src/i2c.rs
```diff
@@ -327,12 +327,32 @@ macro_rules! hal {
...
+                    Scl: ValidPinScl<$I2CX> + AnyPin<Pull = PullUp>,
                     SystemF: Into<HertzU32>,
                 {
                     Self::new_controller(i2c, sda_pin, scl_pin, freq.into(), resets, system_clock.into())
                 }
+
+                $crate::paste::paste! {
+                    /// Configures the I2C peripheral to work in master mode
+                    /// This function can be called without the pull-ups on the I2C pins.
+                    pub fn [<$i2cX _unchecked>]<F, SystemF>(
```
jannic, line 340: I'm not convinced by the `_unchecked` suffix, as it's otherwise often used to skip some runtime check. Any better ideas?

jannic:
The added check immediately uncovered an issue in a doc comment, where the I2C pins were not configured with pull-ups: https://github.com/rp-rs/rp-hal/actions/runs/6644179376/job/18052788902

rp2040-hal/src/i2c.rs
ithinuel: How about a `new_with_external_pull_up` ?

jannic: That wouldn't fit. The difference is just that this function doesn't check the pull type. It could still use internal pull-ups.
Perhaps `new_unchecked_pull_mode`?
IMHO for this method it doesn't matter if the name is long and ugly. I don't expect it to be used very often. It's just for completeness, in case someone wants to do something strange.

Sympatron: I like `new_with_external_pull_up` a lot better. The meaning of "unchecked" in this context is not very clear. Having external pull-ups is quite common in general and might be desirable to tailor your design to your EMI and speed constraints.

jannic: Ok, changed the function name. I still omitted the `new` part, because it would be inconsistent. Now the two functions available are:
```
    pub fn i2c0<F, SystemF>(
        i2c: I2C0,
        sda_pin: Sda,
        scl_pin: Scl,
        freq: F,
        resets: &mut RESETS,
        system_clock: SystemF,
    ) -> Self
    where
        F: Into<HertzU32>,
        Sda: ValidPinSda<I2C0> + AnyPin<Pull = PullUp>,
        Scl: ValidPinScl<I2C0> + AnyPin<Pull = PullUp>,
        SystemF: Into<HertzU32>,
```
```
    pub fn i2c0_with_external_pull_up<F, SystemF>(
        i2c: I2C0,
        sda_pin: Sda,
        scl_pin: Scl,
        freq: F,
        resets: &mut RESETS,
        system_clock: SystemF,
    ) -> Self
    where
        F: Into<HertzU32>,
        Sda: ValidPinSda<I2C0>,
        Scl: ValidPinScl<I2C0>,
        SystemF: Into<HertzU32>,
```

ithinuel (approved):

PR #711, opened 2023-11-01
Add uart_is_busy() method for UART devices

Rahix:
The `uart_is_busy()` method returns `true` as long as the UART is busy transmitting data.  It will return `false` once all bits (including stop bits) have been transmitted.


9names (approved):
Lgtm, thanks!

PR #712, opened 2023-11-02
Properly report break conditions

jannic:
A break condition is also a framing error. As it is the more specific code, return the break error.

### DIFF
--- a/rp2040-hal/src/uart/reader.rs
+++ b/rp2040-hal/src/uart/reader.rs
@@ -146,7 +146,7 @@ pub(crate) fn read_raw<'b, D: UartDevice>(
             }
 
             // A break condition is also a framing error.
-            // As it is the more specifc code, return
+            // As it is the more specific code, return
             // the break error.
             if read.be().bit_is_set() {
                 error = Some(ReadErrorType::Break);
--- a/rp2040-hal/src/uart/reader.rs
+++ b/rp2040-hal/src/uart/reader.rs
@@ -133,23 +133,18 @@ pub(crate) fn read_raw<'b, D: UartDevice>(
 
             let read = device.uartdr.read();
 
+            // If multiple status bits are set, report
+            // the most serious or most specific condition,
+            // in the following order of precedence:
+            // overrun > break > parity > framing
             if read.oe().bit_is_set() {
                 error = Some(ReadErrorType::Overrun);
-            }
-
-            if read.pe().bit_is_set() {
-                error = Some(ReadErrorType::Parity);
-            }
-
-            if read.fe().bit_is_set() {
-                error = Some(ReadErrorType::Framing);
-            }
-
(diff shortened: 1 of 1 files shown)


jannic:
I wonder what error should be returned if both FE and PE are set. Is one of them more specific and should be preferred?

jannic:
embassy-rp uses this order of precedence, which looks sensible:
```
            if dr.oe() {
                Err(nb::Error::Other(Error::Overrun))
            } else if dr.be() {
                Err(nb::Error::Other(Error::Break))
            } else if dr.pe() {
                Err(nb::Error::Other(Error::Parity))
            } else if dr.fe() {
                Err(nb::Error::Other(Error::Framing))
            } else {
                Ok(dr.data())
            }
```

9names:
> I wonder what error should be returned if both FE and PE are set. Is one of them more specific and should be preferred?

FWIW I would have frame error as a higher priority.
Sure, your data is corrupt either way but a parity error only tells you that an odd number of bits are wrong, a framing error tells you *why* the sum of bits is wrong.

jannic:
I don't agree that the framing error tells you _why_ the parity was wrong. Both flags can be set because of a single bit-error. (If one, or an odd number of, data/parity bits are wrong, you get a parity error. If the stop bit is wrong, you get a framing error.)
But I agree that the framing error should probably have priority: It's more specific. You exactly know _which_ bit was wrong. So it tells you _slightly_ more than the parity error.
Now the difference between the two choices is so small that consistency with other implementations seems to be more important to me. I'll check what the C SDK does.

rp2040-hal/src/uart/reader.rs
```diff
@@ -133,19 +133,17 @@ pub(crate) fn read_raw<'b, D: UartDevice>(
...
                 error = Some(ReadErrorType::Break);
-            }
-
-            if read.pe().bit_is_set() {
+            } else if read.pe().bit_is_set() {
                 error = Some(ReadErrorType::Parity);
-            }
-
-            if read.fe().bit_is_set() {
+            } else if read.fe().bit_is_set() {
```
ithinuel, line 146: Is the parity bit relevant if there's a framing error?

jannic: I didn't find error handling code in the C sdk. If both FE and PE are set, embassy reports the parity error, most (but not all) linux serial drivers I checked as well. So there doesn't seem to be a consensus on how to handle that case. And honestly, I think it doesn't matter much.

If there are random bit errors, PE will happen more often than FE. If parity is misconfigured, there will only PE errors (and each character will trigger it). If no parity bit is configured, PE is irrelevant. In general, both errors indicate that the character is probably damaged, so error handling for both is probably identical anyway.

ithinuel (approved):

PR #713, opened 2023-11-04
Fix UART transmit_flushed method

jannic:
The UART is not completely flushed when the TX FIFO becomes empty, as long as there are still bits left in the TX shift register.

Therefore, transmit_flushed() should check the busy bit, not the txfe bit.


9names (approved):

ithinuel (commented):
I think this is an appropriate solution.
Do we still need `uart_is_busy` with this ?

jannic:
> I think this is an appropriate solution. Do we still need `uart_is_busy` with this ?

I considered removing it, but decided to keep it:
- it doesn't really hurt (neither in code size, nor by cluttering the docs)
- it is nicer and more obvious to write `uart_is_busy()` than `transmit_flushed().is_err()` if you want to check if the uart is busy.

PR #718, opened 2023-11-12
Add a short explanation on how to use RingOscillator as a random number source

jannic:


rp2040-hal/src/rosc.rs
```diff
@@ -1,5 +1,16 @@
 //! Ring Oscillator (ROSC)
-// See [Chapter 2 Section 17](https://datasheets.raspberrypi.org/rp2040/rp2040_datasheet.pdf) for more details
+//! See [Chapter 2 Section 17](https://datasheets.raspberrypi.org/rp2040/rp2040_datasheet.pdf) for more details
+//!
```
9names, line 3: Add a blank line here so the "See chapter" comment doesn't become part of the heading (which is copied into the parent document)
```suggestion
//!
//! See [Chapter 2 Section 17](https://datasheets.raspberrypi.org/rp2040/rp2040_datasheet.pdf) for more details
//!
```

rp2040-hal/src/rosc.rs
```diff
@@ -1,5 +1,16 @@
...
+//! In addition to its obvious role as a clock source, [`RingOscillator`] can also be used as a random number source
+//! for the [`rand`] crate:
+//!
+//! ```no_run
+//! # let mut pac = rp2040_pac::Peripherals::take().unwrap();
+//! use rp2040_hal::rosc::RingOscillator;
+//! use rand::Rng;
+//! let mut rnd = RingOscillator::new(pac.ROSC).initialize();
+//! let random_value: u32 = rnd.gen();
+//! ```
```
9names, line 14: Add link to rand so the link on line 5 goes somewhere
```suggestion
//! ```
//! [`rand`]: https://docs.rs/rand
```

9names (changes requested):

9names (approved):

PR #723, opened 2023-11-19
Make PIO IRQ IDs into ZSTs

9ary:
This allows getting the IRQ with `pio.irq::<IRQn>()`, where IRQn is a type generic.

This is a minor breaking change. I struggled for a while trying to get something similar working with the existing const generic, without success.

### DIFF
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -123,20 +123,29 @@ impl<P: PIOExt> PIO<P> {
         self.pio
     }
 
-    /// This PIO's IRQ0 interrupt.
-    pub fn irq0(&self) -> Interrupt<'_, P, 0> {
+    /// This PIO's interrupt by index.
+    pub fn irq<const IRQ: usize>(&self) -> Interrupt<'_, P, IRQ> {
+        struct IRQSanity<const X: usize>;
+        impl<const IRQ: usize> IRQSanity<IRQ> {
+            const CHECK: () = assert!(IRQ <= 1, "IRQ index must be either 0 or 1");
+        }
+
+        #[allow(clippy::let_unit_value)]
+        let _ = IRQSanity::<IRQ>::CHECK;
         Interrupt {
             block: self.pio.deref(),
             _phantom: core::marker::PhantomData,
         }
     }
 
+    /// This PIO's IRQ0 interrupt.
+    pub fn irq0(&self) -> Interrupt<'_, P, 0> {
+        self.irq()
+    }
+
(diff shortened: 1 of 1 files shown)


ithinuel:
Hello, thank you for the contribution.
Could you elaborate on your usecase/motivation for this change?

9ary:
This allows me to write a module which uses PIO while being generic over the IRQ it uses. See https://github.com/redolution/cubane/blob/e35d1be482b2e81d074fc699efcec44f840ab4f3/src/exi.rs, in particular these two lines:
- https://github.com/redolution/cubane/blob/e35d1be482b2e81d074fc699efcec44f840ab4f3/src/exi.rs#L166
- https://github.com/redolution/cubane/blob/e35d1be482b2e81d074fc699efcec44f840ab4f3/src/exi.rs#L180

I can't use a match block over the const generic here because the two IRQs have different type signatures, type-level selection is the only option as far as I can tell.
I've tried several approaches to make this work while retaining the const generic, but to no avail. I'm not sure it's actually possible without the ability to construct the object myself, so I fixed it at the root.

ithinuel:
I see, thank you :) 

Only adding `fn irq<const IRQ: usize>(&self) -> Interrupt<'_, P, IRQ>` wouldn't be good as it'd allow to create an `Interrupt` with any value of `IRQ` where only 0 and 1 are valid.

9ary:
Yup, hence I implemented the same construct as is already used for things like state machine and DMA channel IDs.

ithinuel (approved):

ithinuel:
Actually I found a way to check for that that's transparent to the user:

```rust
pub fn irq<const IRQ: usize>(&self) -> Interrupt<'_, P, IRQ> {                                                                    
    struct IRQSanity<const X: usize>;                                                                                             
    impl<const IRQ: usize> IRQSanity<IRQ> {                                                                                       
        const CHECK: () = assert!(IRQ == 0 || IRQ == 1, "Irq Index must be either 0 or 1");                                       
    }                                                                                                                             

    let _test = IRQSanity::<IRQ>::CHECK;                                                                                          
    Interrupt {                                                                                                                   
        block: self.pio.deref(),                                                                                                  
        _phantom: core::marker::PhantomData,                                                                                      
    }                                                                                                                             
}
```

9ary:
Oh that's neat, although the error message is a bit weird.
Fwiw since it's a usize, `IRQ <= 1` should be enough as the condition.

9ary:
Updated with your suggestion, so this is no longer a breaking change.

ithinuel (approved):

Issue #724, opened 2023-12-02
ClockManager "partial move"


jnthbdn:
Hello,

I'm using the main version to put the rp2040 in deepsleep. All in all it works great, but I have a small problem.

When I create a `RealTimeClock` object I have to give it ownership over the `RtcClock` in my `ClockManager`. From then on, I can no longer pass my `ClockManager` as a parameter (`&mut`) to any function... Because the borrow checker tells me that the object is `partial move`.

How can I get round this problem?

I used the example https://github.com/rp-rs/rp-hal/blob/cfd6c128a1688186f419e3bd60ce469ddeb7fb47/rp2040-hal/examples/rtc_irq_example.rs.

Here is a simplified code example.

```rust
let mut pac = pac::Peripherals::take().unwrap();
let mut clocks = hal::clocks::ClocksManager::new(pac.CLOCKS);
// [...]
let rtc = configure_rtc(pac.RTC, clocks.rtc_clock, &mut pac.RESETS, &mut core);
// [...]

loop {
    set_next_wake_up(600);
    configure_clock_low_power(&mut clocks, &xosc, XTAL_FREQ_HZ, &mut plls);
    cortex_m::asm::wfi();
    configure_clock(&mut clocks, &xosc, XTAL_FREQ_HZ, &mut plls, &mut pac.RESETS);
}

// In configure_rtc
let mut rtc = hal::rtc::RealTimeClock::new(
    pac_rtc,
    rtc_clock,
    reset,
    rtc::DateTime {
        year: 0,
        month: 1,
        day: 1,
        day_of_week: rtc::DayOfWeek::Monday,
        hour: 0,
        minute: 0,
        second: 0,
    },
)
.unwrap();
// [...]
```

I did try to make my own `ClockManager`, but I find it a bit cumbersome and there are a lot of macros!

Can you help me ? Maybe there is something I don't understand with Rust ^^

Thanks !

jnthbdn:
I just found something very strange... Why does the `new` function of `RealTimeClock` need the ownership of rtc_clock, since in the code we're just reading the frequency (where a simple reference would be sufficient)?

https://github.com/rp-rs/rp-hal/blob/cfd6c128a1688186f419e3bd60ce469ddeb7fb47/rp2040-hal/src/rtc/mod.rs#L68

ithinuel:
Hi, 

[...]

PR #725, opened 2023-12-03
Fix RealTimeClock & UsbBus ownership

jnthbdn:
The RealTimeClock works great, but it takes ownership of the `rtc_clock` field in the `ClockManager`, which makes it invalid (partial move).
Moreover, the `RealTimeClock` struct doesn't allow freeing `rtc_clock`, the `new` function doesn't do anything about it, so the instance is destroyed.

Following the model of other functions present in the HAL for clock configuration (such as `setup_pll_blocking` or `configure_clock` from ClockManger), I modified the parameter so that it was a reference.

#724

Thanks for this brilliant HAL!

### DIFF
--- a/rp2040-hal/src/rtc/mod.rs
+++ b/rp2040-hal/src/rtc/mod.rs
@@ -36,6 +36,7 @@ pub use self::datetime::{DateTime, DayOfWeek, Error as DateTimeError};
 /// A reference to the real time clock of the system
 pub struct RealTimeClock {
     rtc: RTC,
+    clock: RtcClock,
 }
 
 impl RealTimeClock {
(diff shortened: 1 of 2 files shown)


jnthbdn:
Thanks for your review, I quite agree with your comment, it would make the function's intentions clearer.

But the more I look at the code, the more I wonder if my RP is really relevant...

Let me explain, as you said in issue #724, it's interesting that ownership is taken in order to ensure that clocks are not subsequently modified. But there's still the problem of the ClockManager being partially moved. So in the end, even though it's a big change, wouldn't it be better to make the ClockManager fields `Option<T>`?
Changing this line
https://github.com/rp-rs/rp-hal/blob/cfd6c128a1688186f419e3bd60ce469ddeb7fb47/rp2040-hal/src/clocks/macros.rs#L22 with `pub [<$name:snake>]: Option<$name>`

Because in this PR we've solved the RealTimeClock problem, but the `new` function in the `UsbBus` does the same thing.
https://github.com/rp-rs/rp-hal/blob/cfd6c128a1688186f419e3bd60ce469ddeb7fb47/rp2040-hal/src/usb.rs#L441

ithinuel:
Would it make sense to rather than having fields public, adding a `split` method and include as part of its output a extra entry offering a subset of what the clock manager does but that does not collide with the individual clocks ?

jnthbdn:
I modified the PR by adding `free` functions to the `RealTimeClock` and `UsbBus` structs.

ithinuel (approved):
LGTM

PR #728, opened 2023-12-17
Solve the issue of not being able to override existing builtin functions under LTO.

DianQK:
Solve the https://github.com/rp-rs/rp2040-project-template/issues/71 issue of not recognizing builtin functions under LTO.
Also set the builtin functions in compiler_builtins to weak.

### DIFF
--- a/rp2040-hal/src/lib.rs
+++ b/rp2040-hal/src/lib.rs
@@ -94,6 +94,7 @@ pub use sio::Sio;
 pub use spi::Spi;
 pub use timer::Timer;
 pub use watchdog::Watchdog;
+extern crate compiler_builtins;
 // Re-export crates used in rp2040-hal's public API
 pub extern crate fugit;
 
--- a/rp2040-hal/src/lib.rs
+++ b/rp2040-hal/src/lib.rs
@@ -37,6 +37,7 @@
 
 #![warn(missing_docs)]
 #![no_std]
+#![feature(naked_functions)]
 
 #[doc(hidden)]
 pub use paste;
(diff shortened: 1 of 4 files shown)


rp2040-hal/src/lib.rs
```diff
@@ -37,6 +37,7 @@
 
 #![warn(missing_docs)]
 #![no_std]
+#![feature(naked_functions)]
```
ithinuel, line 40: Is my understanding correct that this is not stable yet ?

DianQK: Yeah, it's only on nightly. Are we going to use it on stable too?

jannic: This library targets stable rust, yes. So nightly features can't be used.

DianQK: It seems that even stabilizing naked is not enough. Using compiler_builtins requires nightly.
But we can change it to default weak in rust. I think the key here is still to stabilize the naked first.

DianQK: Could you please help me verify that the current modification works?

jannic: Sure, what can I do?
Currently, it doesn't build. Obviously not with stable, but I also get errors with nightly:
```
$ cargo +nightly build --release
[...]
   Compiling rp2040-hal v0.9.1 (/home/jan/rp2040/rp-rs/rp-hal/rp2040-hal)
error: invalid instruction
   --> rp2040-hal/src/sio.rs:299:22
    |
299 |                     "hwdivider_head",
    |                      ^
    |
note: instantiated into assembly here
   --> <inline asm>:1:2
    |
1   |     hwdivider_head
    |     ^
[...]
```

DianQK: I'm not sure about this error yet. But I was able to compile it successfully at https://github.com/rp-rs/rp2040-project-template. I also need add `compiler_builtins = { version = "0.1.104", features = ["weak-intrinsics"] }` to the rp2040-project-template.

DianQK:
I believe there are ways to not use `naked` here.

PR #729, opened 2023-12-17
Add SM as an associated type of ValidStateMachine in pio

Sympatron:
Since some functions in the `pio` module expect a `ValidStateMachine` and some require `(PIOExt, StateMachineIndex)` it's difficult to write generic code without access to `ValidStateMachine::SM`. This PR fixes this.

### DIFF
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -425,6 +425,8 @@ impl Sealed for SM3 {}
 pub trait ValidStateMachine: Sealed {
     /// The PIO block to which this state machine belongs.
     type PIO: PIOExt;
+    /// The index type of this state machine
+    type SM: StateMachineIndex;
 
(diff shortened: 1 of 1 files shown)


jannic:
Can you give a quick example where this is useful?
(Not a full example to be added to the repo, just a quick code snipped showing the use case.)

rp2040-hal/src/pio.rs
```diff
@@ -453,6 +455,7 @@ pub type PIO1SM3 = (PIO1, SM3);
 
 impl<P: PIOExt, SM: StateMachineIndex> ValidStateMachine for (P, SM) {
     type PIO = P;
+    type SM = SM;
```
jannic, line 458: Perhaps call the type something like `SMIdx`? This `type SM = SM` looks a little bit strange, and `SM` isn't really referring to a specific state machine, only the full tuple `(P, SM)` does.

PR #730, opened 2023-12-18
Replace asm macros by rust macros

jannic:
According to https://doc.rust-lang.org/stable/reference/inline-assembly.html#directives-support the assembly directive `.macro` is not part of the supported subset.

The reference further explains: "The result of using other directives is assembler-specific (and may cause an error, or may be accepted as-is)"

So while it obviously works right now, it might cause issues in future versions of rust. Not sure if this would be avoided as a breaking change. To be on the safe side, replace asm macros by rust macros.

### DIFF
--- a/rp2040-hal/src/sio.rs
+++ b/rp2040-hal/src/sio.rs
@@ -202,79 +202,97 @@ impl SioFifo {
     }
 }
 
+macro_rules! concatln {
+    ($(,)*) => {
+        ""
+    };
+    ( $e:expr ) => {
+        $e
+    };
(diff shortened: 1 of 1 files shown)


thejpster:
Have we checked that the asm generated is the same / equivalent? And/or have we run some tests on hardware?

jannic:
> Have we checked that the asm generated is the same / equivalent? And/or have we run some tests on hardware?

Unfortunately the generated binaries are different. But the changes are not in the division functions. It's things like changed register allocations all over the place. I don't know why rustc acts that way.

The few binaries I tried on real hardware do work.

thejpster (approved):

PR #734, opened 2023-12-19
Upgrade embedded-hal 1.0 dependency to 1.0.0-rc.3

jannic:


rp2040-hal/src/timer.rs
```diff
@@ -167,7 +167,17 @@ macro_rules! impl_delay_traits {
...
+        // For now, just use microsecond delay, internally. Of course, this
+        // might cause a much longer delay than necessary. So a more advanced
+        // implementation would be desirable for sub-microsecond delays.
+        let us = ns / 1000 + if ns % 1000 == 0 { 0 } else { 1 };
+        // With rustc 1.73, this can be replaced by:
+        // let us = ns.div_ceil(1000);
+        self.delay_us_internal(us)
+    }
+
     fn delay_us(&mut self, us: u32) {
```
thejpster, line 181: We might want to implement `delay_ms` here too, as we can do better than the default (which calls `delay_ns`)? Although arguably the inefficiency of the division is irrelevant, given the user wants to stop the CPU from doing anything useful for many 100'000s of clock cycles.

thejpster (approved):

rp2040-hal/src/timer.rs
jannic: Still a good point, as we already have a manual implementation of `delay_us()` for the embedded-hal 0.2 trait. We should do the same here.

PR #736, opened 2023-12-19
Implement embedded_hal::pwm::SetDutyCycle of embedded-hal 1.0

jannic:
Also add a new example pwm_blink_embedded_hal_1 which uses this trait, to have a minimal test case.

### DIFF
--- a/rp2040-hal/examples/pwm_blink_embedded_hal_1.rs
+++ b/rp2040-hal/examples/pwm_blink_embedded_hal_1.rs
@@ -1,0 +1,122 @@
+//! # PWM Blink Example
+//!
+//! If you have an LED connected to pin 25, it will fade the LED using the PWM
+//! peripheral.
+//!
+//! It may need to be adapted to your particular board layout and/or pin assignment.
+//!
+//! See the `Cargo.toml` file for Copyright and license details.
+
+#![no_std]
+#![no_main]
+
+// Ensure we halt the program on panic (if we don't mention this crate it won't
+// be linked)
+use panic_halt as _;
+
+// Alias for our HAL crate
+use rp2040_hal as hal;
+
+// Some traits we need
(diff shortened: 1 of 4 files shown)


rp2040-hal/examples/pwm_blink_embedded_hal_1.rs
```diff
@@ -0,0 +1,122 @@
...
+    // Infinite loop, fading LED up and down
+    loop {
+        // Ramp brightness up
+        for i in (LOW..=HIGH).skip(100) {
+            delay.delay_us(8);
+            channel.set_duty_cycle(i).unwrap();
+        }
+
+        // Ramp brightness down
+        for i in (LOW..=HIGH).rev().skip(100) {
```
ithinuel, line 113: the first loop goes from `(LOW+100)` to `HIGH` but the second goes to `(HIGH-100)` to `LOW` so, to have both run on the same (but reversed) range, shouldn't this be skip then rev instead ?

```suggestion
        for i in (LOW..=HIGH).skip(100).rev() {
```

jannic: To be honest, I barely looked at that code, I just copied `pwm_blink.rs` and modified it to have at least one user of the new added API.

With your suggested change, the LED never turns off completely. And that way, the code doesn't make much sense: If it should count from 100 to HIGH and back, why not just set `LOW = 100` and omit the `.skip(100)`?

@thejpster: According to `git blame`, you added that code. Do you remember why you added the `skip(100)` calls? If there's no reason to keep them, I'd prefer to remove them.

jannic: Ok, I removed them :-)

thejpster (approved):

PR #738, opened 2023-12-21
add missing ValidFunction implementation for DynFunction.

ithinuel:
This missing impl effectively prevented anyone from using a `Pin<_, DynFunction, _>`.


thejpster:
This now works:

```rust
let mut uart0_tx = pins
        .gpio0
        .reconfigure::<gpio::DynFunction, gpio::DynPullType>()
        .into_dyn_pin();
```

A helper method like `into_really_dynamic_pin()` or `erase_all_the_type_state()` would be useful :)

PR #747, opened 2024-01-01
[i2c] (re)introduce async.await implementation

ithinuel:
This re-introduces the async.await i2c driver that was once remove because of requiring nightly at the time.

Now that the necessary features have been stabilized, we can all enjoy a bit of async.await-ness :)

# Requires
- #761
- #762
- #763
- #764

### DIFF
--- a/on-target-tests/tests/i2c_loopback.rs
+++ b/on-target-tests/tests/i2c_loopback.rs
@@ -1,0 +1,128 @@
+//! This test needs a connection between:
+//!
+//! | from GPIO (pico Pin) | to GPIO (pico Pin) |
+//! | -------------------- | ------------------ |
+//! |         0 (1)        |       2 (4)        |
+//! |         1 (2)        |       3 (5)        |
+
+#![no_std]
+#![no_main]
+#![cfg(test)]
+
+use defmt_rtt as _; // defmt transport
+use defmt_test as _;
+use panic_probe as _;
+use rp2040_hal as hal; // memory layout // panic handler
+
+use hal::pac::interrupt;
+
+/// The linker will place this boot block at the start of our program image. We
+/// need this to help the ROM bootloader get our code up and running.
+/// Note: This boot block is not necessary when using a rp-hal based BSP
+/// as the BSPs already perform this step.
+#[link_section = ".boot2"]
+#[used]
+pub static BOOT2: [u8; 256] = rp2040_boot2::BOOT_LOADER_GENERIC_03H;
+
+/// External high-speed crystal on the Raspberry Pi Pico board is 12 MHz. Adjust
+/// if your board has a different frequency
+const XTAL_FREQ_HZ: u32 = 12_000_000u32;
+
+pub mod i2c_tests;
+
+#[interrupt]
+unsafe fn I2C1_IRQ() {
+    i2c_tests::blocking::peripheral_handler();
+}
+
+#[defmt_test::tests]
+mod tests {
+    use crate::i2c_tests::{self, blocking::State, ADDR_10BIT, ADDR_7BIT};
+
+    #[init]
+    fn setup() -> State {
+        i2c_tests::blocking::setup(super::XTAL_FREQ_HZ, ADDR_7BIT)
+    }
+
+    #[test]
+    fn write(state: &mut State) {
+        i2c_tests::blocking::write(state, ADDR_7BIT);
+        i2c_tests::blocking::write(state, ADDR_10BIT);
+    }
+
+    #[test]
+    fn write_iter(state: &mut State) {
+        i2c_tests::blocking::write_iter(state, ADDR_7BIT);
+        i2c_tests::blocking::write_iter(state, ADDR_10BIT);
+    }
+
+    #[test]
+    fn write_iter_read(state: &mut State) {
+        i2c_tests::blocking::write_iter_read(state, ADDR_7BIT, 1..=1);
+        i2c_tests::blocking::write_iter_read(state, ADDR_10BIT, 2..=2);
+    }
+
+    #[test]
+    fn write_read(state: &mut State) {
+        i2c_tests::blocking::write_read(state, ADDR_7BIT, 1..=1);
+        i2c_tests::blocking::write_read(state, ADDR_10BIT, 2..=2);
+    }
+
+    #[test]
+    fn read(state: &mut State) {
+        i2c_tests::blocking::read(state, ADDR_7BIT, 0..=0);
+        i2c_tests::blocking::read(state, ADDR_10BIT, 1..=1);
+    }
+
+    #[test]
+    fn transactions_read(state: &mut State) {
+        i2c_tests::blocking::transactions_read(state, ADDR_7BIT, 0..=0);
+        i2c_tests::blocking::transactions_read(state, ADDR_10BIT, 1..=1);
+    }
+
+    #[test]
+    fn transactions_write(state: &mut State) {
+        i2c_tests::blocking::transactions_write(state, ADDR_7BIT);
+        i2c_tests::blocking::transactions_write(state, ADDR_10BIT);
+    }
+
+    #[test]
+    fn transactions_read_write(state: &mut State) {
+        i2c_tests::blocking::transactions_read_write(state, ADDR_7BIT, 1..=1);
+        i2c_tests::blocking::transactions_read_write(state, ADDR_10BIT, 2..=2);
+    }
+
+    #[test]
+    fn transactions_write_read(state: &mut State) {
+        i2c_tests::blocking::transactions_write_read(state, ADDR_7BIT, 1..=1);
+        i2c_tests::blocking::transactions_write_read(state, ADDR_10BIT, 2..=2);
+    }
+
+    #[test]
+    fn transaction(state: &mut State) {
+        i2c_tests::blocking::transaction(state, ADDR_7BIT, 7..=9);
+        i2c_tests::blocking::transaction(state, ADDR_10BIT, 7..=9);
+    }
+
+    #[test]
+    fn transactions_iter(state: &mut State) {
+        i2c_tests::blocking::transactions_iter(state, ADDR_7BIT, 1..=1);
+        i2c_tests::blocking::transactions_iter(state, ADDR_10BIT, 2..=2);
+    }
+
+    #[test]
+    fn embedded_hal(state: &mut State) {
+        i2c_tests::blocking::embedded_hal(state, ADDR_7BIT, 2..=2);
+        i2c_tests::blocking::embedded_hal(state, ADDR_10BIT, 2..=7);
+    }
+
+    // Sad paths:
+    // invalid tx buf on write
+    // invalid rx buf on read
+    //
+    // invalid (rx/tx) buf in transactions
+    //
+    // Peripheral Nack
+    //
+    // Arbritration conflict
+}
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -17,9 +17,12 @@ targets = ["thumbv6m-none-eabi"]
 
 [dependencies]
 cortex-m = "0.7.2"
-embedded_hal_0_2 = { package = "embedded-hal", version = "0.2.5", features = ["unproven"] }
+embedded_hal_0_2 = { package = "embedded-hal", version = "0.2.5", features = [
+  "unproven",
+] }
 embedded-hal = "1.0.0"
 embedded-hal-nb = "1.0.0"
+embedded-hal-async = "1.0.0"
 embedded-dma = "0.2.0"
 embedded-io = "0.6.1"
 fugit = "0.3.6"
@@ -54,6 +57,13 @@ hd44780-driver = "0.4.0"
 pio-proc = "0.2.0"
 dht-sensor = "0.2.1"
 rand = { version = "0.8.5", default-features = false }
+nostd_async = { version = "0.6.1", features = ["cortex_m"] }
+futures = { version = "0.3.30", default-features = false, features = [
+  "async-await",
+] }
+defmt-rtt = "0.4.0"
+panic-probe = { version = "0.3.1", features = ["print-defmt"] }
+defmt = "0.3"
 
 [features]
 # Minimal startup / runtime for Cortex-M microcontrollers
@@ -146,6 +156,14 @@ required-features = ["critical-section-impl"]
 [[example]]
 name = "i2c"
 required-features = ["critical-section-impl"]
+
+[[example]]
+name = "i2c_async"
+required-features = ["critical-section-impl", "rt"]
+
+[[example]]
+name = "i2c_async_cancelled"
+required-features = ["critical-section-impl", "rt", "defmt"]
 
 [[example]]
 name = "lcd_display"
--- a/rp2040-hal/examples/i2c.rs
+++ b/rp2040-hal/examples/i2c.rs
@@ -94,7 +94,7 @@ fn main() -> ! {
     );
 
     // Write three bytes to the I²C device with 7-bit address 0x2C
-    i2c.write(0x2c, &[1, 2, 3]).unwrap();
+    i2c.write(0x2Cu8, &[1, 2, 3]).unwrap();
 
     // Demo finish - just loop until reset
 
--- a/rp2040-hal/examples/i2c_async.rs
+++ b/rp2040-hal/examples/i2c_async.rs
@@ -1,0 +1,126 @@
+//! # I²C Example
+//!
+//! This application demonstrates how to talk to I²C devices with an RP2040.
+//! in an Async environment.
+//!
+//! It may need to be adapted to your particular board layout and/or pin assignment.
+//!
+//! See the `Cargo.toml` file for Copyright and license details.
+
+#![no_std]
+#![no_main]
+
+// Ensure we halt the program on panic (if we don't mention this crate it won't
+// be linked)
+use panic_halt as _;
+
+// Alias for our HAL crate
+use rp2040_hal as hal;
+
+// Some traits we need
+use hal::{
+    fugit::RateExtU32,
+    gpio::bank0::{Gpio20, Gpio21},
+    i2c::Controller,
+    I2C,
+};
+
+// Import required types & traits.
+use embedded_hal_async::i2c::I2c;
+use hal::{
+    gpio::{FunctionI2C, Pin, PullUp},
+    pac::{self, interrupt},
+    Clock,
+};
+
+/// The linker will place this boot block at the start of our program image. We
+/// need this to help the ROM bootloader get our code up and running.
+/// Note: This boot block is not necessary when using a rp-hal based BSP
+/// as the BSPs already perform this step.
+#[link_section = ".boot2"]
+#[used]
+pub static BOOT2: [u8; 256] = rp2040_boot2::BOOT_LOADER_GENERIC_03H;
+
+/// External high-speed crystal on the Raspberry Pi Pico board is 12 MHz. Adjust
+/// if your board has a different frequency
+const XTAL_FREQ_HZ: u32 = 12_000_000u32;
+
+/// Bind the interrupt handler with the peripheral
+#[interrupt]
+unsafe fn I2C0_IRQ() {
+    use hal::async_utils::AsyncPeripheral;
+    I2C::<pac::I2C0, (Gpio20, Gpio21), Controller>::on_interrupt();
+}
+
+/// The function configures the RP2040 peripherals, then performs a single I²C
+/// write to a fixed address.
+async fn demo() {
+    let mut pac = pac::Peripherals::take().unwrap();
+
+    // Set up the watchdog driver - needed by the clock setup code
+    let mut watchdog = hal::Watchdog::new(pac.WATCHDOG);
+
+    // Configure the clocks
+    let clocks = hal::clocks::init_clocks_and_plls(
+        XTAL_FREQ_HZ,
+        pac.XOSC,
+        pac.CLOCKS,
+        pac.PLL_SYS,
+        pac.PLL_USB,
+        &mut pac.RESETS,
+        &mut watchdog,
+    )
+    .ok()
+    .unwrap();
+
+    // The single-cycle I/O block controls our GPIO pins
+    let sio = hal::Sio::new(pac.SIO);
+
+    // Set the pins to their default state
+    let pins = hal::gpio::Pins::new(
+        pac.IO_BANK0,
+        pac.PADS_BANK0,
+        sio.gpio_bank0,
+        &mut pac.RESETS,
+    );
+
+    // Configure two pins as being I²C, not GPIO
+    let sda_pin: Pin<_, FunctionI2C, PullUp> = pins.gpio20.reconfigure();
+    let scl_pin: Pin<_, FunctionI2C, PullUp> = pins.gpio21.reconfigure();
+
+    // Create the I²C drive, using the two pre-configured pins. This will fail
+    // at compile time if the pins are in the wrong mode, or if this I²C
+    // peripheral isn't available on these pins!
+    let mut i2c = hal::I2C::new_controller(
+        pac.I2C0,
+        sda_pin,
+        scl_pin,
+        400.kHz(),
+        &mut pac.RESETS,
+        clocks.system_clock.freq(),
+    );
+
+    // Unmask the interrupt in the NVIC to let the core wake up & enter the interrupt handler.
+    // Each core has its own NVIC so these needs to executed from the core where the IRQ are
+    // expected.
+    unsafe {
+        pac::NVIC::unpend(hal::pac::Interrupt::I2C0_IRQ);
+        pac::NVIC::unmask(hal::pac::Interrupt::I2C0_IRQ);
+    }
+
+    // Asynchronously write three bytes to the I²C device with 7-bit address 0x2C
+    i2c.write(0x76u8, &[1, 2, 3]).await.unwrap();
+
+    // Demo finish - just loop until reset
+    core::future::pending().await
+}
+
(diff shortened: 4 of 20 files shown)


rp2040-hal/src/async.rs
```diff
@@ -0,0 +1,18 @@
+macro_rules! block_on {
+    ($me:ident, $is_ready:ident ($($arg:expr),*), $setup_flags:ident ($($arg2:expr),*)) => {
+        future::poll_fn(|cx| {
```
jannic: Uh, I didn't know about `poll_fn`. TIL!
That makes https://github.com/jannic/nb-async-adapter/blob/main/src/lib.rs mostly obsolete, I guess :smile:

rp2040-hal/src/i2c.rs
```diff
@@ -230,29 +230,47 @@ valid_pins! {
     }
 }
 
-/// Operational mode of the I2C peripheral.
+/// Operational mode of the I2C Block.
```
jannic, line 233: Do you think this is an improvement? I'd prefer "peripheral". (Very subjective, of course!)

rp2040-hal/Cargo.toml
```diff
@@ -54,6 +55,7 @@ hd44780-driver = "0.4.0"
 pio-proc = "0.2.0"
 dht-sensor = "0.2.1"
 rand = { version = "0.8.5", default-features = false }
+nostd_async = { version = "0.6.1", features = ["cortex_m"] }
```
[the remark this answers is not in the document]
ithinuel: AFAIU Embassy opted to have its own embedded-hal impl from scratch in embassy-rp.
Is there usecases where it'd make sense for an embassy user to rely on rp2040-hal's drivers instead?

I figured demonstrating an alternative light-weight executor would make a good example.

rp2040-hal/src/i2c.rs
ithinuel: I renamed these to remove ambiguity with peripheral the IP block in the rp2040 chip and peripheral the operational mode of the IP block.

I don't have a strong opinion about IP Block/Peripheral/Controller in place of Peripheral/Slave/Master, feel free to offer alternatives :)

ithinuel:
This reminded me that we do not have doc nor the ability to really use IRQ based application for i2c. I shall fix that too in an upcoming PR.

rp2040-hal/examples/i2c_async.rs
```diff
@@ -0,0 +1,108 @@
...
+/// as the BSPs already perform this step.
+#[link_section = ".boot2"]
+#[used]
+pub static BOOT2: [u8; 256] = rp2040_boot2::BOOT_LOADER_GENERIC_03H;
+
+/// External high-speed crystal on the Raspberry Pi Pico board is 12 MHz. Adjust
+/// if your board has a different frequency
+const XTAL_FREQ_HZ: u32 = 12_000_000u32;
+
+/// The function configures the RP2040 peripherals, reads the temperature from
```
thejpster, line 45: It does not, in fact, do this.

rp2040-hal/examples/i2c_async_irq.rs
```diff
@@ -0,0 +1,143 @@
...
+/// Waker setter method updating the waker of the async task.
+pub fn i2c_waker_setter(waker: Waker) {
+    critical_section::with(|cs| {
+        *WAKER.borrow_ref_mut(cs) = Some(waker);
+    });
+}
+
+#[interrupt]
+fn I2C1_IRQ() {
+    let i2c1 = unsafe { &rp2040_hal::pac::Peripherals::steal().I2C1 };
```
thejpster, line 62: Should add a note here about why this is acceptable. Something about no other code (main loop or interrupt) will use I2C1 concurrently, and thus you won't get a race when modifying the ic_intr_mask register on I2C1.

rp2040-hal/examples/i2c_async_irq.rs
```diff
@@ -0,0 +1,143 @@
...
+    i2c1.ic_intr_mask.modify(|_, w| {
+        w.m_tx_abrt()
+            .enabled()
+            .m_tx_empty()
+            .enabled()
+            .m_rx_full()
+            .enabled()
+    });
+    critical_section::with(|cs| WAKER.borrow_ref_mut(cs).take().map(Waker::wake));
```
thejpster, line 71: What does this do? Can we write it out as multiple lines with intermediate variable names and some comments?

thejpster: I think I'd also prefer `if let Some(waker) = borrowed_global_waker.take() { Wake::wake(waker) }` or similar, even if clippy might say "yeah, but you can just use .map(Waker::wake)", because it's non-obvious where why you would grab a global and then map it to a different type, and then drop the value.

rp2040-hal/Cargo.toml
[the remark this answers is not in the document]
jannic: Yes, embassy-rp is independent from rp2040-hal. But embassy-executor should be compatible with rp2040-hal. (Also, AFAIU. It's not like I did much async embedded stuff yet.)

For learning about how async works, I very much like the idea of a light-weight stand-alone executor. However I'm not so sure if it's the best example if someone just wants to start writing an async firmware. I _think_ embassy is by far the most popular embedded async executor, and it would be best to show novices the most common way of using async. It's much easier to find help in case something goes wrong.

Anyway, that's a broader discussion and we don't need to decide about it in this pull request. For now, the examples are perfectly find.

rp2040-hal/src/i2c.rs
jannic: Good point. I'll think about it.

rp2040-hal/src/i2c/controller/eh1_0_async.rs
```diff
@@ -0,0 +1,306 @@
...
+        }
+
+        Ok(())
+    }
+
+    /// Writes to the i2c bus consuming bytes for the given iterator.
+    async fn write_iter_async<A, U>(&mut self, address: A, bytes: U) -> Result<(), Error>
+    where
+        U: IntoIterator<Item = u8>,
+        A: AddressMode + 'static + Into<u16>,
```
jannic, line 225: I wonder if we should try to keep inherent implementations free from embedded-hal defined types. If we use them in our own API, it'll be difficult to upgrade embedded-hal later.

We did that in a few places with embedded-hal 0.2, and it'll make moving everything to embedded-hal 1.0 more difficult.

Also, we do not even support 10bit address mode yet, afaik.

rp2040-hal/src/i2c.rs
jannic: What about Controller / Target instead of Master / Slave? Then Peripheral doesn't need renaming.
Controller / Target seems to be the terminology chosen by NXP (successor of Philips, inventor of I2C) so it's as official as it can get.
https://community.nxp.com/t5/Other-NXP-Products/Alternative-inclusive-terms-for-master-and-slave-in-I2S-I2C/m-p/1245101#M10574
The same words are used by wikipedia, https://en.wikipedia.org/wiki/I%C2%B2C

thejpster:
I'm happy once jannic is happy

rp2040-hal/src/i2c/controller/eh1_0_async.rs
```diff
@@ -0,0 +1,365 @@
...
+
+            // else enqueue
+            let last = bytes.peek().is_none();
+            self.i2c.ic_data_cmd.write(|w| {
+                if do_stop && last {
+                    w.stop().enable();
+                } else {
+                    w.stop().disable();
+                }
+                unsafe { w.dat().bits(byte) }
```
jannic, line 280:
```suggestion
                // Safety: the `.bits()` fields are marked `unsafe` but all bit values are valid
                unsafe { w.dat().bits(byte) }
```

rp2040-hal/src/i2c/controller/eh1_0_async.rs
```diff
@@ -0,0 +1,365 @@
...
+impl<'i2c, A, B, F, G, H, T> future::Future for PollI2c<'i2c, A, B, F, G, H, T>
+where
+    F: FnMut(&mut I2C<A, B, Controller<Async>>) -> Poll<T>,
+    G: FnMut(&mut I2C<A, B, Controller<Async>>),
+    H: FnMut(&mut I2C<A, B, Controller<Async>>),
+{
+    type Output = T;
+
+    fn poll(self: core::pin::Pin<&mut Self>, cx: &mut core::task::Context<'_>) -> Poll<T> {
+        // SAFETY: We are not moving anything.
```
jannic, line 36: I don't think this is sufficient to guarantee safety. If I'm not mistaken, creating a `&mut` aliasing some other reference is enough to invoke undefined behavior.

jannic: Sorry - I didn't properly notice the reason for the `unsafe` block in the first place.
It's not about the reference being unique or not. We already have a unique reference to that data, so we know it's not aliased. It's just that the data is inside a `Pin`.

rp2040-hal/examples/i2c_async_irq.rs
ithinuel: One things is the clearing of the flags from the IRQ, another one is to soundly manage the waker.

I can't think of a solution that does not invole stealing for the former since we give that instance to the driver.

For the latter, I beleive this is executor specific, so we cannot really have a one size fits all. At worst, there should exist some `critical-section` based channel implementations. I haven't checked though.

rp2040-hal/src/i2c/controller/eh1_0_async.rs
ithinuel: Yes, the concern here is about the pinned/unpinned value.
This is what is done in `poll_fn` so I figured it was safe to do this although I haven't found something that'd explain this.
The doc for pin and unpin aren't very clear to me.

rp2040-hal/src/i2c.rs
ithinuel: Sounds good to me.
That covers the operation modes. Should we rename `Peripheral` into `Target`?

What about the IP Block ? Is this what we should call `peripheral`?

rp2040-hal/src/i2c/controller/eh1_0_async.rs
ithinuel: rather than the inherent method, should I impl [the crate](https://crates.io/crates/i2c-write-iter) I made to cover the `write_iter` that was removed between 0.2.7 and 1.0 ?

jannic: Doesn't that complicate things further? It would mean that users of those methods would need to depend on your crate as well and import the traits, before they can call the functions.

rp2040-hal/src/i2c.rs
jannic: Exactly. I think that way we can minimize confusion.

ithinuel: created https://github.com/rp-rs/rp-hal/issues/752 to track the change.

rp2040-hal/examples/i2c_async_irq.rs
jannic: I don't think the waker stuff is necessarily executor specific. We could for example use the utility classes from embassy_sync that seem to be executor independent:

> Futures from this crate can run on any executor.

(From https://docs.embassy.dev/embassy-sync/git/default/index.html)

I must admit that I don't fully understand the implications, and that I still need to learn a lot about async rust. But having all the low level stuff shown in this example in user code doesn't seem very attractive.
Look at the corresponding embassy code: https://github.com/embassy-rs/embassy/blob/main/examples/rp/src/bin/i2c_async.rs
Not as single `unsafe`, no mention of a waker. Sure, that's embassy specific code. But I see no reason why it shouldn't be portable to other executors. In fact, as I understand it, embassy is designed so the HALs it provides should actually be portable to other executors.

rp2040-hal/src/i2c/controller/eh1_0_async.rs
```diff
@@ -0,0 +1,364 @@
...
+    };
+}
+
+impl<T, Sda, Scl> I2C<T, (Sda, Scl), Controller<Async>>
+where
+    T: SubsystemReset + Deref<Target = RegisterBlock>,
+    Sda: ValidPinSda<T>,
+    Scl: ValidPinScl<T>,
+{
+    /// Configures the I2C peripheral to work in controller mode and implement async.await APIs.
```
jannic, line 94: Perhaps add a warning regarding cancellation?
```suggestion
    /// Configures the I2C peripheral to work in controller mode and implement async.await APIs.
    ///
    /// Warning: Cancelling a Future returned by this controller (that is, dropping it before `poll` returned `Ready`)
    /// can cause async code to block for relatively long time spans, if the code running on `drop` needs to
    /// terminate running i2c transfers. If your code can't tolerate such delays, make sure to always finish polling
    /// the futures. 
```

rp2040-hal/examples/i2c_async_irq.rs
```diff
@@ -0,0 +1,164 @@
...
+/// Waker setter method updating the waker of the async task.
+pub fn i2c_on_pending(waker: Waker) {
+    critical_section::with(|cs| {
+        *SAVED_WAKER.borrow_ref_mut(cs) = Some(waker);
+    });
+}
+
+fn i2c_on_cancel() {
+    critical_section::with(|cs| {
+        *SAVED_WAKER.borrow_ref_mut(cs) = None;
```
jannic, line 62: Is this really needed? What would happen if it wasn't done?
- the `Waker` inside `SAVED_WAKER` would stay around. That doesn't cost any additional resources (the memory usage of `SAVED_WAKER` doesn't change).
- it may happen that a later `I2C1_IRQ` interrupt calls `Waker::wake(waker)`. That may lead to a spurious wakeups, but that should be cheap - probably cheaper than taking the critical section in `i2c_on_cancel`

Are there any adverse effects of keeping the waker around that I'm missing?

rp2040-hal/src/i2c/controller/eh1_0_async.rs
ithinuel: Indeed. I wish it hadn't been removed from the embedded-hal in the first place :/

I'll make this driver impl the trait and keep this dependency feature-gated.

rp2040-hal/examples/i2c_async_irq.rs
ithinuel: > We could for example use the utility classes from embassy_sync that seem to be executor independent

Potentially, `embassy-sync` defines an [`AtomicWaker`](https://github.com/embassy-rs/embassy/blob/main/embassy-sync/src/waitqueue/atomic_waker.rs#L8) type that essentially does what this example does without embassy. I'm not convinced adding this dependency would solve our issues though.

> Not as single unsafe, no mention of a waker. Sure, that's embassy specific code

It's hidden behind a macro but there definitely is unsafe and wakers involved. The `bind_interruts` macro generates an unsafe IRQ handlers that calls an unsafe method from the driver.

I could use this approach too, at least with the rp2040-hal owned static storage for the storage. That'd save use from needing to expose those callbacks.

I'll explore that. A thing I'd like to avoid is having an unused I2C instance from taking space for its static atomic that'd never be used.

https://github.com/embassy-rs/embassy/blob/main/embassy-rp/src/lib.rs#L105-L124
https://github.com/embassy-rs/embassy/blob/main/embassy-rp/src/i2c.rs#L335-L341

jannic: > It's hidden behind a macro but there definitely is unsafe and wakers involved.

Sure - if you dig deep enough, everything is using unsafe. The point is: The HAL does the heavy lifting, the user can use a safe abstraction.

> A thing I'd like to avoid is having an unused I2C instance from taking space for its static atomic that'd never be used.

Keeping the abstractions zero-cost, especially when they are not used, is a worthy goal. That said, I wouldn't object to a few tens of bytes constant RAM overhead if it significantly improves usability.

ithinuel: A little update.
I made progress and the implementation looks better than I anticipated.
However, as I progressed on it I noticed several bugs so I'm solving them as they are discovered.
Mostly around input validation, error handling and stop generation and processing.

thejpster:
If this is draft because we have questions around the async-bits, maybe we can update the non-async stuff to be e-h 1.0 compatible in another PR and then rebase this over it? I'm keen to get e-h 1.0 support out of the door because I have other projects stuck until that happens.

ithinuel:
This is in draft because I'm updating i2c implementation as I'm also adding on-board-test suites for controller/target sync/async.
Please, let's procees with #753 first and I'll rebase this PR on main after it's merged.

ithinuel:
Finally it is there !!

rp2040-hal/src/i2c/peripheral.rs
```diff
@@ -94,120 +112,209 @@ where
...
 }
 
-impl<T: Deref<Target = I2CBlock>, PINS> I2CPeripheralEventIterator<T, PINS> {
+/// SAFETY: Takes a non-mutable reference to RegisterBlock but mutates its `ic_intr_mask` register.
+unsafe fn unmask_intr(i2c: &RegisterBlock) {
+    // SAFETY: 0 is a valid value meaning all irq masked.
+    i2c.ic_intr_mask.write_with_zero(|w| {
+        // Only keep these IRQ enabled.
+        w.m_start_det()
+            .disabled()
```
jannic, line 146: The implicit double negation makes this difficult to understand.
Am I right that `disabled` means that a 1 is written to the register bit, meaning that the masking is disabled, ie. the interrupt is active?

rp2040-hal/src/i2c/peripheral.rs
```diff
@@ -94,120 +112,209 @@ where
...
+        me.i2c.ic_enable.write(|w| w.enable().enabled());
+
+        me
     }
 }
 
-impl<T: Deref<Target = I2CBlock>, PINS> I2CPeripheralEventIterator<T, PINS> {
+/// SAFETY: Takes a non-mutable reference to RegisterBlock but mutates its `ic_intr_mask` register.
+unsafe fn unmask_intr(i2c: &RegisterBlock) {
+    // SAFETY: 0 is a valid value meaning all irq masked.
```
jannic, line 142: I don't understand these safety comments.

rp2040-hal/src/i2c/peripheral.rs
```diff
@@ -94,120 +112,209 @@ where
             // hold scl when fifo's full
             w.rx_fifo_full_hld_ctrl().enabled();
             w.ic_restart_en().enabled();
-            w.tx_empty_ctrl().enabled()
+            w.ic_10bitaddr_slave().variant(A::BIT_ADDR_S);
+            //w.tx_empty_ctrl().enabled();
```
jannic, line 116: Did you keep this comment on accident?

ithinuel: Yes

rp2040-hal/examples/i2c.rs
```diff
@@ -94,7 +94,7 @@ fn main() -> ! {
     );
 
     // Write three bytes to the I²C device with 7-bit address 0x2C
-    i2c.write(0x2c, &[1, 2, 3]).unwrap();
+    i2c.write(0x2Cu8, &[1, 2, 3]).unwrap();
```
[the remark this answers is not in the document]
jannic: After thinking more about this, it doesn't seem to be a big issue in practice. While it is ugly in such an artificially short example, in practice it is much more probable that the address is defined somewhere at the top, or passed in as a parameter, and so there is a natural place where the `u8` type can be, without creating an unreadable mess like `0x2Cu8` :-)

rp2040-hal/src/i2c/peripheral.rs
ithinuel: Yes, `disabled` means the mask is disabled, so the interrupt is `not` masked which is done by setting the corresponding bit in the field.

jannic:
IMHO this pull request contains unrelated changes that could be discussed and merged separately.

Changes that could be separate pull request include:
- unconditionally including defmt.x in the linker options
- adding on-target-tests to the top-level workspace
- some cleanups to the CI pipeline

Or am I missing something and these changes are inherently connected with the async i2c support?
(That's part of the problem: Because these changes are part of a bigger pull request, they are missing proper reasoning. If they were separate, there would probably be an explanation in the description of the pull request.)

ithinuel:
@jannic I understand. I created a PR for each commit (and address the previous reviews)

jannic:
> @jannic I understand. I created a PR for each commit (and address the previous reviews)

Thank you!

BTW, as many review comments I made sound so negative: That's just because I'm actively looking for things that I don't particularly like. Overall, you did a great job, and I'm sorry that it takes me so long to complete the review. It's just a lot of changes, and I'm not experienced enough with async rust that I can quickly glance over the changes to understand what's going on.

rp2040-hal/src/async_utils.rs
```diff
@@ -0,0 +1,144 @@
...
+                if let Some(waker) = self.waker.borrow(cs).take() {
+                    Waker::wake(waker);
+                }
+            });
+        }
+        pub fn register(&self, waker: &Waker) {
+            critical_section::with(|cs| {
+                let old = self.waker.borrow(cs).replace(Some(waker.clone()));
+                if old.map(|old| !old.will_wake(waker)).unwrap_or(false) {
+                    panic!("Only one task may be able to await on this driver.");
```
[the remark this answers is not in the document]
jannic: I'm still nervous about this code. It's not that I can definitely say that there's something wrong. It's just that it's not obvious enough that it is correct.

And `will_wake` is allowed to return `false` whenever it wants. The docs say "This function works on a best-effort basis, and may return false even when the Wakers would awaken the same task." Couldn't that cause a panic?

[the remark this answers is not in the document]
jannic: Another idea: Isn't it allowed to call `poll` with varying wakers?
Poll's doc seems to allow it and specifies what should happen: "Note that on multiple calls to poll, only the [Waker](https://doc.rust-lang.org/beta/core/task/struct.Waker.html) from the [Context](https://doc.rust-lang.org/beta/core/task/struct.Context.html) passed to the most recent call should be scheduled to receive a wakeup."
That could also trigger the `!old.will_wake(waker)` branch with the panic.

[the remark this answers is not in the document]
ithinuel: > Isn't it allowed to call poll with varying wakers?
It could but all these wakers will be waking the same task otherwise more fundamental invariant have been broken, e.g. Only 1 future (and therefore task) may hold a mutable reference to an instance of i2c and be awaiting on that instance.

> "Note that on multiple calls to poll, only the Waker from the Context passed to the most recent call should be scheduled to receive a wakeup."

Indeed, however "the Waker from the Context passed to the most recent call" is a grey concept. If an IRQ wakes a running future (eg during the poll_fn) the irq does not yet know about that most recent waker. If that had to be strongly upheld, [`Waker::clone_from`](https://doc.rust-lang.org/src/core/task/wake.rs.html#393-397) wouldn't make much sense to have (to point being to take the right take).

> Couldn't that cause a panic?
That could indeed 🤔 I might have forgotten about that. I had troubles with that while developing so that was helpful but if that's not reliable, we have not choice but to leave the user in the dark on that one.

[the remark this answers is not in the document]
ithinuel: I removed the panic.

rp2040-hal/src/i2c/controller.rs
```diff
@@ -100,89 +115,136 @@ where
...
+
+        let addr = addr.into();
         self.i2c.ic_tar.write(|w| unsafe { w.ic_tar().bits(addr) });
         self.i2c.ic_enable.write(|w| w.enable().enabled());
+        Ok(())
     }
 
-    fn read_and_clear_abort_reason(&mut self) -> Option<u32> {
+    #[inline]
+    fn read_and_clear_abort_reason(&mut self) -> Result<(), Error> {
```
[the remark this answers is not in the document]
jannic: What about `check_and_clear_abort_reason`?
Making it obvious that the abort reason is cleared isn't wrong, so it can be kept in the name. And `check_` doesn't imply that getting an abort reason is a successful result, in the same way as `read_` does.

on-target-tests/tests/i2c_loopback.rs
```diff
@@ -0,0 +1,128 @@
+//! This test needs a connection between:
+//!
+//! | from GPIO (pico Pin) | to GPIO (pico Pin) |
+//! | -------------------- | ------------------ |
+//! |         0 (1)        |       2 (4)        |
+//! |         1 (2)        |       3 (5)        |
```
jannic, line 6: Please also mention this in `on-target-tests/README.md`. It already describes the connection necessary for the SPI test, so the I2C connections should be there as well.

jannic (approved):
I would be lying if I said I understood every detail of this pull request.
But as far as I understand it, it looks good.

As github shows all the changes in the pull request anyway, even if some of the commits were actually already merged by some other pull requests, I did not review #763 and #764 separately. Not sure what's the best way to review multiple changes individually, short of having completely separate branches.

rp2040-hal/src/async_utils.rs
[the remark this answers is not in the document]
ithinuel: Thank you, I never thought about such a usecase.
I think that’d fall indeed under the case where "will_wake" may or may not return true, so we’d have a spurious panic.
The panic may have triggered on the first or the second call, so quite unpredictable.

ithinuel:
If merging this PR does not auto close the two "preceding", I’ll just close them manually adding the reference to this PR in the comment.
I don’t really mind about the details as long as it’s deemed worthy to be merged and it makes it to `main` :)

PR #753, opened 2024-01-10
Make embedded-hal 1.0.0 non-optional

jannic:
~~Marked as draft because I still have to check if the code needs some clean-up.~~

### DIFF
--- a/rp2040-hal/examples/spi.rs
+++ b/rp2040-hal/examples/spi.rs
@@ -89,7 +89,7 @@ fn main() -> ! {
         &mut pac.RESETS,
         clocks.peripheral_clock.freq(),
         16.MHz(),
-        embedded_hal_0_2::spi::MODE_0,
+        embedded_hal::spi::MODE_0,
     );
 
     // Write out 0, ignore return value
--- a/rp2040-hal/examples/spi.rs
+++ b/rp2040-hal/examples/spi.rs
@@ -89,7 +89,7 @@ fn main() -> ! {
         &mut pac.RESETS,
         clocks.peripheral_clock.freq(),
         16.MHz(),
-        embedded_hal::spi::MODE_0,
+        embedded_hal_0_2::spi::MODE_0,
     );
 
     // Write out 0, ignore return value
--- a/rp2040-hal/examples/spi_dma.rs
+++ b/rp2040-hal/examples/spi_dma.rs
@@ -11,7 +11,7 @@
 
 use cortex_m::singleton;
 use cortex_m_rt::entry;
-use embedded_hal_0_2::digital::v2::OutputPin;
+use embedded_hal::digital::OutputPin;
 use hal::dma::{bidirectional, DMAExt};
 use hal::fugit::RateExtU32;
 use hal::pac;
--- a/rp2040-hal/examples/spi_dma.rs
+++ b/rp2040-hal/examples/spi_dma.rs
@@ -67,7 +67,7 @@ fn main() -> ! {
         &mut pac.RESETS,
         clocks.peripheral_clock.freq(),
         16_000_000u32.Hz(),
-        embedded_hal_0_2::spi::MODE_0,
+        embedded_hal::spi::MODE_0,
     );
 
     // Initialize DMA.
--- a/rp2040-hal/examples/spi_dma.rs
+++ b/rp2040-hal/examples/spi_dma.rs
@@ -11,7 +11,7 @@
 
 use cortex_m::singleton;
 use cortex_m_rt::entry;
-use embedded_hal::digital::v2::OutputPin;
+use embedded_hal_0_2::digital::v2::OutputPin;
 use hal::dma::{bidirectional, DMAExt};
 use hal::fugit::RateExtU32;
 use hal::pac;
@@ -67,7 +67,7 @@ fn main() -> ! {
         &mut pac.RESETS,
         clocks.peripheral_clock.freq(),
         16_000_000u32.Hz(),
-        embedded_hal::spi::MODE_0,
+        embedded_hal_0_2::spi::MODE_0,
     );
 
     // Initialize DMA.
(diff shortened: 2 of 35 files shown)


ithinuel:
Shall we make eh1 the embedded_hal and rename 0.2 as embedded_hal_02 ?

When we introduce async, should we make it optional or not?

9names:
> If you make async optional it should be enabled by default IMO.

We should just include it unconditionally. There is no benefit to having it optional.

> v1 should just be `embedded_hal` and v0.2 should be treated as legacy.

It makes no difference to the HAL or it's users what the dependencies are called internally. It only matters that we implement the traits in the HAL so that drivers that depend on those trait versions work.

jannic:
I also think there's no reason to make async optional. It should be possible to implement it in a way that has minimal if async is not used in a firmware. (There is some impact because of additional dependencies which will increase compilation time. I'd hope that this increase won't be significant.)

Regarding the internal naming: That's one of the possible cleanups I was thinking about. Not critical because it should not affect the API, but we still should aim for well readable code within the HAL.

Sympatron:
I looked into renaming `embedded-hal-1` internally and it is a lot more complicated than I thought. This should probably be done in a separate PR so this one can go through quickly.

jannic:
I updated several files to use the eh-1 traits without renaming, and instead rename the eh-0.2 if there are conflicts.
This is mainly cosmetic (just internal, not changing the API).
A notable exception is `spi::FrameFormat`, which uses `Mode` from embedded-hal in one of its variants. I changed that to use `Mode` from embedded-hal 1.0.
To avoid conflicts with #747, I skipped updating the i2c code.

jannic:
> I propose changing the examples to use the new traits by default. Basically, you shouldn't have to know about the 0.2 traits unless you need the backwards-compatibility.
> 
> See https://github.com/rp-rs/rp-hal/tree/extra-1_0-impls.

Yes, definitely!
Should I merge your changes into this pull request, or should we merge this one first and you add a separate one to update the examples?

jonathanpallant:
If you like my proposal, please do merge it into your branch, so we get can it into main in one hit.

rp2040-hal/src/timer.rs
```diff
@@ -189,11 +188,13 @@ impl eh1_0_alpha::delay::DelayNs for Timer {
     }
 }
 
-/// Implementation of the embedded_hal::Timer traits using rp2040_hal::timer counter
+/// Implementation of the embedded_hal_0_2::Timer traits using rp2040_hal::timer counter.
+///
+/// There is no Embedded HAL 1.0 equivalent at this time.
```
jannic, line 193: Perhaps we should note that in many cases, `DelayNs` (which is implemented by `Timer`) could be used instead?

I see two use cases where `CountDown` can't be replaced by `DelayNs`:
- you don't want to block, but do something else when `wait` returns `WouldBlock`
- you use the periodic timer (and manually doing repeated `delay_ms()` calls would be too imprecise)

rp2040-hal/examples/spi.rs
```diff
@@ -89,7 +89,7 @@ fn main() -> ! {
         &mut pac.RESETS,
         clocks.peripheral_clock.freq(),
         16.MHz(),
-        embedded_hal::spi::MODE_0,
+        embedded_hal_0_2::spi::MODE_0,
```
thejpster, line 92: This should use HAL 1.0

rp2040-hal/examples/spi_dma.rs
```diff
@@ -67,7 +67,7 @@ fn main() -> ! {
         &mut pac.RESETS,
         clocks.peripheral_clock.freq(),
         16_000_000u32.Hz(),
-        embedded_hal::spi::MODE_0,
+        embedded_hal_0_2::spi::MODE_0,
```
thejpster, line 70: This should use HAL 1.0

rp2040-hal/examples/watchdog.rs
```diff
@@ -21,8 +21,9 @@ use rp2040_hal as hal;
 use hal::pac;
 
 // Some traits we need
-use embedded_hal::digital::v2::OutputPin;
-use embedded_hal::watchdog::{Watchdog, WatchdogEnable};
+use embedded_hal::digital::OutputPin;
+// Embedded HAL 1.0.0 doesn't have an Watchdog trait, so use the one from 0.2
+use embedded_hal_0_2::watchdog::{Watchdog, WatchdogEnable};
```
thejpster, line 26: We should probably ensure that any e-h 0.2 related functionality is available at inherent methods so you don't *have* to use eh 0.2. It's not like forcing people to use e-h 0.2 gives you portability, because I expect most HALs to drop support for it.

thejpster:
I find that *little and often* is a good policy, and this PR is already large, so I think we should merge it and then work on more clean-up stuff later. In particular, we need this in so we can rebase the I2C changes over it, and we then need to work on embedded-io support for the serial ports.

jannic:
> I find that _little and often_ is a good policy, and this PR is already large, so I think we should merge it and then work on more clean-up stuff later. In particular, we need this in so we can rebase the I2C changes over it, and we then need to work on embedded-io support for the serial ports.

Definitely. That's why I marked the pull request as ready for review 3 days ago. (Sorry, I forgot to update the top comment accordingly.)
I don't claim that the patch contains all changes that should be done. It's a first step, but there are still many improvements possible. Please review and decide if the general approach is ok, and look for mistakes I might have made. If it's good enough, we can merge it and then improve in separate pull requests.

rp2040-hal/examples/spi.rs
jannic: :+1: 
I relied upon the compiler notifying me if the 0.2 types were still used. But as `new` only requires `Into<FrameFormat>`, and that is implemented for both 0.2 and 1.0 types, no error was shown.

thejpster (approved):

PR #757, opened 2024-01-15
Add on-chip voltage regulator (VREG) voltage setting function

AkiyukiOkayasu:
Added an on-chip voltage regulator voltage setting function.  
Increasing core voltage may be necessary for overclocking. Implemented with reference to [hardware_vreg](https://www.raspberrypi.com/documentation/pico-sdk/hardware.html#hardware_vreg) in pico-sdk.


rp2040-hal/src/vreg.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// 1.25V
+    Voltage1_25 = 0b1110u8,
+    /// 1.30V
+    Voltage1_30 = 0b1111u8,
+}
+
+/// Set voltage to the on-chip voltage regulator
+///
+/// * `voltage` - The voltage (from enumeration \ref vreg_voltage) to apply to the voltage regulator
+pub fn vreg_set_voltage(vreg_dev: &mut VREG_AND_CHIP_RESET, voltage: VregVoltage) {
```
ithinuel, line 43: I'm not convinced by the benefit from having this in the hal when the function merely adds an enum.

Is it really better than just calling `vreg_dev.modify(|_, w| w.vsel.variant(voltage)` ?
Where `voltage` is a `VSEL_A`.

Except I realize writing this and checking the doc that `VSEL_A` does not exist.

IMHO the real solution would be to add the enum to the svd patch so that it gets generated in the PAC.
We wouldn't need this method in the hal.

ithinuel (changes requested):

rp2040-hal/src/vreg.rs
AkiyukiOkayasu: I agree that VSEL_A is needed for PAC, I am not familiar with svd and may struggle, but I will try to contribute to rp2040_pac.  
Still, I think HAL should have more human-friendly functions.

`pac.VREG_AND_CHIP_RESET.write(|w| w.vsel.variant(voltage));` and `vreg_set_voltage(&mut pac.VREG_AND_CHIP_RESET, VSEL_A::Voltage1_20);` the latter is easier for humans to understand that the voltage is being set.
Also, if provided in HAL, additional documentation such as Usage can be written, as [watchdog](https://docs.rs/rp2040-hal/latest/rp2040_hal/watchdog/index.html) does.

on-target-tests/tests/vreg_change_voltage.rs
```diff
@@ -0,0 +1,113 @@
...
+        )
+        .ok()
+        .unwrap();
+        State {
+            vreg: pac.VREG_AND_CHIP_RESET,
+        }
+    }
+
+    #[test]
+    fn change_onchip_regulator_voltage(state: &mut State) {
```
jannic, line 64: How do we know that this test works reliably on every device?
The datasheet says: "Note that RP2040 may not operate reliably with its digital core supply (DVDD) at a voltage other than 1.1V."
Perhaps this test should only cover a smaller range of voltages, eg. 1.05 - 1.15V?

AkiyukiOkayasu: In my experience, I have set the core voltage to 1.2V on more than 50 RP2040s and have never had a problem. Also, in the [thread about overclocking on the RP2040](https://forums.raspberrypi.com/viewtopic.php?t=301902), there is no example of a failure to increase the core voltage.  
However, this is not normal usage, there is no guarantee that it will work reliably on all RP2040s.

I noticed that with the default system clock, lowering the core voltage too much can cause problems. It should be a smaller test case range, as you mentioned. Especially in the direction of lowering the core voltage.  
1.20V should be fine, so i'll reduce it to 1.05 - 1.20V. 
If you are concerned, I would reduce it more. Reducing it is easy:)

AkiyukiOkayasu:
Reduced test cases and improved documentation comments. And rebased commit log to tidy it up.

AkiyukiOkayasu:
As one might expect from the slope of this graph, 0.80V may not work at the default system clock frequency.
Screenshot 2024-02-26 at 16 15 39
https://forums.raspberrypi.com/viewtopic.php?t=301902&start=50#p1820686

However, the RP2040 C/C++ SDK has a [minimum voltage of 0.85V](https://www.raspberrypi.com/documentation/pico-sdk/vreg_8h.html#ab2019bc9d9ea4c0c02ba2a853ffcd0b0), so 0.80V seems like an unusual value.
I tested at the default system clock frequency and 0.80V did not work. The 0.85V did work, but there is no guarantee that it will work with all RP2040s.

ithinuel:
What I mean is that the tests as implemented there do not provide anyproof that things are:
- implemented correctly
- actually work understress

Writing to and Reading from a register is trivial enough that I don’t think it requires an on-target tests.

For example, if there’s a possibility to actually verify that the voltage is applied as expected that would make a good test case.
EG, if the core voltage is exposed to an external pin that can easily be probed with an ADC, and the tests were to sample that value and check it matches the configuration, that would be great.

AkiyukiOkayasu:
@ithinuel 
> EG, if the core voltage is exposed to an external pin that can easily be probed with an ADC, and the tests were to sample that value and check it matches the configuration, that would be great.

With the Raspberry Pi Pico, that is impossible.  
Unfortunately, it appears that the test has to be removed.

ithinuel (approved):

PR #761, opened 2024-02-06
Bump msrv from 1.65 to 1.75

ithinuel:

PR #762, opened 2024-02-06
Move on-target-tests back to the work space

ithinuel:
- requires #761


jannic:
> * requires [Bump msrv from 1.65 to 1.75 #761](https://github.com/rp-rs/rp-hal/pull/761)

Are you sure? I reverted that change and both the CI pipeline and the on-target tests still succeed.

Asides from that: @9names, you excluded the on-target-tests from the workspace in https://github.com/rp-rs/rp-hal/pull/457. As it seems like with ithinuels changes, the CI pipeline is happy with them added back to the workspace, do you think there's still some reason to keep them separate?

9names:
I have no reason to exclude them from the workspace

9names (approved):
Changes look fine to me

ithinuel:
@jannic requires in the sense that I'd like to avoid having to rebase the PRs, so having them merged in-order should help reduce unnecessary work.

jannic:
> @jannic "requires" in the sense that I'd like to avoid having to rebase the PRs, so having them merged in-order should help reduce unnecessary work. (and this PR includes the changes of its preceding)

Ah ok, understand. Personally I prefer truly independent pull requests, but they of course have the risk of merge conflicts.

PR #763, opened 2024-02-06
Bump rustc nightly for cargo udeps

ithinuel:
Requires:
- #761 
- #762


thejpster (approved):
Who originally removed the on target tests from the workspace?

jannic:
> Who originally removed the on target tests from the workspace?

See https://github.com/rp-rs/rp-hal/pull/762

Sorry, splitting the large pull-request into multiple smaller ones maybe was a bad idea. I thought it would make reviewing easier, but it seems like it caused some confusion.

PR #765, opened 2024-02-06
[i2c] Implement i2c-write-iter traits

ithinuel:
Requires:

- #761
- #762
- #763
- #764
- #747

### DIFF
--- a/on-target-tests/Cargo.toml
+++ b/on-target-tests/Cargo.toml
@@ -71,5 +71,5 @@ nostd_async = { version = "0.6.1", features = ["wfe"] }
 futures = { version = "0.3.30", default-features = false, features = [
   "async-await",
 ] }
-i2c-write-iter = "1.0.0"
+i2c-write-iter = { version = "1.0.0", features = ["async"] }
 itertools = { version = "0.12.0", default-features = false }
--- a/on-target-tests/Cargo.toml
+++ b/on-target-tests/Cargo.toml
@@ -51,9 +51,10 @@ defmt-test = "0.3.1"
 panic-probe = { version = "0.3", features = ["print-defmt"] }
 
 rp2040-hal = { path = "../rp2040-hal", features = [
-    "defmt",
-    "critical-section-impl",
-    "rt",
+  "critical-section-impl",
+  "defmt",
+  "rt",
+  "i2c-write-iter",
 ] }
 # Needed to set spi frequencies
 fugit = "0.3.6"
@@ -70,4 +71,5 @@ nostd_async = { version = "0.6.1", features = ["wfe"] }
 futures = { version = "0.3.30", default-features = false, features = [
   "async-await",
 ] }
+i2c-write-iter = "1.0.0"
 itertools = { version = "0.12.0", default-features = false }
--- a/on-target-tests/tests/i2c_tests/non_blocking.rs
+++ b/on-target-tests/tests/i2c_tests/non_blocking.rs
@@ -359,3 +359,34 @@ pub async fn transaction<A: ValidAddress>(
     .collect();
     assert_eq!(g, h);
 }
+
+pub async fn transaction_iter<A: ValidAddress>(state: &mut State, addr: A) {
+    use i2c_write_iter::non_blocking::I2cIter;
+    reset(state, addr);
+
+    let samples: FIFOBuffer = Generator::seq().take(25).collect();
+    let controller = state.controller.as_mut().expect("controller's missing.");
+    let case = async {
+        controller
+            .transaction_iter(
+                addr,
+                [i2c_write_iter::Operation::WriteIter(
+                    samples.clone().into_iter(),
+                )]
+                .into_iter(),
+            )
+            .await
+            .expect("Successful write_iter");
+        wait_with(&state.payload, |p| p.stop_cnt != 1).await;
+    };
+
+    futures::select_biased! {
+        _ = case.fuse() => {}
+        _ = target_handler(
+            &state.payload,
+            state.target.as_mut().take().expect("target’s missing"),
+        ).fuse() => {}
+    }
+
+    assert_eq!(samples, state.payload.borrow().vec);
+}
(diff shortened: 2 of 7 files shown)


on-target-tests/Cargo.toml
```diff
@@ -70,4 +71,6 @@ nostd_async = { version = "0.6.1", features = ["wfe"] }
 futures = { version = "0.3.30", default-features = false, features = [
   "async-await",
 ] }
+i2c-write-iter = "1.0.0"
 itertools = { version = "0.12.0", default-features = false }
+portable-atomic = { version = "1.6.0", features = ["critical-section"] }
```
jannic, line 76: This seems to be unused. At least compiling the tests still works if I comment this line out.

jannic: In fact, I can revert all changes to this file and it still builds.

ithinuel: Good catch, thanks!

ithinuel: Oh, well the portable-atomic dep is redundant with the feature already enabled on heapless.
And it builds without those changes because the dum-dum I am forgot to add tests for this.

ithinuel: Added some tests to check it actually works.

rp2040-hal/src/i2c/controller.rs
```diff
@@ -458,3 +481,16 @@ impl<A: ValidAddress, T: Deref<Target = Block>, PINS> eh1::I2c<A> for I2C<T, PIN
...
 }
+
+#[cfg(feature = "i2c-write-iter")]
+impl<A: i2c_write_iter::AddressMode + ValidAddress, T: Deref<Target = Block>, PINS>
+    i2c_write_iter::I2cIter<A> for I2C<T, PINS, Controller>
+{
+    fn transaction_iter<'a, O, B>(&mut self, address: A, operations: O) -> Result<(), Self::Error>
+    where
+        O: IntoIterator<Item = i2c_write_iter::Operation<'a, B>>,
+        B: IntoIterator<Item = u8>,
```
jannic, line 492: This signature causes a (IMHO) surprising limitation: All Operations must use exactly the same type B.
So for example, this doesn't work:
```
        controller
            .transaction_iter(
                addr,
                [
                    i2c_write_iter::Operation::WriteIter(
                        [1, 2, 3],
                    ),
                    i2c_write_iter::Operation::WriteIter(
                        [1, 2, 3, 4],
                    ),
                ]
                .into_iter(),
            )
            .await
```

The error message is:
```
error[E0308]: mismatched types
   --> on-target-tests/tests/i2c_tests/non_blocking.rs:379:25
    |
378 |                     i2c_write_iter::Operation::WriteIter(
    |                     ------------------------------------ arguments to this enum variant are incorrect
379 |                         [1, 2, 3, 4],
    |                         ^^^^^^^^^^^^ expected an array with a fixed size of 3 elements, found one with 4 elements
    |
help: the type constructed contains `[u8; 4]` due to the type of the argument passed
   --> on-target-tests/tests/i2c_tests/non_blocking.rs:378:21
    |
378 | /                     i2c_write_iter::Operation::WriteIter(
379 | |                         [1, 2, 3, 4],
    | |                         ------------ this argument influences the type of `i2c_write_iter`
380 | |                         // samples.clone().into_iter(),
381 | |                     ),
    | |_____________________^
```

As you can see I just modified the on-target-test so send two arrays of different sizes.
Am I missing something or does this limit the usefulness of the iterator based API?

ithinuel: This should be easily worked around using a slice rather than an array:

```rust
controller
  .transaction_iter(
    addr,
    [
      i2c_write_iter::Operation::WriteIter(
        [1, 2, 3][..].iter().cloned(),
      ),
      i2c_write_iter::Operation::WriteIter(
        [1, 2, 3, 4][..].iter().cloned(),
      ),
    ]
    .into_iter(),
  )
  .await
```

This might not be the smoothest, but it should work.
I’m also open to alternatives :)

ithinuel: See: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a1851ce0c724a18df386fe3bb913ae93

Some level of `Either` (or equivalent eg with more variants) can be used too, as long as they are `IntoIterator<Item = u8>`.

jannic: > This should be easily worked around using a slice rather than an array:

Sorry, while this obviously works, it's not easy. :smiling_face_with_tear: 

I don't know a good alternative. If it was easy, I'm sure someone would have found a solution in https://github.com/rust-embedded/embedded-hal/issues/367 and the function would still be in embedded-hal.

on-target-tests/tests/i2c_tests/non_blocking.rs
```diff
@@ -359,3 +359,34 @@ pub async fn transaction<A: ValidAddress>(
...
+    reset(state, addr);
+
+    let samples: FIFOBuffer = Generator::seq().take(25).collect();
+    let controller = state.controller.as_mut().expect("controller's missing.");
+    let case = async {
+        controller
+            .transaction_iter(
+                addr,
+                [i2c_write_iter::Operation::WriteIter(
+                    samples.clone().into_iter(),
```
jannic, line 374: Perhaps this should be:
```suggestion
                    samples.iter().cloned(),
```
Advantages:
- doesn't clone the whole buffer
- works the same way if samples is a slice

on-target-tests/tests/i2c_tests/non_blocking.rs
```diff
@@ -359,3 +359,34 @@ pub async fn transaction<A: ValidAddress>(
...
+                )]
+                .into_iter(),
```
jannic, line 376: this `into_iter()` isn't needed, as the array already implements `IntoIterator`.

jannic (approved):

PR #768, opened 2024-02-12
gpio: Add getters to the various pad overrides (not type tracked)

ithinuel:


rp2040-hal/src/gpio/mod.rs
```diff
@@ -535,6 +547,18 @@ impl<I: PinId, F: func::Function, P: PullType> Pin<I, F, P> {
         self.id.io_ctrl().modify(|_, w| w.inover().variant(variant));
     }
 
+    /// Set the output enable override.
```
9names, line 550:
```suggestion
    /// Get the output enable override.
```

jannic (approved):

PR #769, opened 2024-02-12
`Cargo.toml`: add `categories` & `keywords`

rursprung:
this makes it easier to find the crate, esp. when filtering by category.

note that this will only become visible on crates.io once a new release has been published.

### DIFF
--- a/rp2040-hal/Cargo.toml
+++ b/rp2040-hal/Cargo.toml
@@ -4,10 +4,12 @@ version = "0.9.1"
 authors = ["The rp-rs Developers"]
 edition = "2021"
 homepage = "https://github.com/rp-rs/rp-hal"
-description = "A Rust Embeded-HAL impl for the rp2040 microcontroller"
+description = "A Rust Embedded-HAL impl for the rp2040 microcontroller"
 license = "MIT OR Apache-2.0"
 rust-version = "1.75"
 repository = "https://github.com/rp-rs/rp-hal"
+categories = ["embedded", "hardware-support", "no-std", "no-std::no-alloc"]
+keywords = ["embedded", "hal", "raspberry-pi", "rp2040", "embedded-hal"]
 
 [package.metadata.docs.rs]
 features = ["rt", "rom-v2-intrinsics", "defmt", "rtic-monotonic"]


rp2040-hal-macros/Cargo.toml
```diff
@@ -6,6 +6,8 @@ readme = "README.md"
 version = "0.1.0"
 edition = "2021"
 repository = "https://github.com/rp-rs/rp-hal"
+categories = ["embedded", "hardware-support", "no-std", "no-std::no-alloc"]
+keywords = ["embedded", "hal", "raspberry-pi", "rp2040", "embedded-hal"]
```
jannic, line 10: Are these actually useful? I don't think someone searching for one of these keywords would be interested in rp2040-hal-macros.

rp2040-hal/Cargo.toml
```diff
@@ -4,10 +4,12 @@ version = "0.9.1"
 authors = ["The rp-rs Developers"]
 edition = "2021"
 homepage = "https://github.com/rp-rs/rp-hal"
-description = "A Rust Embeded-HAL impl for the rp2040 microcontroller"
+description = "A Rust Embedded-HAL impl for the rp2040 microcontroller"
 license = "MIT OR Apache-2.0"
 rust-version = "1.75"
 repository = "https://github.com/rp-rs/rp-hal"
+categories = ["embedded", "hardware-support", "no-std", "no-std::no-alloc"]
+keywords = ["embedded", "hal", "raspberry-pi", "rp2040", "embedded-hal"]
```
jannic, line 12: Not sure about "raspberry-pi", as this usually means the raspberry single board computers. But as `embassy-rp` also includes this keyword, there's at least some precedent.

rursprung: yeah, i actually took the keywords from embassy (minus `async`)

rp2040-hal-macros/Cargo.toml
rursprung: i think it might help depending on what people search for. e.g. ["embedded rp2040"](https://crates.io/search?q=embedded+rp2040) only yields `rp2040-hal` as the 5th result - though i of course have no numbers for how often people search things like this. i think the categories are more important, but the keywords certainly don't hurt

jannic: This comment is only about the macros crate. Finding `rp2040-hal` easily is of course useful!
But who needs `rp2040-hal-macros`, if not as a dependency of `rp2040-hal`?

rursprung: that's true. i just thought "well, that can't hurt". i can also remove it here if you prefer

rursprung: ok, done

jannic (approved):

jannic:
Thank you!

PR #770, opened 2024-02-14
Updates necessary because of rp2040-pac update to v0.6.0

AkiyukiOkayasu:
This is a proof of concept regarding https://github.com/rp-rs/rp2040-pac/pull/91.  
It is not yet suitable to be merged.

Still some work to be done and will be updated from time to time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,6 @@ rp2040-hal = { path = "./rp2040-hal" }
 
 [patch.crates-io]
 rp2040-hal = { path = "./rp2040-hal" }
-rp2040-pac = { git = "https://github.com/rp-rs/rp2040-pac/", rev = "615f600538a13f750438dafe07014ffad0cad51f" }
 
 # Those profiles are only used for `on-target-tests` compilation
 [profile.dev]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ rp2040-hal = { path = "./rp2040-hal" }
 
 [patch.crates-io]
 rp2040-hal = { path = "./rp2040-hal" }
-rp2040-pac = { git = "https://github.com/AkiyukiOkayasu/rp2040-pac/", rev = "4488b0e9a82ae4ae3ba699cf0fc7a3974abc93c6" }
+rp2040-pac = { git = "https://github.com/AkiyukiOkayasu/rp2040-pac/", rev = "a886967ad7f6a923ffdbcbdbd94e934c8b773ef5" }
 
 # Those profiles are only used for `on-target-tests` compilation
 [profile.dev]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ rp2040-hal = { path = "./rp2040-hal" }
 
 [patch.crates-io]
 rp2040-hal = { path = "./rp2040-hal" }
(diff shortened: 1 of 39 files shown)


AkiyukiOkayasu:
I have finished making changes to my updated PAC for use at HAL.  
When a new version of rp2040-pac is released, I will need to remove the following.

https://github.com/AkiyukiOkayasu/rp-hal/blob/bd428da43df775eba855fcb9c52af119359d532d/Cargo.toml#L11

AkiyukiOkayasu:
I noticed that dma_spi_loopback_u16 and dma_spi_loopback_u8 in on-target-tests are no longer terminating.  

transfer.wait() is no longer terminating. 
https://github.com/AkiyukiOkayasu/rp-hal/blob/bd428da43df775eba855fcb9c52af119359d532d/on-target-tests/tests/dma_spi_loopback_u16.rs#L126

jannic:
> I noticed that dma_spi_loopback_u16 and dma_spi_loopback_u8 in on-target-tests are no longer terminating.
> 
> transfer.wait() is no longer terminating. https://github.com/AkiyukiOkayasu/rp-hal/blob/bd428da43df775eba855fcb9c52af119359d532d/on-target-tests/tests/dma_spi_loopback_u16.rs#L126

I found the bug: https://github.com/AkiyukiOkayasu/rp-hal/pull/1

Overall, it looks like the updated PAC does work, but we need to carefully check that there are no more bugs of that kind in the ported rp2040-hal. Anyway, I think we can merge https://github.com/rp-rs/rp2040-pac/pull/91.

jannic:
A quick search found three more occurrences of that pattern, which are likely bugs as well:
```
$ rg "&.*\(\) as"
rp2040-hal/src/adc.rs
697:        DmaReadTarget(&self.adc.device.fifo() as *const _ as u32, PhantomData)

rp2040-hal/src/uart/reader.rs
247:        (&self.device.uartdr() as *const _ as u32, u32::MAX)

rp2040-hal/src/uart/writer.rs
203:        (&self.device.uartdr() as *const _ as u32, u32::MAX)
```

Right now I'm too tired to check and fix them. And the regex might be too simplistic to find all instances.

AkiyukiOkayasu:
Thank you for fixing it.  

Perhaps these may also be bugs. I will check when I have time.  
https://github.com/rp-rs/rp-hal/blob/e175d2a4865a42f75f84fde4682eae7cc41d8daf/rp2040-hal/src/pio.rs#L1431-L1436

https://github.com/rp-rs/rp-hal/blob/e175d2a4865a42f75f84fde4682eae7cc41d8daf/rp2040-hal/src/pio.rs#L1625-L1630

AkiyukiOkayasu:
@jannic I think I have no more to fix. Please give your review.

jannic:
> @jannic I think I have no more to fix. Please give your review.

I already looked through all the changes, they look good to me. I'll do the final approval as soon as it can actually be merged, ie. when pac v0.6.0 is released.

If we decide to include the new naming scheme from https://github.com/rust-embedded/svd2rust/commit/30eb39898041c734023ba9124d684003ba96bc37 in rp2040-pac 0.6.0, we need to update rp-hal as well.

In any case, I don't want to delay this much longer. If there's no new release of svd2rust really soon now, we should go with what's available now, merge https://github.com/rp-rs/rp2040-pac/pull/92 as it is, release pac v0.6, and merge this pull request. And then release rp2040-hal 0.10.0 as well.

jannic (approved):

PR #779, opened 2024-03-03
Fix some more nightly clippy warnings

jannic:
I'm not sure how useful these `Default` implementations are in practice. But I like the pattern "if there is a non-argument new() constructor, `Default` should be implemented as well".
https://rust-lang.github.io/rust-clippy/master/index.html#/new_without_default


ithinuel (approved):
Me neither (especially on private types) but if it makes clippy happy 🤷

PR #788, opened 2024-04-03
Enable transfer size of PIO DMA to be specified

### LINKED ISSUE #554: Configure TransmittedWord for PIO SM DMA
It is not possible to configure DMA to use a 1 byte transfer size when writing to the PIO SM. It is currently [hard-coded to u32](https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/src/pio.rs#L1583) -

```rust
impl<SM: ValidStateMachine> WriteTarget for Tx<SM> {
    type TransmittedWord = u32;
    // ...
}
```

I believe this is possible based on section 2.5. DMA section of the datasheet -

> The transfer size can be either 32, 16, or 8 bits. This is configured once per channel: source transfer size and
destination transfer size are the same. The DMA performs standard byte lane replication on narrow writes, so byte data
is available in all 4 bytes of the databus, and halfword data in both halfwords.

I'm happy to make the change but would like some advice on where to start. 

I can see that SPI has a similar-ish thing going on and there's a [generic to support it](https://github.com/rp-rs/rp-hal/blob/636992e311bdd0a457e5797208a1c0b888756208/rp2040-hal/src/spi.rs#L422). Would a new generic parameter representing expected word size on [Tx](https://github.com/rp-rs/rp-hal/blob/636992e311bdd0a457e5797208a1c0b888756208/rp2040-hal/src/pio.rs#L1412) and associated field (Defaults to `u32`. Named `write_target_word_size`?) on the [PIOBuilder](https://github.com/rp-rs/rp-hal/blob/636992e311bdd0a457e5797208a1c0b888756208/rp2040-hal/src/pio.rs#L1840) be appropriate?

Whilst I don't have a use-case for it myself, if I was adding the above would it also be appropriate to add another to allow control of read target word size?

### Context 

This is to avoid pre-procesing the data (e.g. [as performed here allow for left shifting](https://github.com/rp-rs/ws2812-pio-rs/blob/main/src/lib.rs#L175)). I plan on combining this with a [change in the rgb crate](https://github.com/kornelski/rust-rgb/commit/fb667ccdbd481bbd589a812018146d52beed783c) to allow for direct DMA transfers.

jsgf:
Rather than always assuming PIO DMA transfers are u32, also allow u16 and u8. The hardware will replicate sub-32-bit transfers across the whole 32 bit word, which is the same as accessing the Tx/Rx fifo directly.

This implements a rather general type-level description of a transfer, so maybe it should go in the `dma` module rather than being `pio` specific?

This fixes #554

### DIFF
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -592,10 +592,10 @@ impl<SM: ValidStateMachine, State> StateMachine<SM, State> {
     ///
     /// The program can be uninstalled to free space once it is no longer used by any state
     /// machine.
-    pub fn uninit(
+    pub fn uninit<RxSize, TxSize>(
         mut self,
-        _rx: Rx<SM>,
-        _tx: Tx<SM>,
+        _rx: Rx<SM, RxSize>,
+        _tx: Tx<SM, TxSize>,
     ) -> (UninitStateMachine<SM>, InstalledProgram<SM::PIO>) {
         self.sm.set_enabled(false);
         (self.sm, self.program)
@@ -1314,7 +1314,7 @@ unsafe impl<SM: ValidStateMachine + Send, RxSize> Send for Rx<SM, RxSize> {}
 
 // Safety: `Rx` is marked Send so ensure all accesses remain atomic and no new concurrent accesses
 // are added.
-impl<SM: ValidStateMachine> Rx<SM> {
+impl<SM: ValidStateMachine, RxSize: TransferSize> Rx<SM, RxSize> {
     unsafe fn block(&self) -> &pac::pio0::RegisterBlock {
         &*self.block
     }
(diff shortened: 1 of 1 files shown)


jannic:
Thanks, Jeremy! I didn't have a closer look yet, but my first impression is that this is a good way to fix #554.

What's not so nice is that it looks like the new type parameters sometimes need to be specified manually, as you can see from the examples failing to build in the CI pipeline. Do you have an idea if that can be avoided somehow?

jsgf:
I haven't had a chance to dig into it, but I suspect that second commit was a mistake.

jsgf:
OK I think that should be OK now.

jsgf:
The current API doesn't infer the transfer size at `build` as I'd hoped, but using the `transfer_size` method seems fine.

jannic (approved):
LGTM. I also did a quick test writing u8 values to the PIO and checking that the values are replicated to all 4 byte lanes. That should prove that `Byte` writes are correctly executed on real hardware.

9names (approved):
Looks fine to me too.
As noted, would be good to lift this into DMA layer as a good next step, and maybe a layer-agnostic type after that, but those should be in their own PR's.

PR #790, opened 2024-04-08
Move TransferSize to `dma` module

jsgf:
This is a generic way to represent a DMA channel transfer size at the type level, so move it to `dma` where it can be used more generally.

Followup to #788

### DIFF
--- a/rp2040-hal/src/dma/mod.rs
+++ b/rp2040-hal/src/dma/mod.rs
@@ -295,3 +295,33 @@ pub enum DMAError {
     /// was specified.
     IllegalConfig,
 }
+
+/// Constraint on transfer size types
+pub trait TransferSize: Sealed {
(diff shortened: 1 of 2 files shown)


jannic (approved):

ithinuel (commented):
It seems to be purely type level and not specific to a peripheral.
Why not hosting it in `crate::typelevel` ?

jsgf:
It is specifically about DMA though. Are you thinking it could be used for other things?

jannic:
Even if it could be used for other things than DMA, I don't think it would fit into the typelevel module. They are representing a particular concept, the bit width of the DMA transfers. `typelevel` is about much more abstract concepts. I don't think it makes sense to put those together.

ithinuel (approved):

PR #791, opened 2024-04-08
Make PioBuilder::from_installed_program consistent with from_program

jsgf:
Make the default shift direction `Left` for `from_installed_program` to match `from_program`.
Otherwise this causes very unexpected breakage when updating to the recommended API.


jannic (changes requested):
Please have a look at the doc comment on from_installed_program: Changing the default shift direction was the very reason for creating a new function `from_installed_program`.

Perhaps we can make the note more obvious somehow?

jannic:
I guess you saw a deprecation warning like this and just changed the method name without seeing the documentation:
```
warning: use of deprecated associated function `rp2040_hal::pio::PIOBuilder::<P>::from_program`: please use `from_installed_program` instead
```

Would it be an improvement to change that message? For example:
```
warning: use of deprecated associated function `rp2040_hal::pio::PIOBuilder::<P>::from_program`: please use `from_installed_program` instead and update shift direction if necessary
```
Or is that getting to long? Any alternative suggestions?

jsgf:
Hm, yes I did completely miss the documentation change - I was just reacting to the, erm, very surprising output from my DACs :)

Part of the problem was the VS code / Copilot suggested the change without me really having to read the message. I've put an explicit direction change into my code anyway, so it's more robust. I guess a possible option is to not have a default but take the direction as a parameter (either explicitly, or do something clever with typestates to require it in the builder). But that seems like overkill.

Issue #793, opened 2024-04-13
Support for #[thread_local]?


jsgf:
It would be useful to be able to use `#[thread_local]` to have functionally `static` variables which are per-core on rp2040.

Right now you can (with the nightly feature enabled) add the `#[thread_local]` to a static variable, but it fails to build for several reasons:
- the linker script doesn't handle the thread-local sections properly
- the compiler generates calls to `__aeabi_read_tp` which does not exist

The first is simple, but the second raises some interesting questions. `__aeabi_read_tp` seems to be the ABI's interface with the platform to get the per-"thread" base. It seems that commonly a coprocessor control register is used, but [picolib](https://github.com/picolibc/picolibc/blob/58d6157cc2135df5043d62c3e89feedc20ffcd57/newlib/libc/picolib/machine/arm/read_tp.S#L71) uses the SIO CPUID register on rp2040 to index into an array of per-core pointers. This seems like a reasonable mechanism.

The Raspberry Pi C/C++ SDK doesn't seem to support this at all, so offers no guidance.

Whatever the mechanism, the rp2040_hal::multicore APIs would need to be extended to expose how much per-thread memory is needed, to allow callers to provide suitable memory when spawning on a new core.

9names:
> It would be useful to be able to use #[thread_local] to have functionally static variables which are per-core on rp2040.

I'm having trouble coming up with a use-case where thread_local variables would provide a benefit, what did you have in mind?

jsgf:
I've been using @cbiffle's lilos executor for my projects. It's strictly single core but there's no reason why you couldn't run two instances, one on each rp2040 core. It relies on static vars for some of its state so I was thinking that thread local would allow two instances to coexist in ram while sharing code.

Otherwise I'd have to work out how to link in two instances with their own statics which would be annoyingly redundant.

This specific case doesn't seem all that niche, esp since 'static lifetime data is so useful for DMA and such.

jsgf:
> Whatever the mechanism, the rp2040_hal::multicore APIs would need to be extended to expose how much per-thread memory is needed, to allow callers to provide suitable memory when spawning on a new core.

Actually thinking about it, we could just statically allocate bss spaceat link time for the per-thread of for each core, and make this completely transparent to users.

PR #794, opened 2024-04-15
Support per-core state using #[thread_local]

jsgf:
- Controlled by `thread_local` feature
- Does not require nightly features, but using it does
- Intercepts `__pre_init` to copy .tdata into the per-core state
- Implements `__aeabi_read_tp` to returns per-core state for code compiler generates when accessing `thread_local` variables
- Needs linker script support to set up the layout and symbols

This is based on [picolibc](https://github.com/picolibc/picolibc/blob/58d6157cc2135df5043d62c3e89feedc20ffcd57/newlib/libc/picolib/machine/arm/read_tp.S#L71)'s support for TLS; its the only example I found of it for rp2040.

Prototype for #793

### DIFF
--- a/rp2040-hal/src/multicore.rs
+++ b/rp2040-hal/src/multicore.rs
@@ -335,18 +335,14 @@ impl<'p> Core<'p> {
 #[cfg(all(target_arch = "arm", feature = "thread_local"))]
 mod thread_local {
     use core::arch::global_asm;
-    use core::ptr::addr_of;
+    use core::ptr::{addr_of, addr_of_mut};
 
     extern "C" {
-        static TLS_CORE_0: u8;
-        static TLS_CORE_1: u8;
+        static mut TLS_CORE_0: u8;
+        static mut TLS_CORE_1: u8;
+        static __tdata_start: u8;
+        static __tdata_len: u8;
     }
-    // Not really a const pointer, but we reform it into mut in the asm
-    static mut TLS_STATE: [*const u8; 2] = [
-        // Point to linker-allocated space in .bss
-        unsafe { addr_of!(TLS_CORE_0) },
-        unsafe { addr_of!(TLS_CORE_1) },
-    ];
 
     // Define `__aeabi_read_tp` called by the compiler to get access to
     // thread-local storage.
@@ -358,41 +354,28 @@ mod thread_local {
         ".type __aeabi_read_tp,%function",
 
         "__aeabi_read_tp:",
-        "    push {{r1, lr}}",
-        "    ldr r1, =0xd0000000",      // Load SIO CPUID addr
-        "    ldr r1, [r1]",             // Get current CPUID
-        "    lsls r1, r1, #2",          // Scale by 4
-        "    ldr r0, ={tls_state}",     // Load TLS_STATE base addr
-        "    ldr r0, [r0, r1]",         // Load CPU per-thread
-        "    pop {{r1, pc}}",
+        "    ldr r0, =0xd0000000",      // Load SIO CPUID addr
+        "    ldr r0, [r0]",             // Load CPUID
+        "    cmp r0, #0",               // Check core 0
+        "    ldr r0, ={core_0}",        // Set TLS_CORE_0
+        "    beq 1f",                   // skip if done
+        "    ldr r0, ={core_1}",        // Set TLS_CORE_1
+        "1:  bx lr",
 
         ".popsection",
-        tls_state = sym TLS_STATE,
+        core_0 = sym TLS_CORE_0,
+        core_1 = sym TLS_CORE_1,
     }
 
     // Intercept __pre_init to hook into the startup code to copy the tdata into
     // TLS_CORE_[01].
-    global_asm! {
-        ".pushsection .text.__pre_init",
-        ".align 4",
-        ".p2align 4,,15",
-        ".global __pre_init",
-        ".type __pre_init,%function",
-
-        "__pre_init:",
-        "    push {{lr}}",
-        "    ldr r0, ={tls_core_0}",
-        "    ldr r1, =__tdata_start",
-        "    ldr r2, =__tdata_len",
-        "    bl __aeabi_memcpy",
-        "    ldr r0, ={tls_core_1}",
-        "    ldr r1, =__tdata_start",
-        "    ldr r2, =__tdata_len",
-        "    bl __aeabi_memcpy",
-        "    pop {{pc}}",
-
-        ".popsection",
-        tls_core_0 = sym TLS_CORE_0,
-        tls_core_1 = sym TLS_CORE_1,
+    //
+    // NB: Run as the very first thing, nothing has been initialized and memory
+    // could be in arbitrary state, so we only deal with things via raw pointers.
+    #[cortex_m_rt::pre_init]
+    unsafe fn tls_pre_init_hook() {
+        for dst in [addr_of_mut!(TLS_CORE_0), addr_of_mut!(TLS_CORE_1)] {
+            core::ptr::copy(addr_of!(__tdata_start), dst, addr_of!(__tdata_len) as usize);
+        }
     }
 }
--- a/rp2040-hal/src/multicore.rs
+++ b/rp2040-hal/src/multicore.rs
@@ -290,3 +290,68 @@ impl<'p> Core<'p> {
         }
     }
 }
+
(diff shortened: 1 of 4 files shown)


jannic:
A few random thoughts:
- How does this interact with re-spawning the second core? `core1.spawn()` is not marked unsafe and could be called multiple times. While that's probably not a good idea in most cases, it shouldn't be unsound.
- Please provide a usage example and some documentation. Even if this is still a prototype, knowing how it is meant to be used makes it easier to think about.
- Would it be useful to provide a hook to chain another `__pre_init` function somehow? Not sure how it could look like, and there are not too many use cases for `__pre_init` functions, but if you want to combine thread_local with something else that needs it, both features just defining the `__pre_init` symbol would clash.
- There are discussions about deprecating `static mut` in rust. I think the current consensus is that only references to `static mut` are going to be banned. But I'm not following the discussion closely.

jsgf:
@jannic Thanks for your comments.

> How does this interact with re-spawning the second core? core1.spawn() is not marked unsafe and could be called multiple times. While that's probably not a good idea in most cases, it shouldn't be unsound.

There's no problem with doing this in principle. The per-core data has the same semantics as regular static data/bss - it will be in the same state when you re-spawn on the core as it was before. This is a little different from traditional per-thread data, which will be re-initialized each time you start a thread. But I think since the core is a persistent entity, this current behaviour makes sense.

That said, it wouldn't be too hard to reinitialize the per-core data on spawn.

I don't think there's any additional soundness issues here. If the per-thread/core data is `static mut` then it will be subject to the normal safety requirements that mutable statics need.

> Please provide a usage example and some documentation

Will do.

> Would it be useful to provide a hook to chain another __pre_init function somehow? Not sure how it could look like, and there are not too many use cases for __pre_init functions, but if you want to combine thread_local with something else that needs it, both features just defining the __pre_init symbol would clash.

Yeah I was thinking about this. I really just need a hook to initialize the per-core data before the core 0 main comes up. I think it would make sense to add a dedicated hook in the cortex-m-rt startup code for this, since in principle this could apply to any multicore implementation.

Otherwise I don't have any great ideas for how to add a chainable __pre_init, unless it's something like ctors.

> There are discussions about deprecating static mut in rust. I think the current consensus is that only references to static mut are going to be banned. But I'm not following the discussion closely.

What's the context here? Are you concerned about this implementation's use of `static mut` or just the relationship between `static mut` and this feature in general? I haven't seen anything about deprecating it in general, but I'd be interested in pointers if you have them. I don't see a way of doing without it, esp for embedded uses, unless there's some alternative mechanism which gives access to non-stack, non-heap preallocated memory.

jannic:
> What's the context here?

For example https://internals.rust-lang.org/t/pre-rfc-deprecate-then-remove-static-mut/20072/106. Or https://github.com/rust-lang/rust/pull/117556/. I didn't follow it in detail, so your usage could be unrelated to the planned / discussed changes.

jsgf:
@jannic I see. The proposal is literally about `static mut`, not the general concept. We can use `SyncUnsafeCell` instead (or whatever mechanism is introduced).

jannic:
Exactly. And I don't think they'll break existing code, at least not with a long grace period. So using static mut now is not out of question. It's just that I'd prefer to use a different mechanism from the start if we introduce a new feature now. Of course, only if such a mechanism already exists.

rp2040-hal/src/multicore.rs
```diff
@@ -290,3 +331,68 @@ impl<'p> Core<'p> {
...
+        ".type __aeabi_read_tp,%function",
+
+        "__aeabi_read_tp:",
+        "    push {{r1, lr}}",
+        "    ldr r1, =0xd0000000",      // Load SIO CPUID addr
+        "    ldr r1, [r1]",             // Get current CPUID
+        "    lsls r1, r1, #2",          // Scale by 4
+        "    ldr r0, ={tls_state}",     // Load TLS_STATE base addr
+        "    ldr r0, [r0, r1]",         // Load CPU per-thread
+        "    pop {{r1, pc}}",
```
ithinuel, line 367: The [Run-time ABI for the Arm® Architecture](https://github.com/ARM-software/abi-aa/blob/main/rtabi32/rtabi32.rst#thread-local-storage-new-in-v201) specifies that only `r0` may be clobbered by `__aeabi_read_tp`.

This won’t be a visible issue until someone starts enabling networking & tls operations.

rp2040-hal/src/multicore.rs
```diff
@@ -290,3 +331,68 @@ impl<'p> Core<'p> {
...
+        "    ldr r0, ={tls_core_1}",
+        "    ldr r1, =__tdata_start",
+        "    ldr r2, =__tdata_len",
+        "    bl __aeabi_memcpy",
+        "    pop {{pc}}",
+
+        ".popsection",
+        tls_core_0 = sym TLS_CORE_0,
+        tls_core_1 = sym TLS_CORE_1,
+    }
```
ithinuel, line 397: Why not using a regular rust funtion placed in that section with the [`link_section` attribute](https://doc.rust-lang.org/reference/abi.html#the-link_section-attribute) ?

EG:
```rust
#[no_mangle]
#[link_section = ".text.__pre_init"]
unsafe fn tls_pre_init_hook() {
    for dts in [ addr_of!(TLS_CORE_0), addr_of!(TLS_CORE_1) ] {
        core::ptr::copy(addr_of!(__tdata_start), dst, addr_of!(__tdata_len));
    }
}
```

rp2040-hal/src/multicore.rs
```diff
@@ -290,3 +331,68 @@ impl<'p> Core<'p> {
...
+    extern "C" {
+        static TLS_CORE_0: u8;
+        static TLS_CORE_1: u8;
+    }
+    // Not really a const pointer, but we reform it into mut in the asm
+    static mut TLS_STATE: [*const u8; 2] = [
+        // Point to linker-allocated space in .bss
+        unsafe { addr_of!(TLS_CORE_0) },
+        unsafe { addr_of!(TLS_CORE_1) },
+    ];
```
ithinuel, line 349: Since we have a (small) set number of cores, I think we can do without a table and `_read_tp` to simply:
```
1. Load SIO::CPUID to r0
2. Compare with 0
3. Set `r0` to `addr_of!(TLS_CORE_0)` (preserves the flags set by 2.)
4. if equal flags set, skip next instr
5. Set `r0` to `addr_of!(TLS_CORE_1)`
6. return
```

ithinuel (changes requested):

rp2040-hal/src/multicore.rs
9names: I don't think encouraging people to put rust code in pre-init is a good idea (in general) - [there are warnings on the cortex-m pre_init macro for a reason](https://docs.rs/cortex-m-rt/latest/cortex_m_rt/attr.pre_init.html#safety).
rust-embedded moved away from having r0 written in rust for this same reason - it is easier to conform with conditions required pre-init in asm than it is in rust.

rp2040-hal/src/multicore.rs
jsgf: It's preserving r1 with push/pop. Am I missing something here?

rp2040-hal/src/multicore.rs
jsgf: Yeah I wasn't sure about whether Rust would be valid that early. Though the code using `addr_of` avoids forming any reference which could cause problems. And maybe sprinkling some `MaybeUninit` would make things a bit more robust.

rp2040-hal/src/multicore.rs
ithinuel: Oh right. 🤔 I guess it's still allowed to modify the stack pointer provided it restores it too.

jsgf:
Question: How can I config ci to handle a nightly only feature like this? 

Aside from that I'm still working on testing so it's not ready for merging.

ithinuel:
We only merge stable compatible code, so it’s not something we really have to deal with.

jsgf:
Ok so this is in the back burner until thread local is stabilized?

jsgf:
OK I've published this functionality as a separate crate:
- https://github.com/jsgf/rp2040-multicore-per-cpu
- https://crates.io/crates/rp2040-multicore-per-cpu

So we can close this for now, and restart the conversation when #[thread_local] is finalized.

PR #796, opened 2024-04-28
Preparations for release 0.10.1

9names:
Update CHANGELOG.md 
Set version to 0.10.1 in preparation for the release

This is a rebase of changes from main on top of 0.10.0, excluding PIO DMA changes:

Here is the selection of commits used via rebase -i
```text
pick 4a677a6 Simplify ceiling division in delay calculation
pick caa0ea0 Fix float_to_fix64 return value & docs for f32 trig functions
drop 9947302 Enable transfer size of PIO DMA to be specified
drop 6ce5672 Add some more type params
drop caaef72 Remove type params on `build` to avoid inference failures
drop 36a05f2 Fix Tx::transfer_size
drop efe574a Move TransferSize to `dma` module
pick 48dbfcc Slightly improve deprecation note on from_program
pick effcc46 Fix debugging after halt()
pick 0d819d3 Implement embedded_io traits for Reader/Writer
pick 8e15127 Implement `send_break` support
```

Leaving this as a draft PR so it doesn't get merged. Will tag the final commit hash as v0.10.1 and close this PR when reviewed.

PR #797, opened 2024-04-28
Preperations for release 0.10.2

9names:
Update CHANGELOG.md
Set version to 0.10.2 in preparation for the release

This is based on main, not 0.10.x branch.
The only changes from 0.10.1 -> 0.10.2 are the PIO DMA transfer size changes. We *think* these are not breaking changes, so we're doing 0.10.1 and 0.10.2 as a tactical release strategy - will yank 0.10.2 if this release causes people grief.

See #796 for the changes that were cherry-picked from main for that release.

### DIFF
--- a/rp2040-hal/CHANGELOG.md
+++ b/rp2040-hal/CHANGELOG.md
@@ -7,6 +7,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.10.2] - 2024-04-28
+
+### Changed
+
+- Move TransferSize to dma module - #790 @jsgf
+- Enable transfer size of PIO DMA to be specified - #788 @jsgf
+
+## [0.10.1] - 2024-04-28
+
+### Added
+
+- Implement send_break support - #700 @ithinuel
+- Implement embedded_io traits for Reader/Writer - #781 @jannic
+
+### Changed
+
+- Fix debugging after halt() - #785 @jannic
+- Slightly improve deprecation note on from_program - #792 @jannic
+- Fix float_to_fix64 return value & f32 trig function doc corrections - #787 @Text-Input
+- Simplify ceiling division in delay calculation - #783 @jannic
+
 ## [0.10.0] - 2024-03-10
 
 ### Added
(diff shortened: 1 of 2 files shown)


ithinuel (approved):

jannic:
I'm in favor of a 0.10.1 release, but I'm not yet convinced if 0.10.2 with the DMA changes should be released, and the git history looks a bit confusing: Why is there a v0.10.x branch containing (the preparations for) 0.10.1, but 0.10.2 is a merge request to the main branch?

IMHO it would be less confusing if we either decided that the DMA changes are actually fine for a 0.10.1 release and release it from the main branch, or accept that the risk of a breaking change is too high, release 0.10.1 from a separate branch, and don't do a 0.10.2 with the DMA changes. But is it important to have an easily understandable git history? Or do we mainly care about the actual releases on crates.io, and it doesn't matter much from what git branch they came?

9names:
Sorry about that, thought your comments on that branch were also broad approval of the approach.
I did consider pulling these changes into the 0.10.x branch, wasn't sure if that would be better or worse as far as keeping history goes.

If we pull the changes into 0.10.x instead, and then merge that into main to get the changelog updated does that make the git history more or less messy than it would be going this way?

jannic:
Don't worry, it doesn't make sense to hold off progress just in case someone might come up with another opinion. And cleanliness of the git history is only a minor detail. Usually, nobody cares.

Still, connecting v0.10.x and main with a merge commit is a good idea. That way, the v0.10.x branch will contain all 0.10.x releases, and we avoid the risk that a later 0.10.3 release accidentally misses the changes from 0.10.2.

About the possibly breaking changes: There are two relevant modifications between 0.10.1 and main:
- Adding type parameters to `Tx` and `Rx`, e.g. `pub struct Rx<SM: ValidStateMachine>` → `pub struct Rx<SM: ValidStateMachine, RxSize = Word>`
- Adding the same type parameters to `uninit`, `pub fn uninit(self, rx, tx)` → `pub fn uninit<RxSize, TxSize>(self, rx, tx)`

According to https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-parameter-default the first one is fine. And the example from https://doc.rust-lang.org/cargo/reference/semver.html#fn-generic-new where adding a type parameter to a function can be a breaking change doesn't apply here: Previously, `uninit` didn't have type parameters, so there shouldn't be any existing callers with the wrong number of type parameters. A few quick experiments with examples actually calling `uninit` didn't show any issues with that update either.

Issue #782, opened 2024-03-16
embedded_io::serial::Reader implementations drop data on error


jannic:
While working on https://github.com/rp-rs/rp-hal/pull/781 I noticed a flaw in the error handling of the reader implementation:
```
impl<D: UartDevice, P: ValidUartPinout<D>> embedded_io::Read for UartPeripheral<Enabled, D, P> {
    fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Error> {
        nb::block!(self.read_raw(buf)).map_err(|e| e.err_type)
    }
}
```
The error returned by `read_raw` does not only contain the error itself (in `err_type`) but also `pub discarded: &'err [u8]`, possibly data that was already read from the peripheral before the error occurred. That data gets silently dropped.

Instead, it may be better to check if the length of `discarded` is non-zero, and if so, return an `Ok` result containing that data. In that case, the error needs to be stored somewhere, to be returned in the next call to `Read`.

PR #798, opened 2024-05-01
Avoid discarding already read bytes on error

jannic:
Closes: #782

It would be preferable to avoid the code duplication between `UartPeripheral` and `Reader`. However, before starting a bigger code restructuring, I'd like to get some feedback on the general approach:
- Is it the right trade-off to reserve some memory in `UartPeripheral` and `Reader` to store the read error?
- What are alternative approaches?
  - Ignore the issue, accept discarding bytes on error. (Might be acceptable, assuming that serial errors are rare and the received message is probably garbled anyway. Should be clearly documented.)
  - Don't implement `embedded_io::Read` at all. (Breaking change, lost functionality, IMHO not acceptable)
  - Don't implement `embedded_io::Read` directly on `UartPeripheral` and `Reader`, but instead on a separate wrapper type. (Breaking change, slightly less nice API, zero-cost when not using `embedded_io::Read`)

### DIFF
--- a/rp2040-hal/src/uart/peripheral.rs
+++ b/rp2040-hal/src/uart/peripheral.rs
@@ -21,6 +21,7 @@ pub struct UartPeripheral<S: State, D: UartDevice, P: ValidUartPinout<D>> {
     device: D,
     _state: S,
     pins: P,
+    read_error: Option<ReadErrorType>,
 }
 
 impl<S: State, D: UartDevice, P: ValidUartPinout<D>> UartPeripheral<S, D, P> {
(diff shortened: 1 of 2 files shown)


thejpster (approved):

thejpster:
I think this is quite reasonable.

thejpster:
Leaving this for a second opinion. @9names ?

ithinuel:
If the UART has/is using a FIFO, I’m afraid there’s no way to tell when did the error occur.
So if we get a parity error for example, we cannot tell which byte is corrupted.

jannic:
IIRC the error bits are also stored in the FIFO, together with the data words they belong to.

9names:
image

9names:
image

PR #799, opened 2024-05-02
Fix oneshot adc read waiting indefinitely

mjptree:
Currently, the `Adc::read` typically gets stuck in an infinite loop when running in `OneShot` configuration. This bug was introduced in version 0.10 and does not exist on prior versions. It results from the `CS` register being read once outside of the wait loop and then never again.


jannic:
Thanks for finding and fixing this bug! (And sorry for introducing it...)
I guess we should work on the test coverage.

PR #803, opened 2024-05-18
Add support for aborting DMA transfers

### LINKED ISSUE #722: DMA abort
I'm implementing an SPI-like slave for a project, and I would like to be able to abort the DMA transfer if CS is de-asserted before completion.

Currently it looks like the HAL doesn't implement this feature at all.

Reading section 2.5.5.3. _Channel Abort_ of the datasheet, the sequence is fairly simple (including the errata workarounds):
- save and disable the channel's IRQ
- set the corresponding bit in the CHAN_ABORT register
- poll the busy flag until it clears
- clear the spurious interrupt
- restore the channel's IRQ enable

I'll probably PR this, just opening a tracking issue for now.

EliseZeroTwo:
This MR implements support for aborting DMA transfers, following 2.5.5.3 of the datasheet and #722.

This should close #722.

### DIFF
--- a/rp2040-hal/src/dma/single_buffer.rs
+++ b/rp2040-hal/src/dma/single_buffer.rs
@@ -112,4 +112,38 @@ where
 
         (self.ch, self.from, self.to)
     }
+
+    /// Aborts the current transfer, returning the channel and targets
+    pub fn abort(mut self) -> (CH, FROM, TO) {
+        let irq0_was_enabled = self.ch.is_enabled_irq0();
+        let irq1_was_enabled = self.ch.is_enabled_irq1();
+        self.ch.disable_irq0();
+        self.ch.disable_irq1();
+
+        let chan_abort = unsafe { &*crate::pac::DMA::ptr() }.chan_abort();
+        let abort_mask = (1 << self.ch.id()) as u16;
+
+        chan_abort.write(|w| unsafe { w.chan_abort().bits(abort_mask) });
+
+        while chan_abort.read().chan_abort().bits() != 0 {}
+
+        while !self.is_done() {}
+
+        self.ch.check_irq0();
+        self.ch.check_irq1();
+
+        if irq0_was_enabled {
+            self.ch.enable_irq0();
+        }
+
+        if irq1_was_enabled {
+            self.ch.enable_irq1();
+        }
+
+        // Make sure that memory contents reflect what the user intended.
+        cortex_m::asm::dsb();
+        compiler_fence(Ordering::SeqCst);
+
+        (self.ch, self.from, self.to)
+    }
 }
--- a/rp2040-hal/src/dma/single_channel.rs
+++ b/rp2040-hal/src/dma/single_channel.rs
@@ -31,6 +31,11 @@ pub trait SingleChannel: Sealed {
         }
     }
 
+    /// Check if the DMA_IRQ_0 signal for this channel is enabled.
+    fn is_enabled_irq0(&mut self) -> bool {
+        unsafe { ((*DMA::ptr()).inte0().read().bits() & (1 << self.id())) != 0 }
+    }
+
     #[deprecated(note = "Renamed to disable_irq0")]
     /// Disables the DMA_IRQ_0 signal for this channel.
     fn unlisten_irq0(&mut self) {
@@ -73,6 +78,11 @@ pub trait SingleChannel: Sealed {
         unsafe {
             write_bitmask_set((*DMA::ptr()).inte1().as_ptr(), 1 << self.id());
         }
+    }
+
+    /// Check if the DMA_IRQ_1 signal for this channel is enabled.
+    fn is_enabled_irq1(&mut self) -> bool {
+        unsafe { ((*DMA::ptr()).inte1().read().bits() & (1 << self.id())) != 0 }
     }
 
     #[deprecated(note = "Renamed to disable_irq1")]


rp2040-hal/src/dma/single_channel.rs
```diff
@@ -31,6 +31,14 @@ pub trait SingleChannel: Sealed {
         }
     }
 
+    /// Check if the DMA_IRQ_0 signal for this channel is enabled.
+    fn is_enabled_irq0(&mut self) -> bool {
+        // Safety: We only use the atomic alias of the register.
+        unsafe {
+            (*DMA::ptr()).inte0().read().bits() & 1 << self.id() != 0
+        }
```
9names, line 39: I'd feel more comfortable if this wasn't relying on operator precedence and order of operations.
```suggestion
        let enabled = unsafe {(*DMA::ptr()).inte0().read().bits() };
        (enabled & (1 << self.id())) != 0
```

rp2040-hal/src/dma/single_channel.rs
```diff
@@ -75,6 +83,14 @@ pub trait SingleChannel: Sealed {
         }
     }
 
+    /// Check if the DMA_IRQ_1 signal for this channel is enabled.
+    fn is_enabled_irq1(&mut self) -> bool {
+        // Safety: We only use the atomic alias of the register.
+        unsafe {
+            (*DMA::ptr()).inte1().read().bits() & 1 << self.id() != 0
+        }
```
9names, line 91: As above
```suggestion
        let enabled = unsafe {(*DMA::ptr()).inte1().read().bits() };
        (enabled & (1 << self.id())) != 0
```

9names (changes requested):

rp2040-hal/src/dma/single_buffer.rs
```diff
@@ -112,4 +112,35 @@ where
...
+        let irq0_was_enabled = self.ch.is_enabled_irq0();
+        let irq1_was_enabled = self.ch.is_enabled_irq1();
+        self.ch.disable_irq0();
+        self.ch.disable_irq1();
+
+        unsafe { &*crate::pac::DMA::ptr() }.chan_abort().write(|w| unsafe { w.bits(1 << self.ch.id() as u32)});
+
+        while unsafe { &*crate::pac::DMA::ptr() }.chan_abort().read().bits() != 0 {}
+
+        while !self.is_done() {};
```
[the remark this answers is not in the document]
EliseZeroTwo: This makes sense, I have changed it to be like this, however I left the `chan_abort` call on line 124 outside of the unsafe block as the function is not marked as unsafe and prevented another layer of nesting.

rp2040-hal/src/dma/single_channel.rs
EliseZeroTwo: Done! Thank you for catching this

9names (approved):
LGTM

rp2040-hal/src/dma/single_buffer.rs
```diff
@@ -112,4 +112,39 @@ where
...
     }
+
+    /// Aborts the current transfer, returning the channel and targets
+    pub fn abort(mut self) -> (CH, FROM, TO) {
+        let irq0_was_enabled = self.ch.is_enabled_irq0();
+        let irq1_was_enabled = self.ch.is_enabled_irq1();
+        self.ch.disable_irq0();
+        self.ch.disable_irq1();
+
+        // Safety: We only use the atomic alias of the register.
```
jannic, line 123: Is this really true? As I understand it, `chan_abort.write(|w| unsafe { w.chan_abort().bits(abort_mask) });` translates to a direct (volatile) write to the CHAN_ABORT register.

This _may_ be harmless. The datasheet isn't clear about what happens if you write a 0 to a bit in this register (ie. if it overwrites a previously set abort bit on some other channel). But even if this happens to be harmless, I think the safety comment is still misleading.

I think the quoted line should use `write_bitmask_set`.

rp2040-hal/src/dma/single_channel.rs
```diff
@@ -31,6 +31,12 @@ pub trait SingleChannel: Sealed {
         }
     }
 
+    /// Check if the DMA_IRQ_0 signal for this channel is enabled.
+    fn is_enabled_irq0(&mut self) -> bool {
+        // Safety: We only use the atomic alias of the register.
```
jannic, line 36: This is also not true. The read is fine, AFAICT, but it's not because of some atomic alias.

rp2040-hal/src/dma/single_channel.rs
```diff
@@ -75,6 +81,12 @@ pub trait SingleChannel: Sealed {
         }
     }
 
+    /// Check if the DMA_IRQ_1 signal for this channel is enabled.
+    fn is_enabled_irq1(&mut self) -> bool {
+        // Safety: We only use the atomic alias of the register.
```
jannic, line 86: as above

jannic (changes requested):

rp2040-hal/src/dma/single_buffer.rs
9names: It's another one of the registers that is labelled SC in the datasheet, but violates that contract:
"The bit will remain high until any in-flight transfers have been flushed through the address and data FIFOs"
https://github.com/raspberrypi/pico-sdk/issues/1466

I'd be inclined to err on the side of caution with this register, if only because of the other caveat in the datasheet:
"After writing, this register must be polled until it returns all-zero. Until this point, it is unsafe to restart the channel"

My naive intuition is that you should only need to wait for the channel you care about, not all-zero, so this is already surprising information.

> I think the quoted line should use write_bitmask_set.

To do the above to-the-letter this code should not be re-entrant, and we should only ever write to it when it is already all zeroes (so plain write/write_with_zero should suffice).
Should we wrap this in some kinda of lock (critical-section)?

rp2040-hal/src/dma/single_channel.rs
9names: To be fair this is a nit, you know this particular safety comment isn't ever going to lead to violated safety invariants.
I was going to fix it up port-merge to avoid having to press "approve" again on the CI run because this is a first time contributor.

jannic: Yes, of course, it doesn't actually affect correctness of the code. But it can be confusing for someone trying to understand the code. Instead of a wrong safety comment, better remove it or write something like TBD.

rp2040-hal/src/dma/single_buffer.rs
jannic: A critical section is only appropriate if we are sure that the operation strictly bounded to a few processor cycles. Otherwise it could cause unacceptable latency for interrupt handling. I'm not sure how long it could take, worst case, for the flag to get cleared.

jannic: The C SDK also does a simple write to the register:
https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_dma/include/hardware/dma.h#L527-L532
I think we would be fine doing the same, only the safety comment should be updated. What about:
```
// Safety: While we are not sure if this write call clears ABORT bits for other channels, we don't really care about these bits as we check for BUSY anyway, due to RP2040-E13. The C SDK works the same way.
```

EliseZeroTwo: Thank you for pointing this out, I was under the impression that it was due to how the comment was used in other places. I've addressed this now.

Sorry for taking a week to get around to this, I have been sick.

jannic: Don't worry, even without giving a reason it's totally fine to contribute to open source projects at whatever pace suits you best. I hope you are well again!

jannic (approved):

PR #807, opened 2024-06-01
Fix set_low() and set_high() implementation for OutputPin

martinsp:
There is a bug in the implementation of  `set_low()` and `set_high()` functions in `OutputPin` for embedded-hal 1.0. The call to `set_low()` doesn't have any effect because `set_output_enable_override(OutputEnableOverride::Disable)` was set inside `new`

The fix aligns the implementation to be the same as used for embedded-hal 0.2 traits


jannic (approved):
Ouch - yes, of course!
LGTM. Thanks for fixing!

PR #802, opened 2024-05-04
Fix some clippy lints triggered by rust beta

jannic:

PR #811, opened 2024-06-29
gpio: add `PinGroup::set_u32` to allow setting each pin to a different state

ithinuel:


rp2040-hal/src/timer.rs
```diff
@@ -292,7 +292,7 @@ pub trait Alarm: Sealed {
     /// called, this will trigger interrupt whenever this timestamp is reached.
     ///
     /// The RP2040 is unable to schedule an event taking place in more than
-    /// `u32::max_value()` microseconds.
+    /// `u32::MAX` microseconds.
```
jannic, line 295: This is already covered by https://github.com/rp-rs/rp-hal/pull/802

ithinuel: Nice let's merge it then and I'll drop that commit 👍

jannic (approved):

PR #819, opened 2024-07-18
Install cargo-udeps and cargo-hack as binaries.

thejpster:
Speeds up the CI as then we don't have to wait for them to compile.


thejpster:
It is slightly slower than a cache hit, but much quicker than a cache miss (31s to install cargo-hack, 5m17 to install cargo-udeps).

9names (approved):
LGTM - I also hate long CI times.

PR #827, opened 2024-08-15
Remove code setting IC_DATA_CMD.RESTART bit

### LINKED ISSUE #825: I2C contract violation
The sensor I am writing a driver for requires I2C communication of the form
image
where `INDEX` is some internal 16 bit register and DATA can be any number of bytes.

The I2C transaction contract states:
- Data from adjacent operations of the same type are sent after each other without an SP or SR.
- Between adjacent operations of a different type an SR and SAD+R/W is sent.

Therefore I would expect this code to work
```rust
fn write(&mut self, mut reg: u16, data: &[u8]) -> Result<(), Error<I2C::Error>> {
    self.i2c.transaction(self.address, &mut [Operation::Write(&reg.to_be_bytes()), Operation::Write(data)])?;
}
```

Instead it appears an SP is sent between the two `Write` operations forcing unnecessary copies:
```rust
fn write(&mut self, mut reg: u16, data: &[u8]) -> Result<(), Error<I2C::Error>> {
    // self.i2c.transaction(self.address, &mut [Operation::Write(&reg.to_be_bytes()), Operation::Write(data)])?;

    let buf = &mut [0; 16];
    let data_len = buf.len() - 2;

    for chunk in data.chunks(data_len) {
        let chunk_buf = &mut buf[..chunk.len() + 2];

        let reg_bytes = reg.to_be_bytes();
        chunk_buf[0] = reg_bytes[0];
        chunk_buf[1] = reg_bytes[1];
        chunk_buf[2..].copy_from_slice(chunk);

        self.i2c.write(self.address, chunk_buf)?;

        reg += chunk.len() as u16;
    }

    Ok(())
}
```

Given that I need to send around 80kB of data this seems wasteful.

jannic:
According to the datasheet, a restart is automatically triggered "if the transfer direction is changing from the previous command". The RESTART bit is only needed to trigger an unconditional RESTART sequence.

According to the contract of I2c::transaction, "Data from adjacent operations of the same type are sent after each other without an SP or SR" and "Between adjacent operations of a different type an SR and SAD+R/W is sent". This looks like it is exactly what the hardware provides out of the box.

(See: https://docs.rs/embedded-hal/1.0.0/ebedded_hal/i2c/trait.I2c.html#tymethod.transaction)

This fixes #825. Unfortunately, it breaks some of the on-target tests, which expect the additional restart conditions.


jannic:
BTW, regarding the failing on-target tests, I already started to fix them.

ithinuel:
The change seem correct to me; And yes, updating the expected stop/restart of the on-target-test will be a bit tedious. It was a pain to figure in the first place because of how latency can cause several events to be merged/undicernable.

ithinuel (approved):

PR #828, opened 2024-08-16
Deletes the top-level workspace and moves the examples.

thejpster:
RP2350 stuff is coming, and you can't build rp2040-hal for Armv8-M and you can't build rp235x-hal for Armv6-M, and so they cannot be in the same workspace. Ergo, if the rp2350-hal is coming to this repo (and it is), we cannot have a workspace any more.

I also moved the examples into their own package, because that's much neater than having to deal with memory.x being in magically the right place, or a bunch of dev-dependencies that are not relevant to most crate users.


thejpster:
I'll rebase this

rp2040-hal/Cargo.toml
```diff
@@ -18,57 +18,35 @@ targets = ["thumbv6m-none-eabi"]
...
-rp2040-pac = { version = "0.6.0", features = ["critical-section"] }
 paste = "1.0"
 pio = "0.2.0"
+rand_core = "0.6.3"
 rp2040-hal-macros = { version = "0.1.0", path = "../rp2040-hal-macros" }
+rp2040-pac = { version = "0.6.0", features = ["critical-section"] }
+rtic-monotonic = { version = "1.0.0", optional = true }
 usb-device = "0.3"
 vcell = "0.1"
 void = { version = "1.0.2", default-features = false }
```
ithinuel, line 45: As we’re changing the deps order in the manifest, I think it’d be nice to have the optional dependencies separated from the mandatory ones to give them more visibility.

rp2040-hal-examples/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+futures = { version = "0.3.30", default-features = false, features = ["async-await"] }
+hd44780-driver = "0.4.0"
+nb = "1.0"
+nostd_async = { version = "0.6.1", features = ["cortex_m"] }
+panic-halt = "0.2.0"
+panic-probe = { version = "0.3.1", features = ["print-defmt"] }
+pio = "0.2.0"
+pio-proc = "0.2.0"
+rp2040-boot2 = "0.3.0"
+rp2040-hal = { path = "../rp2040-hal", version = "0.10.0", features = ["binary-info", "critical-section-impl", "rt", "defmt"] }
```
ithinuel, line 36: Although It was an eyesore, I think it was a good thing to see what features each example required.

thejpster: Feature unification. The HAL is only built once and needs all the features set that are required by any bin. With examples it would just silently not build any examples when the required feature wasn't set, which I found frustrating (`cargo build ` wouldn't build any examples, but `cargo build --all-features` would).

We could split each example into its own package, but that seems extreme.

Or I guess I could document each required feature at the top of each binary, but I've got no way of checking that that's correct.

rp2040-hal/Cargo.toml
thejpster: Done

rp2040-hal-examples/Cargo.toml
ithinuel: You are right. I found it nice that from a user experience, targeting a specific examples would tell "that examples requires those features".
But I’m fine with this as well.

Would it be good in the future to mutualize the examples and have a "portability"/"bsp" module in the example crate to map to either rp2040 or rp235x ?

thejpster: I think it's very unlikely that anyone *wouldn't* want the `critical-section-impl` or `rt` features. In particular, not using `rt` gives you a 'default' vector table which is broken and doesn't work. I'm also going to change all the examples so they include binary info by default, because it's just nice to have.

That leaves the 'defmt' feature. I could add a `defmt` feature to the examples package, and then mark the various binary crates within it that require the defmt feature in the HAL to require that feature to be set on the package.

ithinuel (approved):

PR #829, opened 2024-08-16
Rationalise all the license files and copyright notices.

thejpster:
I set the copyright to "Copyright (c) 2021-2024 The rp-rs Developers". This goes in the MIT license text, and the Apache 2.0 NOTICE file. The Apache 2.0 text has also been factory reset (the boilerplate had been incorrectly changed).

Every package in this repo must contain LICENSE-MIT, LICENSE-APACHE and NOTICE, plus the standard text from the README pointing to the two license files.

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -1,4 +1,5 @@
-Apache License
+
+                                 Apache License
                            Version 2.0, January 2004
                         http://www.apache.org/licenses/
 
@@ -186,7 +187,7 @@ Apache License
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2021 rp-rs organization
+   Copyright [yyyy] [name of copyright owner]
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/rp2040-hal-macros/README.md
+++ b/rp2040-hal-macros/README.md
@@ -10,18 +10,12 @@ Currently, it just unlocks all spinlocks before calling the entry function.
 
 # License
 
-Licensed under either of
+The contents of this repository are dual-licensed under the _MIT OR Apache 2.0_
+License. That means you can choose either the MIT license or the Apache 2.0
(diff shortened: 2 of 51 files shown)


rp2040-hal-macros/README.md
```diff
@@ -10,18 +10,12 @@ Currently, it just unlocks all spinlocks before calling the entry function.
...
-Unless you explicitly state otherwise, any contribution intentionally submitted
-for inclusion in the work by you, as defined in the Apache-2.0 license, shall be
-dual licensed as above, without any additional terms or conditions.
+The contents of this repository are dual-licensed under the _MIT OR Apache 2.0_
+License. That means you can choose either the MIT license or the Apache 2.0
+license when you re-use this code. See [`LICENSE-MIT`](./LICENSE-MIT) or
+[`LICENSE-APACHE`](./LICENSE-APACHE) for more information on each specific
+license. Our Apache 2.0 notices can be found in [`NOTICE`](./NOTICE).
+
+Any submissions to this project (e.g. as Pull Requests) must be made available
```
jannic, line 19: IANAL, but I think the old wording regarding contributions was better.

jannic:
Overall, LGTM. I wonder if the change of wording regarding contributions is significant, legally.

jannic (approved):

rp2040-hal-macros/README.md
thejpster: We had two kinds - do you mean the one with the `# Contributing` header? I can use that.

jannic (approved):

LICENSE-APACHE
```diff
@@ -186,7 +187,7 @@ Apache License
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2021 rp-rs organization
+   Copyright [yyyy] [name of copyright owner]
```
ithinuel, line 190: Shouldn’t this be kept as `2024 rp-rs organization` ? or maybe 2021.

ithinuel: never mind 🤦

PR #830, opened 2024-08-16
Extract picotool binary info to a separate `rp-binary-info` crate.

Dirbaio:
I'd like to reuse this from Embassy instead of copypasting
it, so as discussed in Matrix I'm opening a PR to etract it to a separate crate.

Having a separate crate causes a few changes to the public API. I think they're
nice on their own:

- Macros are now exported from the `binary_info` module, and no longer have a prefix: `rp2040_hal::binary_info::rp_program_name!`, vs previously `rp2040_hal::binary_info_rp_program_name!`.
- The `binary-info` feature makes the entire `rp2040_hal::binary_info` unavailable, instead of just not declaring the `static` (since the crate is not compiled at all).


thejpster:
All makes sense to me, but I would like to see a feature to control the static within the new crate - it might be useful to have the crate, but you want to create your own static to put it in a differently named linker section, for example.

thejpster:
Also, I've re-worded the README in the bit discussing licenseing, so please could you pick up the new wording? See https://github.com/rp-rs/rp-hal/blob/main/README.md#license

thejpster:
This fell over https://github.com/bluss/arrayvec/issues/273 - re-running with 0.7.6

rp2040-hal/src/lib.rs
```diff
@@ -52,7 +52,8 @@ pub mod adc;
 #[macro_use]
 pub mod async_utils;
 pub(crate) mod atomic_register_access;
-pub mod binary_info;
+#[cfg(feature = "binary-info")]
```
thejpster, line 55:
```suggestion

```

thejpster:
This looks great - thank you! I think the top-level module needs to always be included though (and props to udeps for catching that we had a dependency that was never used, which tipped me off)

PR #831, opened 2024-08-17
spi: move set_format to all states

ithinuel:
There are other drivers in the wild (uboot, zephyr) that set the format without disabling the peripheral, so we’ll assume it is fine to do so.

### DIFF
--- a/rp2040-hal/src/spi.rs
+++ b/rp2040-hal/src/spi.rs
@@ -278,7 +278,7 @@ impl<D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<Disabled, D, P, DS> {
     ///
     /// Valid pin sets are in the form of `(Tx, Sck)` or `(Tx, Rx, Sck)`
     ///
-    /// If you pins are dynamically identified (`Pin<DynPinId, _, _>`) they will first need to pass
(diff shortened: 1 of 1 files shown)


thejpster:
LGTM. Just needs a format.

rp2040-hal/src/spi.rs
```diff
@@ -287,6 +270,23 @@ impl<D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<Disabled, D, P, DS> {
...
     }
+}
+
+impl<D: SpiDevice, P: ValidSpiPinout<D>, const DS: u8> Spi<Disabled, D, P, DS> {
+    /// Create new not initialized Spi bus. Initialize it with [`.init`][Self::init]
+    /// or [`.init_slave`][Self::init_slave].
+    ///
+    /// Valid pin sets are in the form of `(Tx, Sck)` or `(Tx, Rx, Sck)`
+    ///
+    /// If you pins are dynamically identified (`Pin<DynPinId, _, _>`) they will first need to pass
```
thejpster, line 281:
```suggestion
    /// If your pins are dynamically identified (`Pin<DynPinId, _, _>`) they will first need to pass
```

thejpster (approved):

Issue #836, opened 2024-08-22
`embedded_io::ReadReady/WriteReady` not implemented for uart peripheral


antbern:
I am working on a project where I need to do some non-blocking reading from the uart peripheral based on the `embedded_io` traits. The regular [`Read`](https://docs.rs/embedded-io/0.6.1/embedded_io/trait.Read.html) trait specifies a _blocking_ read such that it will wait for at least one byte to be available to be read. If one wants to do a non-blocking read, you need to use the [`ReadReady`](https://docs.rs/embedded-io/0.6.1/embedded_io/trait.ReadReady.html) trait function `read_ready` to first check  if there is data available before doing the read.

However, only the `Read` trait is implemented (in #727 and #781 ) for the split and unsplit peripherals, but not the `ReadReady` trait. Is there a reason this has been left out?

It looks like the implementation could be quite simple, something like this for the `Reader` split peripheral (not tested):
```rust
impl<D: UartDevice, P: ValidUartPinout<D>> embedded_io::ReadReady for Reader<D, P> {
    fn read_ready(&mut self) -> Result<bool, Self::Error> {
        Ok(is_readable(self.device))
    }
}

```

ithinuel:
It might be simply because no one needed it before. Feel free to open a PR :)

PR #837, opened 2024-08-22
implement `embedded_io::ReadReady` and `WriteReady` traits for uart peripheral

antbern:
Implements  `embedded_io::ReadReady` and `WriteReady` traits for split and unsplit uart peripheral in terms of the already existing functions for checking if the uart is ready to read or write :rocket: 

Solves #836

### DIFF
--- a/rp2040-hal/src/uart/peripheral.rs
+++ b/rp2040-hal/src/uart/peripheral.rs
@@ -490,6 +490,15 @@ impl<D: UartDevice, P: ValidUartPinout<D>> embedded_io::Read for UartPeripheral<
         }
     }
 }
+
+impl<D: UartDevice, P: ValidUartPinout<D>> embedded_io::ReadReady
+    for UartPeripheral<Enabled, D, P>
+{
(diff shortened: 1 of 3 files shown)


jannic:
The PR itself looks good.

However while reviewing it I noticed that we probably have a bug in our implementation of `embedded_io::Write`:
```
    fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Error> {
        self.write_full_blocking(buf);
        Ok(buf.len())
    }
```
This will block if there's not enough FIFO space to hold the whole `buf`.

But that's not compatible with the definition of `WriteReady`: "If this returns true, it’s guaranteed that the next call to [Write::write](https://docs.rs/embedded-io/latest/embedded_io/trait.Write.html#tymethod.write) will not block."

As `WriteReady` can't know how many bytes the next call to `write` will contain, it can only give that guarantee if `write` returns without blocking as long as it can write _a single_ byte. So `write` must not be implemented in terms of `write_full_blocking`.

jannic (approved):

PR #838, opened 2024-08-22
Let UART write return some bytes were written

jannic:
embedded_io::Write::write(buf) should not block until the full buffer was written. Instead, it should only block until at least one byte was written.

(See https://github.com/rp-rs/rp-hal/pull/837#issuecomment-2305371353)


thejpster:
Fixes https://github.com/rp-rs/rp-hal/issues/840

thejpster (approved):
Note to self to also fix RP2350 HAL.

PR #841, opened 2024-08-24
Port UART updates to rp235x-hal

jannic:
The updates from #798, #837 and #838 were only applied to rp2040-hal. This patch ports them to rp235x-hal.

### DIFF
--- a/rp2040-hal-examples/Cargo.toml
+++ b/rp2040-hal-examples/Cargo.toml
@@ -22,6 +22,7 @@ dht-sensor = "0.2.1"
 embedded-alloc = "0.5.1"
 embedded-hal = "1.0.0"
 embedded-hal-async = "1.0.0"
+embedded-io = "0.6.1"
 embedded_hal_0_2 = {package = "embedded-hal", version = "0.2.5", features = ["unproven"]}
 fugit = "0.3.6"
 futures = {version = "0.3.30", default-features = false, features = ["async-await"]}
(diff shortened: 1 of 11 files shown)


thejpster:
What testing has been done? Should I run a demo and try it?

jannic:
Only tested that it still builds. It seems like we don't have on-target tests for UART yet.
I'd consider the port to rp235x relatively low-risk as the changes don't touch any low-level details. But of course, doing some actual testing would be very welcome!

thejpster:
I think we should write a loop back test that sends bytes with good and bad parity, just to check this is working as expected.

thejpster:
I wrote a UART loopback program, and fixed some issues I observed: https://github.com/rp-rs/rp-hal/commit/1f6251a418cb8b105a39e3e491352d3be1073cb5

You can cherry-pick that here if you like it.

jannic:
> I wrote a UART loopback program, and fixed some issues I observed: [1f6251a](https://github.com/rp-rs/rp-hal/commit/1f6251a418cb8b105a39e3e491352d3be1073cb5)
> 
> You can cherry-pick that here if you like it.

Thanks!
Did you use external pull-ups? The uart_loopback example didn't work for me. Took me a while to find out what's going on because without defmt I didn't see the unwrap failing. The fix was enabling the internal pull-ups on the RX pins.
(Only tested on rp2040 yet)

thejpster:
No I didn't. Just a DuPont wire on a Pimoroni Pico 2 Plus.

jannic:
Interesting. I reproducibly got a break error on rp2040 before I activated the pull-ups.

thejpster:
🤷

9names (approved):

PR #842, opened 2024-08-24
Port SPI changes from rp2040-hal to rp235x-hal

jannic:
The updates from #831 were only applied to rp2040-hal.
This patch ports them to rp235x-hal.


thejpster:
What testing has been done? Should I run a demo and try it?

jannic:
Only tested that it still builds.

The actual change is that it's now possible to call `set_format` without first disabling the peripheral. As I understand it from #831, the datasheet could be interpreted in a way that forbids this, but other implementations "in the wild" already do it so it's supposed to be fine. The wording is exactly identical in the RP235x datasheet.

thejpster (approved):

PR #845, opened 2024-08-27
Provide arch module on RP2040

jannic:
This improves code portability between rp2040-hal and rp235x-hal


thejpster:
Just needs a format

PR #847, opened 2024-09-01
Interrupts on RISC-V

thejpster:
) Adds an Xh3irq controller driver
) Adds a default `MachineExternal` handler which uses the Xh3irq to run whichever interrupts are pending
) Adds a bunch of interrupt control stuff to `hal::arch`
) Fixes the `powman_test` example to use the new interrupt APIs - it now works on RISC-V and Arm.

Tested `powman_test` on a Pico 2 in both `riscv32imac-unknown-none-elf` and `thumbv8m.main-none-eabihf`.

### DIFF
--- a/rp235x-hal-examples/src/bin/gpio_irq_example.rs
+++ b/rp235x-hal-examples/src/bin/gpio_irq_example.rs
@@ -29,19 +29,13 @@ use panic_halt as _;
 // Alias for our HAL crate
 use rp235x_hal as hal;
 
-// Some things we need
+// Some traits we need
 use embedded_hal::digital::StatefulOutputPin;
 
-// Our interrupt macro
-use hal::pac::interrupt;
-
-// Some short-cuts to useful types
+// Some more helpful aliases
 use core::cell::RefCell;
 use critical_section::Mutex;
 use hal::gpio;
-
-// The GPIO interrupt type we're going to generate
-use gpio::Interrupt::EdgeLow;
 
 /// Tell the Boot ROM about our application
 #[link_section = ".start_block"]
@@ -70,7 +64,7 @@ type LedAndButton = (LedPin, ButtonPin);
 /// This how we transfer our Led and Button pins into the Interrupt Handler.
 /// We'll have the option hold both using the LedAndButton type.
 /// This will make it a bit easier to unpack them later.
-static GLOBAL_PINS: Mutex<RefCell<Option<LedAndButton>>> = Mutex::new(RefCell::new(None));
+static GLOBAL_STATE: Mutex<RefCell<Option<LedAndButton>>> = Mutex::new(RefCell::new(None));
 
 /// Entry point to our bare-metal application.
 ///
@@ -120,20 +114,25 @@ fn main() -> ! {
 
     // Trigger on the 'falling edge' of the input pin.
     // This will happen as the button is being pressed
-    in_pin.set_interrupt_enabled(EdgeLow, true);
+    in_pin.set_interrupt_enabled(gpio::Interrupt::EdgeLow, true);
 
     // Give away our pins by moving them into the `GLOBAL_PINS` variable.
     // We won't need to access them in the main thread again
     critical_section::with(|cs| {
-        GLOBAL_PINS.borrow(cs).replace(Some((led, in_pin)));
+        GLOBAL_STATE.borrow(cs).replace(Some((led, in_pin)));
     });
 
-    // Unmask the IO_BANK0 IRQ so that the NVIC interrupt controller
-    // will jump to the interrupt function when the interrupt occurs.
-    // We do this last so that the interrupt can't go off while
-    // it is in the middle of being configured
+    // Unmask the IRQ for I/O Bank 0 so that the RP2350's interrupt controller
+    // (NVIC in Arm mode, or Xh3irq in RISC-V mode) will jump to the interrupt
+    // function when the interrupt occurs. We do this last so that the interrupt
+    // can't go off while it is in the middle of being configured
     unsafe {
-        cortex_m::peripheral::NVIC::unmask(hal::pac::Interrupt::IO_IRQ_BANK0);
+        hal::arch::interrupt_unmask(hal::pac::Interrupt::IO_IRQ_BANK0);
+    }
+
+    // Enable interrupts on this core
+    unsafe {
+        hal::arch::interrupt_enable();
     }
 
     loop {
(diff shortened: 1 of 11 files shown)


rp235x-hal-examples/src/bin/i2c_async.rs
```diff
@@ -98,8 +98,8 @@ async fn demo() {
     // Each core has its own NVIC so these needs to executed from the core where the IRQ are
```
9names, line 98: Not really an NVIC here, is it?

rp235x-hal/src/xh3irq.rs
```diff
@@ -0,0 +1,239 @@
...
+        assert_eq!(index, 2);
+        assert_eq!(bitmask, 0b0000_0000_0000_0100);
+    }
+
+    #[test]
+    fn test_interrupt_to_mask_index() {
+        let mask = interrupt_to_mask_index(rp235x_pac::Interrupt::UART1_IRQ);
+        // 34 -> bit 3 | window 2
+        assert_eq!(mask, 0x0004_0002);
+    }
```
9names, line 236: what does the 34 comment mean here?

thejpster: Oh, UART1_IRQ has the value 34

thejpster:
We might want to block this until we've written a new interrupt macro that works like the cortex-m-rt macro and does the static mut transform and type-checks the function signature.

rp235x-hal-examples/src/bin/i2c_async.rs
thejpster: Cleaned up the comment

rp235x-hal/src/xh3irq.rs
thejpster: Cleaned up the test and used another IRQ where the bit number and the index have different numeric values

rp235x-hal-examples/Cargo.toml
```diff
@@ -28,10 +28,15 @@ futures = {version = "0.3.30", default-features = false, features = ["async-awai
...
 pio-proc = "0.2.0"
 rp235x-hal = {path = "../rp235x-hal", version = "0.2.0", features = ["binary-info", "critical-section-impl", "rt", "defmt"]}
 usb-device = "0.3.2"
 usbd-serial = "0.2.2"
+
+[target.'thumbv8m.main-none-eabihf'.dependencies]
+nostd_async = {version = "0.6.1", features = ["cortex_m"]}
+
+[target.riscv32imac-unknown-none-elf.dependencies]
+nostd_async = {version = "0.6.1"}
```
jannic, line 42: I don't think `nostd_async` should be used without the `cortex_m` feature. The fallback implementation of `interrupt::free` just doesn't disable any interrupts and unsafely creates a `CriticalSection` token anyway:

https://github.com/sammhicks/nostd_async/blob/master/src/interrupt.rs#L5-L10

thejpster: Yeah. Not sure what we should do instead though.

jannic: I'd say either embassy-executor (AFAIKT the most popular no_std executor), or some minimal toy executor like the one I suggested for on-target tests in #848.

There are two conflicting goals when choosing the executor to be used for the examples:
- Examples should be concise
- Examples should introduce knowledge that's applicable to real-world tasks

Depending on what you value more, you could justify both choices.
In any case, nostd_async IMHO isn't a good choice, as it may look like it's usable for real (non-example) code, while in fact it has severe issues.

I'm not knowledgeable enough in writing async executors to decide if a third option could be viable: Improving nostd_async. It may only need a few small fixes, or it may be fundamentally flawed - I just don't know.

rp235x-hal/src/arch.rs
```diff
@@ -73,43 +101,476 @@ mod inner {
...
+    /// Just panics.
+    ///
+    /// # Safety
+    ///
+    /// Do not call this function - it is called automatically when our
+    /// `MachineExternal` interrupt handler can't find anything better to call.
+    #[no_mangle]
+    #[allow(non_snake_case)]
+    unsafe fn DefaultIrqHandler() {
+        panic!();
```
jannic, line 485: cortex_m does not panic but `loop {}` in the default handler.
It did panic for a short while in the past, but that was reverted here: https://github.com/rust-embedded/cortex-m-rt/pull/289/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759L993

I'm not sure if the reasons for that decision also apply for RISC-V, but at least for the rp2040/rp235x case it would be nicer if both implementations did the same.

jannic:
> We might want to block this until we've written a new interrupt macro that works like the cortex-m-rt macro and does the static mut transform and type-checks the function signature.

As the rp235x is a multi-core processor, and given that https://github.com/rust-embedded/cortex-m/issues/411 indicates that the transformation made by cortex-m-rt is unsound on multi-core systems, I don't think we should just copy it. Is there a good way to make the transform safe on rp235x? Taking a full critical section on each interrupt entry is probably to expensive?

rp235x-hal-examples/src/bin/gpio_irq_example.rs
```diff
@@ -142,21 +144,23 @@ fn main() -> ! {
     }
 }
 
-#[interrupt]
+#[no_mangle]
+#[allow(non_snake_case)]
 fn IO_IRQ_BANK0() {
     // The `#[interrupt]` attribute covertly converts this to `&'static mut Option<LedAndButton>`
     static mut LED_AND_BUTTON: Option<LedAndButton> = None;
+    let led_and_button = unsafe { &mut *core::ptr::addr_of_mut!(LED_AND_BUTTON) };
```
jannic, line 152: This `&mut *core::ptr::addr_of_mut!( ... )` IMHO doesn't make sense.

The whole raison d'être of `addr_of_mut!` is avoiding the `&mut`, because having a `&mut` (even for a short time, and not derefencing it) can potentially cause UB if you are not careful.

But here, you convert it to a `&mut` anyway. So I don't see any advantage of this difficult to read incantation, you can as well just do `unsafe { &mut LED_AND_BUTTON };` instead.

Now, that causes a warning. And this warning looks like it led you to writing the code above:

```
warning: creating a mutable reference to mutable static is discouraged
   --> src/bin/gpio_irq_example.rs:152:35
    |
152 |     let led_and_button = unsafe { &mut LED_AND_BUTTON };
    |                                   ^^^^^^^^^^^^^^^^^^^ mutable reference to mutable static
    |
    = note: for more information, see issue #114447 <https://github.com/rust-lang/rust/issues/114447>
    = note: this will be a hard error in the 2024 edition
    = note: this mutable reference has lifetime `'static`, but if the static gets accessed (read or written) by any other means, or any other reference is created, then any further use of this mutable reference is Undefined Behavior
    = note: `#[warn(static_mut_refs)]` on by default
help: use `addr_of_mut!` instead to create a raw pointer
    |
152 |     let led_and_button = unsafe { addr_of_mut!(LED_AND_BUTTON) };
    |                                   ~~~~~~~~~~~~~              +
```

I think the warning is valid, but the `help` is misleading. It even says "to create a raw pointer", and as written it would return a pointer, not a `&mut`, so the resulting code wouldn't build.

TL;DR:

```suggestion
    #[allow(static_mut_refs)]
    // Safety: As long as this interrupt is only enabled on a single core,
    // and the interrupt isn't somehow called reentrantly, there will not
    // be a second reference to LED_AND_BUTTON at the same time.
    let led_and_button = unsafe { &mut LED_AND_BUTTON };
```

thejpster:
So I guess the risk is that both CPUs enter the same interrupt handler at the same time. If one peripheral interrupt signal is unmasked on both CPUs, that could happen. So perhaps the fix is to design the interrupt API to make that impossible (like a bitmask in a static atomic or the watchdog scratch that records which cpu has which interrupt enabled). Or we could ensure that if you enter an interrupt on Core A, it is masked for the duration on Core B - which is nicer than turning all interrupts off or holding a spin lock.

But given 2040 has the same problem the question is whether we want to fix it here or make a plan and fix it later.

9names:
The user already unmasks the interrupt, and that operation is unsafe. It is almost certainly a bug if they do so, and we have had no reports of people doing this accidentally.

Until someone comes has a use-case where having both cores handle an interrupt is beneficial, I think trying to solve this is not a good use of our time.

My opinion is we should document "don't do this" and move on with our lives

jannic:
> But given 2040 has the same problem the question is whether we want to fix it here or make a plan and fix it later.

That's an option, I just wouldn't want to spend time on porting the macro only to remove it again later.

> Until someone comes has a use-case where having both cores handle an interrupt is beneficial,

Are there situations where it's not avoidable? Any non maskable interrupts or exceptions?

jannic:
But to be honest, there's a second thing I dislike about the interrupt macro, so I'm biased.

For me, it's totally non-obvious that the macro annotation changes the code inside the function.
So if I see this:
```
#[interrupt]
fn SOME_INTERRUPT() {
  static mut SOMETHING: Option<Whatever> = None;
  [...]
  *SOMETHING = foo(); 
}
```
I always feel like the usage of the variable doesn't match the declaration. This was very confusing when I was new to embedded rust, and it still sometimes surprises me.

Therefore I'd prefer something more obvious. Perhaps a macro directly on the static mut?
```
#[interrupt]
fn SOME_INTERRUPT() {
  let something = interrupt_singleton!( didn't think about what comes here yet :Option<Whatever> );
  [...]
  *something = foo(); 
}
```

Probably needs an `unsafe` declaration as well, as using this macro outside an interrupt (or in a re-entrant interrupt) would be unsound. So it's far from a finished proposal, of course. This is only meant as an explanation why I don't like the current interrupt macro.

Back to topic: Personally I'd merge the PR as it is, and not wait for an `#[interrupt]` macro implementation for RISC-V. The current situation is not perfect, but it's good enough, and it's not clear yet how an improvement would look like.

(That said, if someone ported `#[interrupt]` I wouldn't oppose that either. While I still don't like it, I do value consistency.)

jannic:
> So I guess the risk is that both CPUs enter the same interrupt handler at the same time. If one peripheral interrupt signal is unmasked on both CPUs, that could happen.

BTW, do I understand `xh3irq` correctly that it already solves this issue? `xh3irq::get_next_interrupt()` should only return a given interrupt to a single CPU core, even if the interrupt is enabled on both, right?

rp235x-hal-examples/Cargo.toml
jannic: https://github.com/rp-rs/rp-hal/pull/850

thejpster:
> > So I guess the risk is that both CPUs enter the same interrupt handler at the same time. If one peripheral interrupt signal is unmasked on both CPUs, that could happen.
> 
> 
> 
> BTW, do I understand `xh3irq` correctly that it already solves this issue? `xh3irq::get_next_interrupt()` should only return a given interrupt to a single CPU core, even if the interrupt is enabled on both, right?

I think the UART interrupt is cleared by reading the FIFO which can happen before the ISR ends. So it could re-fire before the ISR ends. Not sure either Interrupt Controller can avoid that.

thejpster:
What do we want to do with this? There's been some changes over in riscv-rt since.

winksaville:
> What do we want to do with this? There's been some changes over in riscv-rt since.

I'd like to see interrupts working on risc-v and willing to work on this if you'd like. That
said, I'd need some initial guidance as I'm a noob so it will definitely take me some time.

rp235x-hal/src/arch.rs
thejpster: I'll fix this

rp235x-hal-examples/src/bin/gpio_irq_example.rs
thejpster: This will be a hard error in 2024 Edition, so I'll think of a better solution.

jonathanpallant:
I rebased and took out any use of the unsound static-mut conversion. Now the interrupts just run in a critical section. Not yet tested on hardware.

rp235x-hal/src/arch.rs
jonathanpallant: Changed it to spin

thejpster:
What do we need in order to merge this?

9names:
> Not yet tested on hardware.

Might just be me, but some proof (even anecdotal) that it works on the hardware it's supposed to target would be nice before merging it.

9names (approved):
I tested that both the gpio_irq_example and powman_test examples worked on both riscv32imac-unknown-none-elf and thumbv8m.main-none-eabihf on a Pico 2.

LGTM

thejpster:
Thank you for checking on hardware. Much appreciated. My Pico 2 is under a pile of other stuff at the moment, sadly.

PR #848, opened 2024-09-08
Replace nostd_async with a simple test executor

jannic:
This has some advantages:

- no longer depend on nostd_async, which has known issues (https://github.com/sammhicks/nostd_async/issues/4) and looks rather unmaintained
- the custom executor can check that the waker is only called when expected (currently only implemented as some defmt logging, but this can be improved)
- in case of problems, it is easier to debug than a full-fledged executor

The new dependency on `once_cell` can be removed with rust 1.82.0, which makes `Waker::from_raw` callable from `const` contexts.


ithinuel (approved):

jannic:
@ithinuel Thanks for the approval, it's encouraging that you agree with the approach. I may push some code cleanups to this pull request later today.

sammhicks:
Hi, I'm the developer of `nostd_async`.

There was definitely UB in `nostd_async`, which should now have been resolved.

Miri complains about version `0.6.1`, but not `0.7.0` to `0.7.2`.

However, I'm no longer developing it, so using a different executor is wise.

PR #849, opened 2024-09-09
Disable several warnings that show up with rust 1.82.0(beta)

jannic:
For most, I didn't find a good solution that would work with both rust 1.81.0 and 1.82.0.

clippy::too_long_first_doc_paragraph is an exception, this could be solved by rephrasing the doc comments. However I didn't want to edit all the doc comments for the rom functions.


thejpster:
Ugh, it's a shame this was necessary. Your fix looks OK.

thejpster (approved):

jannic:
This may still change: https://github.com/rust-lang/rust/issues/129352#issuecomment-2339084913

PR #850, opened 2024-09-14
Use embassy-executor instead of nostd_async

jannic:
For the needs of our examples, a much simpler executor would suffice. But it doesn't hurt to use a full-blown executor, the code doesn't become more complicated that way.

### DIFF
--- a/rp2040-hal-examples/Cargo.toml
+++ b/rp2040-hal-examples/Cargo.toml
@@ -41,4 +41,4 @@ rp2040-hal = {path = "../rp2040-hal", version = "0.10.0", features = ["binary-in
 static_cell = "2.1.0"
 
 [target.'cfg( target_arch = "arm" )'.dependencies]
-embassy-executor = {version = "0.6", features = ["arch-cortex-m", "executor-thread"]}
+embassy-executor = {version = "0.5", features = ["arch-cortex-m", "executor-thread"]}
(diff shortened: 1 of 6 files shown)


thejpster:
I'm happy with this but I'd like to see the use declarations kept in some kind of order. Probably:

* core imports
* Hal alias
* Hal imports
* imports from other crates

jannic:
Yes, sorry, I didn't care about the placement but just left them where rust-analyzer put them.

thejpster (approved):

ithinuel:
I'm still concerned about the confusion between rp2*-hal and embassy's own embassy-rp. We already regularly get questions from people trying to mix the two. I'm afraid this would only increase this issue.

@jamesmunns told me about  https://mycelium.elizas.website/maitake/ a few months back and I think it'd be a good alternative candidate.
I haven't had the time to look into it though.

jannic:
> I'm still concerned about the confusion between rp2*-hal and embassy's own embassy-rp. We already regularly get questions from people trying to mix the two. I'm afraid this would only increase this issue.

rp2*-hal + embassy-executor looks like a very good combination, and I don't think we should discourage that because of possible confusion. We should however explain it in our documentation. Both that it's only one possible executor and there are others, and that rp2*-hal and embassy-rp are alternatives and should not be used at the same time.

jannic:
@ithinuel, I had a quick look at maitake. It's definitely interesting, but it doesn't seem to provide a ready-made executor we could easily use in our examples. Therefore I'd still propose to go with embassy-executor, for now. That does not rule out a later switch to an alternative. Especially as the HAL itself doesn't depend on the executor at all. Only the examples need it.

thejpster:
I'm fine merging this and changing it later if we need to.

Another option is lilos but it currently doesn't support RISC-V.

jannic:
Yes, let's not get paralyzed by too many choices. Merging this now, feel free to propose a better alternative any time.

PR #855, opened 2024-09-27
Prepare a 0.0.1 release of the rp235x crates

jannic:
I'm not sure if the crates are well-tested enough for a "real" release, but starting with a 0.0.1 for now wouldn't hurt IMHO.


thejpster:
It might be best to not run the version clock backwards in case people are using git deps. 0.2.0 is still "pretty sketchy" as a version number.

Issue #853, opened 2024-09-25
rp-binary-info is not compatible with picotool for rp2350


romainreignier:
With the latest version (2.0.0) of picotool, `picotool info` works with rp2040 binaries but not with rp2350.

```sh
$ picotool version
picotool v2.0.0 (Linux, GNU-11.4.0, Release)
```

## Steps to reproduce

```sh
$ git clone https://github.com/rp-rs/rp-hal.git
$ cd rp2040-hal-examples
$ cargo build --bin binary_info_demo
$ picotool info -a -t elf target/thumbv6m-none-eabi/debug/binary_info_demo
File target/thumbv6m-none-eabi/debug/binary_info_demo:

Program Information
 name:          rp2040-hal Binary Info Example
 version:       0.1.0
 web site:      https://github.com/rp-rs/rp-hal
 description:   A GPIO blinky with extra metadata.
 binary start:  0x10000000

Fixed Pin Information
 none

Build Information
 pico_board:        pico
 build attributes:  debug
```
 So it works with RP2040.

But doing the same RP2350 provides no info:

```sh
$ cd rp235x-hal-examples
$ cargo build --bin binary_info_demo
$ picotool info -a -t elf target/thumbv8m.main-none-eabihf/debug/binary_info_demo
File target/thumbv8m.main-none-eabihf/debug/binary_info_demo:

Program Information
 none

Fixed Pin Information
 none

Build Information
 none
```

jonathanpallant:
Can you attach your two elf files, or can you run `llvm-objdump -t | rustfilt | sort` on them?

romainreignier:
Sure.

Objdump results:
[rp2040.txt](https://github.com/user-attachments/files/17132593/rp2040.txt)
[rp2350.txt](https://github.com/user-attachments/files/17132596/rp2350.txt)

Elf files with a `.txt` extension to allow upload on Github:
[rp2040_binary_info_demo_bin.txt](https://github.com/user-attachments/files/17132671/rp2040_binary_info_demo_bin.txt)
[rp2350_binary_info_demo_bin.txt](https://github.com/user-attachments/files/17132666/rp2350_binary_info_demo_bin.txt)

jonathanpallant:
RP2020:

```text
100001c0 g     O .boot_info	00000014 rp_binary_info::PICOTOOL_HEADER
100085a4 l     O .bi_entries	0000001c binary_info_demo::PICOTOOL_ENTRIES
[...]

PR #857, opened 2024-09-29
Run picotool step in github action

jannic:
This should catch bugs like https://github.com/rp-rs/rp-hal/issues/853

### DIFF
--- a/.github/workflows/rp2040_hal_examples.yml
+++ b/.github/workflows/rp2040_hal_examples.yml
@@ -14,7 +14,11 @@ jobs:
       - name: Build
         run: cd ${PACKAGE} && cargo build
       - name: Install picotool
-        run: sudo sh -c "apt-get install -y debian-archive-keyring && echo 'deb [signed-by=/usr/share/keyrings/debian-archive-keyring.gpg] http://deb.debian.org/debian testing main contrib non-free' >/etc/apt/sources.list.d/debian.list && apt update && apt-get install -y libxml2-utils picotool"
+        run: |
+          sudo apt-get install -y debian-archive-keyring
+          sudo sh -c "echo 'deb [signed-by=/usr/share/keyrings/debian-archive-keyring.gpg] http://deb.debian.org/debian testing main contrib non-free' >/etc/apt/sources.list.d/debian.list"
+          sudo apt-get update
+          sudo apt-get install -y libxml2-utils picotool
       - name: Test picotool
         run: picotool info ${PACKAGE}/target/${TARGET}/debug/binary_info_demo -t elf -a | tee /dev/stderr | grep -q "rp2040-hal Binary Info Example"
   udeps:
--- a/.github/workflows/rp2040_hal_examples.yml
+++ b/.github/workflows/rp2040_hal_examples.yml
@@ -5,7 +5,7 @@ env:
   TARGET: thumbv6m-none-eabi
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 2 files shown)


.github/workflows/rp2040_hal_examples.yml
```diff
@@ -5,14 +5,18 @@ env:
...
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
         with:
           target: ${{ env.TARGET }}
       - name: Build
         run: cd ${PACKAGE} && cargo build
+      - name: Install picotool
+        run: sudo sh -c "apt-get install -y debian-archive-keyring && echo 'deb [signed-by=/usr/share/keyrings/debian-archive-keyring.gpg] http://deb.debian.org/debian testing main contrib non-free' >/etc/apt/sources.list.d/debian.list && apt update && apt-get install -y libxml2-utils picotool"
```
thejpster, line 17: can this be a multi-line script? the one-liner is hard to read. Also the mix of apt-get and apt is weird.

```suggestion
        run:
          sudo apt-get update
          sudo apt-get install -y debian-archive-keyring libxml2-utils
          echo 'deb [signed-by=/usr/share/keyrings/debian-archive-keyring.gpg] http://deb.debian.org/debian testing main contrib non-free' | sudo tee /etc/apt/sources.list.d/debian.list
          sudo apt-get update
          sudo apt-get install -y picotool"
```

.github/workflows/rp2040_hal_examples.yml
```diff
@@ -5,14 +5,18 @@ env:
...
+      - name: Test picotool
+        run: picotool info ${PACKAGE}/target/${TARGET}/debug/binary_info_demo -t elf -a | tee /dev/stderr | grep -q "rp2040-hal Binary Info Example"
```
thejpster, line 19: What does tee'ing to /dev/stderr do here?

.github/workflows/rp235x_hal_examples_arm.yml
```diff
@@ -5,14 +5,18 @@ env:
...
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
         with:
           target: ${{ env.TARGET }}
       - name: Build
         run: cd ${PACKAGE} && cargo build --target=${TARGET}
+      - name: Install picotool
+        run: sudo sh -c "apt-get install -y debian-archive-keyring && echo 'deb [signed-by=/usr/share/keyrings/debian-archive-keyring.gpg] http://deb.debian.org/debian testing main contrib non-free' >/etc/apt/sources.list.d/debian.list && apt update && apt-get install -y libxml2-utils picotool"
```
thejpster, line 17: Same comment as above.

.github/workflows/rp2040_hal_examples.yml
jannic: `grep` consumes stdout and checks for some expected string - that's what makes the test fail or pass. But then, I also want to see the output of `picotool`, if only to understand the error cause if `grep` doesn't find the expected string. So `tee /dev/stderr` copies the output of `picotool` to stderr where it won't be affected by the following `grep`.

.github/workflows/rp2040_hal_examples.yml
jannic: BTW if you know a better way to install a current version of picotool, I'm all ears. Installing a package from debian testing on ubuntu may be a bit unreliable.

thejpster: https://github.com/raspberrypi/pico-sdk-tools/releases/tag/v2.0.0-5

Grab the tarball and unpack it?

thejpster (approved):

.github/workflows/rp2040_hal_examples.yml
jannic: Good suggestion, I didn't know that they provide binaries. https://github.com/raspberrypi/picotool doesn't mention it in the readme, and the releases page of that repo only has a source tarball.

As the pipeline works as it is, I'll merge this PR, but I consider simplifying the pipeline when I find some time to do it.

PR #859, opened 2024-10-05
Implement InputPin and OutputPin for Pin<I, DynFunction, P>

jannic:
Operations on this are fallible: They return Err(DynPinError::IncompatibleFunction) if the pin is not configured to an incompatible function.


jannic:
What other traits could sensibly be implemented for DynFunction?

thejpster:
I'd just add these for now and add more in another PR. When doing the #920 changes, this impl would have been useful! Without it, dyn pins aren't very useful.

PR #861, opened 2024-10-17
Add debug derive for the GPIO structures

robamu:
Is there a general problem adding these derives to more structures? I am trying to use a driver for a LCD display, and I can not unwrap the returned result structure because Debug is not implemented by the pin structures used by the driver. This applies to both the concrete pins and the downgraded pins.

These were only the derives I needed to pass both dyn pins and concrete pins to the driver. There are probably a lot more structures where t his derive could be added.

### DIFF
--- a/on-target-tests/Cargo.toml
+++ b/on-target-tests/Cargo.toml
@@ -48,7 +48,7 @@ name = "gpio"
 [dependencies]
 cortex-m = "0.7"
 cortex-m-rt = "0.7"
-critical-section = "1.0.0"
+critical-section = "1.2.0"
 defmt = "0.3"
 defmt-rtt = "0.4"
 defmt-test = "0.3.1"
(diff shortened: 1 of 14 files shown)


jannic:
CI failed due to a new warning in rustc 1.82.0.
Fixed in https://github.com/rp-rs/rp-hal/pull/863.

thejpster:
@robamu can you rebase?

robamu:
I can do that

thejpster:
This PR picked up a lot of unrelated commits. Maybe we can fix it by doing a rebase when we merge. The diff itself looks fine.

jannic:
I think the pull request looks a bit strange because @robamu merged with main instead of rebasing. Git should handle that fine. Worst case the history will have a few unnecessary branch points, but yes, "rebase and merge" should avoid even that.

jannic (approved):

PR #865, opened 2024-10-17
Fix or surpress several warnings on rust beta

jannic:


rp2040-hal-examples/Cargo.toml
```diff
@@ -42,3 +42,6 @@ static_cell = "2.1.0"
 
 [target.'cfg( target_arch = "arm" )'.dependencies]
 embassy-executor = {version = "0.5", features = ["arch-cortex-m", "executor-thread"]}
+
+[lints.clippy]
+too_long_first_doc_paragraph = "allow"
```
thejpster, line 47: I'd much rather fix these I think...

thejpster:
Excellent work here. I thought I was going to be annoyed but I think the compiler is making us make our source code better.

jannic:
The `static mut LED_AND_ALARM: Mutex<...>` was embarrassing. Unnecessary `mut` making all uses unsafe. And nobody noticed without the compiler warning.

rp2040-hal-examples/Cargo.toml
jannic: Sure! I didn't want to spend too much time on these right now, as I think they are rather unimportant. Note that I only disabled that lint for the example crates. And I doubt many people will look at the rustdocs for the examples anyway.

In fact, I wonder if most of the doc comments the lint complains about should be changed to normal (non-doc) comments. Eg.:

```
/// The linker will place this boot block at the start of our program image. We
/// need this to help the ROM bootloader get our code up and running.
/// Note: This boot block is not necessary when using a rp-hal based BSP
/// as the BSPs already perform this step.
#[link_section = ".boot2"]
#[used]
pub static BOOT2: [u8; 256] = rp2040_boot2::BOOT_LOADER_GENERIC_03H;
```

That comment explains why BOOT2 is there, which is of course useful in the source code. But does it belong into the documentation?

thejpster: Fair

thejpster (approved):

PR #873, opened 2024-11-13
Implement support for XipCS1 pin function.

Altaflux:
Implement support for XipCS1 pin function which is necessary to implement support for external ram or flash.

Note: The rp235x-pac is missing function select definition for XipCS1:
https://github.com/rp-rs/rp235x-pac/blob/main/svd/RP2350.yaml#L321

But `FUNCSEL_A::GPCK` has the same value (9). We should add a variant specific for XipCS1.

### DIFF
--- a/rp235x-hal/src/gpio/func.rs
+++ b/rp235x-hal/src/gpio/func.rs
@@ -48,6 +48,8 @@ pub enum DynFunction {
     /// The 'Clock' function. This can be used to input or output clock references to or from the
     /// rp235x.
     Clock,
+    /// The XIP CS1 function.
+    XipCs1,
     /// The 'USB' function. Only VBUS detect, VBUS enable and overcurrent detect are configurable.
     /// Other USB io have dedicated pins.
     Usb,
@@ -105,7 +107,7 @@ macro_rules! pin_func {
         })*
     };
 }
-pin_func!(Xip, Spi, Uart, I2c as I2C, Pwm, Pio0, Pio1, Pio2, Clock, Usb, UartAux, Null);
+pin_func!(Xip, Spi, Uart, I2c as I2C, Pwm, Pio0, Pio1, Pio2, Clock, XipCs1, Usb, UartAux, Null);
 
 //==============================================================================
 // SIO sub-types
@@ -169,6 +171,7 @@ impl DynFunction {
         let dyn_pin = id.as_dyn();
         match (self, dyn_pin.bank, dyn_pin.num) {
             (Xip, Bank0, _) => false,
+            (XipCs1, Bank0, 0 | 8 | 19 | 47) => true,
             (Clock, _, 0..=19 | 26..=29) => false,
             (UartAux, _, n) if (n & 0x02) == 0 => false,
             (_, Bank0, 0..=29) => true,
@@ -205,6 +208,7 @@ pin_valid_func!(
     [2, 3, 6, 7, 10, 11, 14, 15, 18, 19, 22, 23, 26, 27]
 );
 pin_valid_func!(bank0 as Gpio, [Clock], [20, 21, 22, 23, 24, 25]);
+pin_valid_func!(bank0 as Gpio, [XipCs1], [0, 8, 19, 47]);
 pin_valid_func!(qspi as Qspi, [Xip, Null], [Sclk, Sd0, Sd1, Sd2, Sd3, Ss]);
 
 macro_rules! pin_valid_func_sio {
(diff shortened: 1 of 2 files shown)


rp235x-hal/src/gpio/pin.rs
```diff
@@ -168,6 +168,7 @@ pub(crate) fn set_function<P: PinId>(pin: &P, function: DynFunction) {
         DynFunction::Pio1 => FUNCSEL_A::PIO1,
         DynFunction::Pio2 => FUNCSEL_A::PIO2,
         DynFunction::Clock => FUNCSEL_A::GPCK,
+        DynFunction::XipCs1 => FUNCSEL_A::GPCK,
```
Altaflux, line 171: The rp235x-pac is missing function select definition for XipCS1:
https://github.com/rp-rs/rp235x-pac/blob/main/svd/RP2350.yaml#L321

But FUNCSEL_A::GPCK has the same value (9). We should add a variant specific for XipCS1.

thejpster: OK I'll go fix turf I'll add a ticket for that

jonathanpallant: https://github.com/rp-rs/rp235x-pac/issues/7

It's not an easy fix, sadly.

thejpster: We should probably limit which pins you can set with ::Clock and ::XipCs1, to avoid surprises.

rp235x-hal/src/gpio/func.rs
```diff
@@ -48,6 +48,8 @@ pub enum DynFunction {
     /// The 'Clock' function. This can be used to input or output clock references to or from the
     /// rp235x.
     Clock,
+    /// The XIP CS1 function.
+    XipCs1,
```
thejpster, line 52: Maybe this is the same DynFunction as Clock? They have the same FUNCSEL value.

Altaflux: I am not sure I understand your comment.
Are you proposing instead of having a DynFunction for XipCs1 it reuses Clock as they have same FUNCSEL value?

Would that mean also changing/removing? https://github.com/rp-rs/rp-hal/pull/873/files#diff-514435c8bc1dac690be86445ca032b00cea05a617a1fef193197402cf99d94e0R211:
```rust
pin_valid_func!(bank0 as Gpio, [XipCs1], [0, 8, 19, 47]);
```

I am not sure this would be a correct change as although both Clock and XipCs1 use the same FUNCSEL value; the pins that can be Clock and the pins that can be XipCs1 are not the same.

If the plan is not to change `rp235x-pac` I think that we keep both `DynFunction::Clock` and `DynFunction::XipCs1` and simply point them to the same `FUNCSEL_A::GPCK` as currently done on: https://github.com/rp-rs/rp-hal/pull/873/files#diff-e4e5afea6f39b9d83036b039760824fc1f6ad0138f1106bdaa18d11d4b932577R171

This way the api exposed by the HAL is sound and matches expected behavior and the discrepancy inside  `rp235x-pac` 's `FUNCSEL_A` is not exposed.

Unless I am misunderstanding your comment.

rp235x-hal/src/gpio/pin.rs
Altaflux: Aren't pins that can be set as XipCs1 already being limited by:
https://github.com/rp-rs/rp-hal/pull/873/files#diff-514435c8bc1dac690be86445ca032b00cea05a617a1fef193197402cf99d94e0R174
and 
https://github.com/rp-rs/rp-hal/pull/873/files#diff-514435c8bc1dac690be86445ca032b00cea05a617a1fef193197402cf99d94e0R211
?

rp235x-hal/src/gpio/func.rs
jonathanpallant: > Are you proposing instead of having a DynFunction for XipCs1 it reuses Clock as they have same FUNCSEL value?

I was. But I think you're right - it's probably better to keep them separate.

rp235x-hal/src/gpio/pin.rs
jonathanpallant: Yup, I missed that. It seems fine.

Altaflux:
Hey @thejpster let me know if there is anything else you would like to be changed or if you would like to wait until https://github.com/rp-rs/rp235x-pac/issues/7 gets addressed.

thanks

thejpster (approved):

thejpster:
Looks ok. I'd like to try it though. Do you have an example?

thejpster:
> If you have a Pimoroni Pico Plus 2

I do. I used it to run the Neotron OS demo at RustConf, as it happens.

Altaflux:
@thejpster sorry for the delay, got caught up with other work.
I have built a sample project that uses FunctionXipCs1 to enable the usage of the PSRAM module of a Pimoroni Pico Plus 2.
It uses the PSRAM as the Heap for the example code (based on the alloc.rs example from `rp235x-hal-examples`).

https://github.com/Altaflux/rp2350-psram-test

thejpster:
The LED is blinking away! Gosh I wish we had probe-rs and defmt working though.

thejpster (approved):

Issue #864, opened 2024-10-17
Replace `static mut` for CORE1_STACK


jannic:
Both in our docs and in some examples, we suggest using a `static mut` for allocating memory for core1's stack:
```
  static mut CORE1_STACK: Stack<SIZE: 4096> = Stack::new();
[...]
    core1.spawn(unsafe { &mut CORE1_STACK.mem }, [...] )
```

With rust 1.83, this will cause a warning (at least it does with current beta), and with edition 2024 the lint will become deny-by-default. (https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html)

Therefore, we should suggest something else. Maybe we should also change the signature of `multicore::Core::spawn` to take a pointer instead of a  `&'static mut` for the stack?

thejpster:
Yeah we shouldn't use a reference here - a pointer is fine. But not a *const u8 because it might not be sufficiently aligned.

For the global a UnsafeCell with a MaybeUninit makes sense because we don't need to init the stack to anything.

thejpster:
Yeah we shouldn't use a reference here - a pointer is fine. But not a *const u8 because it might not be sufficiently aligned.

For the global a UnsafeCell with a MaybeUninit makes sense because we don't need to init the stack to anything.

9names:
> Yeah we shouldn't use a reference here - a pointer is fine. But not a *const u8 because it might not be sufficiently aligned.

The Stack struct itself ensures the address is sufficiently aligned, is that not enough?

jannic:
> If the argument is a pointer to a u8, what forces them to use the Stack struct?

Well if the spawn function takes a pointer it doesn't really matter if it's a `*mut u8` or a `*mut Stack<SIZE>`: Casting pointers is safe, so the function signature doesn't guarantee that the pointer points to anything sensible.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=74374413e46fd1078eb93c58c381e2ae

So that would be an argument to find a different solution, where spawn gets passed some reference or struct that actually guarantees alignment. It's not like the current function signature is wrong. It's just that the most obvious way to get a reference with a static lifetime, using a `static mut`, is dangerous and deprecated.

9names:
We could make `multicore::spawn()` fallible (and return Err if not aligned), or keep it infallible and panic instead?

jannic:
[...]

PR #874, opened 2024-11-13
Fix unsoundness in definition of stack for spawning core1

jannic:
Implementation of the approach suggested in https://github.com/rp-rs/rp-hal/issues/864#issuecomment-2463262058

- [x] also update rp235x code
- [x] test on hardware


thejpster (approved):

thejpster:
Looks great. Much nicer.

Issue #879, opened 2024-12-18
Calling configure_clock runs a todo!() if gpin0 or gpin1 is supplied as clock source


PietPtr:
Minimal example:

```rust
    let gpin0: Pin<Gpio20, FunctionClock, PullNone> = pins.gpio20.reconfigure();

    clocks
        .system_clock
        .configure_clock(&gpin0, 12.MHz()) // 12 MHz is I guess a placeholder, true frequency is unknown (and unnecessary to be known) for my application.
        .unwrap();
```

Triggers this todo: https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/src/clocks/clock_sources.rs#L79

I could of course set this clock up manually, but I'd like to use the rp-hal functions for switching clocks as they already implement switching the glitchy mux correctly. If some other way of nicely doing this is already available I'd love to hear about it!

thejpster:
You might be the first person to try this. I guess when the clock is coming from an external GPIO pin we have to rely on the user to tell us what the clock rate is, and to store that value somewhere for use later (e.g. when calculating UART baud rate dividers).

I would be happy to accept a PR to make this use-case work better.

PR #884, opened 2024-12-27
Add support for external clocks on gpin GPIO pins

PietPtr:
Fixes https://github.com/rp-rs/rp-hal/issues/879 by adding a `GpIn0` and `GpIn1` struct, which implement `ClockSource`, such that a user can call `configure_clocks` for these clock inputs and use the safe clock switching logic already built in.

I tried keeping the gpin.rs style kind of similar to other pin related interfaces (e.g. I2C), let me know if this aligns to your code style :)

I've renamed the type from `GPin` to `GpIn`, I believe it's a "General Purpose (clock) INput", hence the `gpin` name in the datasheet (similarly, there's also a `gpout`).

I've also provided an example which I ran on an RPi Pico, on an oscilloscope I could see that the pulses the example code drives are in sync with the provided clock :tada:

### DIFF
--- a/rp2040-hal-examples/src/bin/gpin.rs
+++ b/rp2040-hal-examples/src/bin/gpin.rs
@@ -1,0 +1,83 @@
+//! # gpin External Clocks example
+//!
+//! This application demonstrates how to clock the processor using an external clock on GPIO20
+//!
+//! It may need to be adapted to your particular board layout and/or pin assignment.
+//!
+//! See the top-level `README.md` file for Copyright and license details.
+
+#![no_std]
+#![no_main]
+
+use embedded_hal_0_2::digital::v2::ToggleableOutputPin;
+// Ensure we halt the program on panic (if we don't mention this crate it won't
+// be linked)
+use panic_halt as _;
+
+// To use the .MHz() function
+use fugit::RateExtU32;
+
+use rp2040_hal::clocks::ClockSource;
+// Alias for our HAL crate
+use rp2040_hal as hal;
+
+// Necessary HAL types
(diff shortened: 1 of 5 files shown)


thejpster:
Looks good to me. But perhaps `GpInX::new(pin)` should also take the input frequency as an argument, so we don't have to default it to 0 Hz?

PietPtr:
Good point, 0 Hz is an annoying default to have anyway as it causes issues in calculating clock dividers. I've updated it.

rp2040-hal-examples/src/bin/gpin.rs
```diff
@@ -0,0 +1,83 @@
...
+///
+/// The function configures the RP2040 to accept an external clock on Gpio20,
+/// then configures the system clock to run off this clock.
+#[rp2040_hal::entry]
+fn main() -> ! {
+    let mut pac = pac::Peripherals::take().unwrap();
+
+    let sio = Sio::new(pac.SIO);
+
+    let _xosc = setup_xosc_blocking(pac.XOSC, EXTERNAL_XTAL_FREQ_HZ.Hz()).unwrap();
```
thejpster, line 57: Do we need to set up the crystal oscillator to make this demo work, or can we take it out?

PietPtr: Yeah true, we don't need it, I've removed it.

thejpster (approved):

thejpster:
Thank you for the PR! Looks good to me. I guess we should do another PR adding support for RP2350 as well.

jannic:
I was curious and saw an opportunity to use my oscilloscope :smile: so I tried it. Synchronized two picos using gpout/gpin, with a 10MHz clock signal.
Works perfectly!

PR #889, opened 2025-01-13
Derive Debug and defmt::Format for StackAllocation

### LINKED ISSUE #888: `multicore::StackAllocation` should `impl Debug`
It would be useful if you could print a `StackAllocation`, for logging purposes.

We should impl `core::fmt::Debug`, and, if the `defmt` feature is set, we should also impl `defmt::Format`.

jannic:
The derived formatting looks like `StackAllocation { mem: 0x20000040..0x20004040 }`. While this exposes an implementation detail (the naming of the `mem` field), I think it's good enough for debug output.

Closes #888


thejpster (approved):

thejpster:
I really need to work out how to merge these HALs so we don't have to do everything twice.

PR #890, opened 2025-01-25
Update embedded-hal-bus to version 0.3.0

jannic:


thejpster:
>  embedded-hal-bus@0.3.0 requires rustc 1.81

thejpster:
My general rule is to use Ferrocene 24.05 for my MSRV, which is based on Rust 1.76.0.

jannic:
> > embedded-hal-bus@0.3.0 requires rustc 1.81

Indeed. I didn't notice this before pushing. As 0.3.0 doesn't provide us any advantage, I'd say we should stay with 0.2.0 for the time being.

9names:
> My general rule is to use Ferrocene 24.05 for my MSRV, which is based on Rust 1.76.0.

Perhaps you should communicate this desire to the HAL team so they are more sensitive to MSRV bumps? https://github.com/rust-embedded/embedded-hal/pull/630 got merged because no-one spoke up

PR #893, opened 2025-02-03
Implement set_u32 for PinGroup on rp235x

jannic:


thejpster:
Seems like a straight copy of the RP2040 one.

thejpster (approved):

PR #895, opened 2025-02-04
Fix implementations of embedded_io::Write for UartPeripheral

### LINKED ISSUE #840: Our implementation of embedded_io::Write is wrong
See https://github.com/rp-rs/rp-hal/pull/837#issuecomment-2305371353

* [x] RP2040 HAL (https://github.com/rp-rs/rp-hal/pull/838)
* [ ] RP235x HAL

jannic:
Fixes #840

### DIFF
--- a/rp2040-hal/src/uart/peripheral.rs
+++ b/rp2040-hal/src/uart/peripheral.rs
@@ -504,6 +504,7 @@ impl<D: UartDevice, P: ValidUartPinout<D>> embedded_io::ReadReady
 
 impl<D: UartDevice, P: ValidUartPinout<D>> embedded_io::Write for UartPeripheral<Enabled, D, P> {
     fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Error> {
+        // Blocks if and only if no bytes can be written.
         let remaining = nb::block!(super::writer::write_raw(&self.device, buf)).unwrap(); // Infallible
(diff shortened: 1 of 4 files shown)


rp2040-hal/src/uart/peripheral.rs
```diff
@@ -504,8 +504,8 @@ impl<D: UartDevice, P: ValidUartPinout<D>> embedded_io::ReadReady
 
 impl<D: UartDevice, P: ValidUartPinout<D>> embedded_io::Write for UartPeripheral<Enabled, D, P> {
     fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Error> {
-        self.write_full_blocking(buf);
-        Ok(buf.len())
+        let remaining = nb::block!(super::writer::write_raw(&self.device, buf)).unwrap(); // Infallible
```
thejpster, line 507: Might be worth repeating the embedded_io::Write rules here ("If the writer is not currently ready to accept more bytes (for example, its buffer is full), this function blocks until it is ready to accept least one byte.")

thejpster: and that write_raw only blocks if the FIFO is full and nothing is written to it.

thejpster (approved):
Looks technically correct, but some comments might help the next person who looks at this :)

PR #898, opened 2025-02-06
Set EXTEXCLALL when enabling multi-core operation

### LINKED ISSUE #897: Fix atomics on RP2350
We probably need a fix similar to https://github.com/embassy-rs/embassy/pull/3851/commits/9da04cc38ea5cc17740bd9921f9f5cbb1c689a31, or configure the MMU.

jannic:
Fixes #897

Also adds an on-target test to verify that atomics are working, and fixes initialization between on-target tests.

### DIFF
--- a/on-target-tests/Cargo.toml
+++ b/on-target-tests/Cargo.toml
@@ -45,6 +45,10 @@ name = "i2c_loopback_async"
 harness = false
 name = "gpio"
 
+[[test]]
+harness = false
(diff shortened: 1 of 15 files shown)


thejpster:
Looks great.

Does the test fail if you don't enable the bit?

thejpster (approved):

thejpster:
Excellent!

PR #900, opened 2025-02-08
Enable co-processors when spawning on core1

### LINKED ISSUE #899: RP2350 multicore enable FPU
Apparently FPU is only activated by default for the first, but not the second core.
Does it make sense to do this by default for the second core as well?

By default running any floating point operations on the second core crashes:
```Rust
let mut mc = Multicore::new(&mut pac.PSM, &mut pac.PPB, &mut sio.fifo);
let cores = mc.cores();
let core1 = &mut cores[1];

let _test = core1.spawn(unsafe { &mut CORE1_STACK.mem }, move || {
    loop {
        for i in 0..20 {
            timer.delay_ns((i as f64 * 20.33333) as u32)
        }

        onboard_led.set_high().unwrap();
    }
});
```

By enabling FPU access, the example succeeds:
```Rust
let _test = core1.spawn(unsafe { &mut CORE1_STACK.mem }, move || {
    unsafe {
        // SCB CPACR Register Address (Coprocessor Access Control Register)
        let scb_cpacr = 0xE000ED88 as *mut u32;

        // Enable CP10 and CP11 (FPU access)
        core::ptr::write_volatile(scb_cpacr, core::ptr::read_volatile(scb_cpacr) | (0xF << 20));

        asm::dsb();
        asm::isb();
    }

    loop {
        for i in 0..20 {
            timer.delay_ns((i as f64 * 20.33333) as u32)
        }

        onboard_led.set_high().unwrap();
    }
});
```

jannic:
Fixes #899

### DIFF
--- a/on-target-tests/tests/multicore.rs
+++ b/on-target-tests/tests/multicore.rs
@@ -100,18 +100,46 @@ mod tests {
         let counter = super::COUNTER.load(Ordering::Acquire);
         assert_eq!(2 * super::STEPS, counter);
     }
+
+    #[test]
+    fn check_floats() {
+        super::STATE.store(3, Ordering::Release);
+        super::calculations();
+        while super::STATE.load(Ordering::Acquire) != 4 {}
+
+        let counter = super::COUNTER.load(Ordering::Acquire);
+        assert_eq!(2 * super::STEPS, counter);
+    }
 }
 
 fn core1_task() {
     loop {
         match STATE.load(Ordering::Acquire) {
+            // check_atomics
             1 => {
                 for _ in 0..STEPS {
                     COUNTER.fetch_add(1, Ordering::Relaxed);
                 }
                 STATE.store(2, Ordering::Release);
             }
+            // response to check_atomics
+            2 => (),
+            // check_floats
+            3 => {
+                calculations();
+                STATE.store(4, Ordering::Release);
+            }
+            // response to check_floats
+            4 => (),
             _ => (),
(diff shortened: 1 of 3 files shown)


on-target-tests/tests/multicore.rs
```diff
@@ -100,18 +100,46 @@ mod tests {
...
     }
+
+    #[test]
+    fn check_floats() {
+        super::STATE.store(3, Ordering::Release);
+        super::calculations();
+        while super::STATE.load(Ordering::Acquire) != 4 {}
+
+        let counter = super::COUNTER.load(Ordering::Acquire);
+        assert_eq!(2 * super::STEPS, counter);
```
thejpster, line 111: Is this right? COUNTER only changes when STATE is 1? Perhaps it could do 1.5 * 2.0 or something and store the result as an integer to be checked on Core 0?

jannic: No, it's not. I forgot to remove these lines and the test succeeds accidentally because the other test case runs first and leaves the counter at that value.

It's not needed to have an assert here. If FP Code doesn't work properly on core1, it will fault and STATE won't be set to 4. It's a bit unfortunate that this just hangs the test instead of returning a useful error message, but if the test passes it shows that the FPU has been properly initialized.

BTW the double-precision coprocessor doesn't work at all in the test cases because they don't use the entry macro.

thejpster (approved):

PR #902, opened 2025-02-20
Fix manual_div_ceil clippy lints

jannic:


jannic:
It was obviously naive to expect that `cargo clippy --fix` creates properly formatted code.

9names (approved):

thejpster:
I think this makes the code worse at opt-level 0.

https://rust.godbolt.org/z/cGdPhs1Kj

An add and divide the compiler can optimise because it can see the divide is by a multiple of two. Using div_ceil means there's a function call and that optimisation no longer applies.

At opt-level=1 it's not as bad, but still different:

https://rust.godbolt.org/z/ceb4xKz8z

rp2040-hal/src/usb.rs
```diff
@@ -256,7 +256,7 @@ impl Inner {
             // size in 64bytes units.
             // NOTE: the compiler is smart enough to recognize /64 as a 6bit right shift so let's
             // keep the division here for the sake of clarity
```
thejpster, line 258: This is not true any more.

rp235x-hal/src/usb.rs
```diff
@@ -161,7 +161,7 @@ impl Inner {
             // size in 64bytes units.
             // NOTE: the compiler is smart enough to recognize /64 as a 6bit right shift so let's
             // keep the division here for the sake of clarity
```
thejpster, line 163: This is not true any more.

thejpster:
That's a good idea

thejpster:
Ok, I'm convinced.

jannic:
> Ok, I'm convinced.
> 
> Does this mean the comments about division still hold true?

I'll have a closer look. It's worth to check the rp2350 version as well, after all it's a different target so the assembly output may be different.

But ignoring the generated code: Do you think the `div_ceil` version is more readable, or is it a misguided approach in the first place?

I think it's clearer, but only by a small margin. So if the generated code turns out to be worse, we should keep the old code. Perhaps with a comment that it's equivalent to div_ceil, but faster.

thejpster:
I think it's the same, or perhaps slightly less clear. But then I'm a grizzled old programmer so I'm familiar with the idea of `X + (N - 1) / N` as an idiom. Perhaps for those less familiar, a function that has nice documentation is better.

jannic:
On rp2350 (arm), depending on the optimization level, div_ceil sometimes leads to shorter (and probably faster?) code:
```
process_old:
        and     r0, r0, #127
        adds    r0, #1
        lsrs    r0, r0, #1
        bx      lr

process_new:
        and     r0, r0, #127
        sub.w   r0, r0, r0, lsr #1
        bx      lr
```
https://rust.godbolt.org/z/1bvr1dY48
(TIL that thumbv8 has this `operand2` that can do implicit shifts. And the code transforms done by the optimizer are impressive. `x.div_ceil(2)` → `x - (x >> 1)`. Needs at least rustc 1.82.0 )

On the other hand, on riscv, in most optimization levels div_ceil needs one instruction more:
```
process_old:
        andi    a0, a0, 127
        addi    a0, a0, 1
        srli    a0, a0, 1
        ret

process_new:
        slli    a1, a0, 25
        srli    a1, a1, 26
        andi    a0, a0, 1
        add     a0, a0, a1
        ret
```
https://rust.godbolt.org/z/1zrf96hKK

I don't think there's a clear winner. I tend to follow the clippy suggestion and use div_ceil, and only disable lints where the suggested code is clearly worse.

rp2040-hal/src/usb.rs
jannic: Looks like still is. Or do you think it's confusing that the comment says `/64` and not `div_ceil(64)`? Both are divisions.

9names:
Did you do the riscv build with the extra extensions enabled? Kinda curious if they have an effdct

thejpster:
Yeah, let's merge!

Issue #907, opened 2025-03-09
How to use Generics with `set_duty` from PWM


DavidFrings:
I'm trying to implement a generic `Motor` struct using `embedded-hal` and `rp2040-hal`. However, I am facing an issue when calling `set_duty_cycle_percent` (or any other `set_duty` method) on a generic PWM channel.

### Example Code:
```rust
use embedded_hal::pwm::SetDutyCycle;
use rp2040_hal::pwm::{Channel, ChannelId, FreeRunning, Slice, SliceId};

pub struct Motor<T, U, V, W>
where
    T: SliceId,
    U: SliceId,
    V: ChannelId + SetDutyCycle,
    W: ChannelId + SetDutyCycle,
{
    pub pin_0: Channel<Slice<T, FreeRunning>, V>,
    pub pin_1: Channel<Slice<U, FreeRunning>, W>,
}

impl<T, U, V, W> Motor<T, U, V, W>
where
    T: SliceId,
    U: SliceId,
    V: ChannelId + SetDutyCycle,
    W: ChannelId + SetDutyCycle,
{
    pub fn forward(self, speed: u8) {
        self.pin_0.set_duty_cycle_percent(speed).expect("PWM-Error");
        self.pin_1.set_duty_cycle_fully_off().expect("PWM-Error");
    }
}
```

### Error:
```rust
error[E0599]: no method named `set_duty_cycle_percent` found for struct `rp2040_hal::pwm::Channel` in the current scope
  --> src\motor.rs:91:20
   |
91 |         self.pin_0.set_duty_cycle_percent(speed).expect("PWM-Error");
   |                    ^^^^^^^^^^^^^^^^^^^^^^ method not found in `Channel<Slice<T, FreeRunning>, V>`
```

This error occurs with all variations of `set_duty`.

### Question:
How can I properly use generics to call `set_duty` methods on `Channel<Slice<T, FreeRunning>, V>`? Do I need additional trait bounds or a different approach?

Any help would be appreciated!

jannic:
You can specify that you need a Channel that implements SetDutyCycle:

```
use rp2040_hal::pwm::{Channel, FreeRunning, Slice, SliceId, ChannelId};

pub struct Motor<T, U, V, W>
where
    T: SliceId,
    U: SliceId,
    V: ChannelId,
    W: ChannelId,
    Channel<Slice<T, FreeRunning>, V>: SetDutyCycle,
[...]

PR #909, opened 2025-03-10
Implement SetDutyCycle generically on ChannelId

jannic:
Issue https://github.com/rp-rs/rp-hal/issues/907 showed that SetDutyCycle was difficult to use because its implementation depends on whether it's an A or B channel.

This change avoids the requirement to manually specify the channel type.

It also removes some unnecessary code duplication.


9names (approved):
Much cleaner, thanks jannic

PR #911, opened 2025-03-12
Bump version of rp-binary-info to 0.1.1

jannic:


rp235x-hal-examples/Cargo.toml
```diff
@@ -33,6 +33,9 @@ panic-halt = "0.2.0"
 pio = "0.2.0"
 pio-proc = "0.2.0"
 rp235x-hal = {path = "../rp235x-hal", version = "0.3.0", features = ["binary-info", "critical-section-impl", "rt", "defmt"]}
+# The examples use features not yet available in rp-binary-info 0.1.0,
+# so the minimum version implied by the rp235x-hal 0.3.0 dependency is not sufficient.
+rp-binary-info = "0.1.1"
```
thejpster, line 38: I guess this also needs a path = ?

jannic: Indeed. Would be fine without path once 0.1.1 is released, but for now path = is required.

thejpster (approved):

PR #913, opened 2025-03-21
pio: enable async.await support

ithinuel:
I didn’t get time to test this on rp235x yet. If anyone has the bandwidth please try it and let me know <3

An example use case:
- https://github.com/rp-rs/i2c-pio-rs/pull/30

### DIFF
--- a/rp2040-hal/src/pio.rs
+++ b/rp2040-hal/src/pio.rs
@@ -665,8 +665,8 @@ impl<SM: ValidStateMachine, State> StateMachine<SM, State> {
     pub fn clear_fifos(&mut self) {
         // Safety: all accesses to these registers are controlled by this instance
         unsafe {
-            let sm = &self.sm.sm();
-            let sm_shiftctrl = &sm.sm_shiftctrl();
+            let sm = self.sm.sm();
+            let sm_shiftctrl = sm.sm_shiftctrl();
             let mut current = false;
             // Toggling the FIFO join state clears the fifo
             sm_shiftctrl.modify(|r, w| {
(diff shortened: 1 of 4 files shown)


rp2040-hal/src/pio/non_blocking.rs
```diff
@@ -0,0 +1,164 @@
...
+};
+use crate::dma::TransferSize;
+
+use super::{Interrupt, PIOExt, PioIRQ, Rx, Tx, ValidStateMachine};
+
+macro_rules! impl_wakeable {
+    ($t:ident, $doc:expr, $offset:expr) => {
+        impl<S: ValidStateMachine, W: TransferSize> Wakeable for $t<S, W> {
+            fn waker() -> &'static IrqWaker {
+                static WAKER: IrqWaker = IrqWaker::new();
```
jannic, line 15: First: I may be completely wrong, feel free to just tell me that I'm writing nonsense. I don't understand the internals of async rust well enough to confidently state that there is an issue with this code. I just don't understand how it is supposed to work.

_If_ I understand this correctly, there is an issue with this `static`: It is shared with all instances of `Rx` (or `Tx`). There are no separate instances per monomorphization (https://doc.rust-lang.org/reference/items/static-items.html#r-items.static.generics). Therefore, if there are multiple state machines running in parallel, and one tries to `await` the `Rx` (or `Tx`) of more than one state machine at the same time, the second one to be polled  will overwrite the waker of the first one.

Of course, this only matters if there actually are multiple wakers. With the trivial mock-executor used in the on-target-tests, where a single futures is called to completion from a single task, this wouldn't make a difference.

But if, for example, the two `Rx` are awaited from different CPU cores, this code may wake the wrong one, and the other one might stall.

ithinuel: Well, you got it right >< I thought there would be 1 instance per monomorphisation. I'll fix this !!

PR #923, opened 2025-05-06
Update pio to version 0.3.0

jannic:


thejpster:
> home@0.5.11 requires rustc 1.81

No idea what that is about

jannic:
> > home@0.5.11 requires rustc 1.81
> 
> No idea what that is about

As discussed on matrix. Fixed in #924

thejpster (approved):

PR #925, opened 2025-05-08
both timers support up to 4 alarms

McBandaska:
in previous version, there was possibility to have 2 timers but only 4 alarms total, here are changes, that makes possible to configure up to 4 alarms for each timer

### DIFF
--- a/rp235x-hal/src/timer.rs
+++ b/rp235x-hal/src/timer.rs
@@ -22,18 +22,19 @@ use crate::{
 /// Instant type used by the Timer & Alarm methods.
 pub type Instant = TimerInstantU64<1_000_000>;
 
-static ALARMS: AtomicU8 = AtomicU8::new(0x0F);
-fn take_alarm(mask: u8) -> bool {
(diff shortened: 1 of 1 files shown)


ithinuel (commented):
The changes seem sensible. 👍 
I didn’t cross check with the datasheet nor the hardware so I’m only leaving this review as a comment.

thejpster:
If there are only four alarms for each timer, we only need a single AtomicU8, right?

Whilst we're here, I think we can use https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU8.html#method.fetch_and instead of a critical section.

jannic:
> If there are only four alarms for each timer, we only need a single AtomicU8, right?

Did you find some reason that two atomics are better than one? (I don't complain, the current code looks clean and using bits from a single u8 likely needs some bit shifts that are less obvious. Just curious.)

thejpster:
I mean it seems obvious now you mention it - yeah, I can see no reason why this can't just be one AtomicU8.

thejpster:
We can always make it even better in another PR.

PR #926, opened 2025-05-09
Add support for PUT and GET FIFO modes

KewlKris:
I added support for the new (as of RP235x) RXFIFO PUT and GET modes. These give a PIO state machine direct access to words in RXFIFO.

I also added an example which demonstrates writing to and reading from RXFIFO, though I don't know if it's super necessary.

Please make any changes/corrections :)

### DIFF
--- a/rp235x-hal-examples/src/bin/pio_putget.rs
+++ b/rp235x-hal-examples/src/bin/pio_putget.rs
@@ -1,0 +1,100 @@
+//! This example tests RXFIFO reads/writes using PUTGET mode. GPIO25 is set high if the
+//! test is successful and low if the test fails.
+//!
+//! If a LED is connected to that pin, like on a Pico board, the LED should turn on.
+#![no_std]
+#![no_main]
+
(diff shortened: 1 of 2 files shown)


thejpster (approved):
LGTM

KewlKris:
Yes you're right it would make sense to mention that. I could definitely see that as a major headache for someone if they don't know about it.

Do you think this is a good way to phrase it? I want to get the point across that it messes with the system without the doc being too wordy...

```rust
/// The memory of the RX FIFO is available for random write access by the state machine, but the system can only read from it.
RxPut,
/// The memory of RX FIFO is available for random read access by the state machine, but the system can only write to it.
RxGet,
/// The memory of RXFIFO is available for random read and write access by the state machine, but the system can no longer read or write to it.
RxPutGet,
```

9names:
I have no issues with that wording, it is concise and accurate.

KewlKris:
Ok thanks

9names (approved):

PR #931, opened 2025-05-31
Add support for PIO2

horazont:
Tested on an Raspberry Pi Pico 2.


thejpster:
Seems sensible. Can you add an example to show it works?

thejpster:
The example uses PIO0 but also PioFunction2, so either I'm confused or it shouldn't actually work, right? If it works, isn't that a bug somewhere?

horazont:
Nah, that's just me pushing broken stuff. Going to fix that later, well spotted!

horazont:
Fixed & tested on my Pico 2 board :)

thejpster (approved):

PR #932, opened 2025-06-01
Fix rp_binary_end

jannic:
Unfortunately the fix to rp_binary_end is a breaking change. And as rp-binary-info is re-exported by the HALs, it will also be a breaking change for them.

I wonder if it's a good idea to re-export this crate from the HALs. If we have a breaking change anyway, we can also use it to remove the re-export. Then future breaking changes to rp-binary-info won't infect the HALs, again.

As far as I can see there's no necessity for the re-export, it's a pure convenience thing. And in this case, the difference between using the re-exports and adding a direct dependency on rp-binary-info is minimal.

### DIFF
--- a/rp-binary-info/src/macros.rs
+++ b/rp-binary-info/src/macros.rs
@@ -30,7 +30,7 @@ macro_rules! str {
     }};
 }
 
-/// Generate a static item containing the given string, and return its
+/// Generate a static item containing the given integer, and return its
 /// [`EntryAddr`](super::EntryAddr).
 ///
 /// You must pass a numeric tag, a numeric ID, and `&CStr` (which is always
@@ -39,6 +39,18 @@ macro_rules! str {
 macro_rules! int {
     ($tag:expr, $id:expr, $int:expr) => {{
         static ENTRY: $crate::IntegerEntry = $crate::IntegerEntry::new($tag, $id, $int);
+        ENTRY.addr()
+    }};
+}
+
+/// Generate a static item containing the given pointer, and return its
+/// [`EntryAddr`](super::EntryAddr).
+///
+/// You must pass a numeric tag, a numeric ID, and a pointer
+#[macro_export]
+macro_rules! pointer {
+    ($tag:expr, $id:expr, $ptr:expr) => {{
+        static ENTRY: $crate::PointerEntry = $crate::PointerEntry::new($tag, $id, $ptr);
         ENTRY.addr()
     }};
 }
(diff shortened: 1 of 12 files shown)


thejpster:
IIRC it started in the HAL but embassy asked us to break it out so they could use it without pulling in our HAL.

I'm fine with not exporting it.

rp-binary-info/src/macros.rs
```diff
@@ -166,4 +178,20 @@ macro_rules! rp_pico_board {
...
+/// by the linker script that is located at the end of the binary.
+#[macro_export]
+macro_rules! rp_binary_end {
+    ($ptr:ident) => {{
+        $crate::pointer!(
+            $crate::consts::TAG_RASPBERRY_PI,
+            $crate::consts::ID_RP_BINARY_END,
+            // `unsafe` only needed because MSRV does not yet
+            // contain https://github.com/rust-lang/rust/pull/125834
+            unsafe { core::ptr::addr_of!($ptr).cast() }
```
thejpster, line 192: Will this trip 'unneeded unsafe' warnings?

rp-binary-info/src/types.rs
```diff
@@ -189,4 +189,36 @@ impl IntegerEntry {
     }
 }
 
+/// An alias for IntegerEntry, taking a pointerinstead of an integer
```
thejpster, line 192: Missing space

rp-binary-info/src/types.rs
```diff
@@ -189,4 +189,36 @@ impl IntegerEntry {
...
+#[repr(C)]
+pub struct PointerEntry {
+    header: EntryCommon,
+    id: u32,
+    value: *const (),
+}
+
+impl PointerEntry {
+    /// Create a new `StringEntry`
```
thejpster, line 201: Typo

rp-binary-info/src/types.rs
```diff
@@ -189,4 +189,36 @@ impl IntegerEntry {
...
+    pub const fn new(tag: u16, id: u32, value: *const ()) -> PointerEntry {
+        PointerEntry {
+            header: EntryCommon {
+                data_type: DataType::IdAndInt,
```
thejpster, line 205: We should point out we're relying on an int and a pointer to have the same representation - there's an implicit transmute happening here to bypass an issue with pointers in const functions.

rp-binary-info/src/macros.rs
jannic: No, not even on nightly.

I guess this is one of the lints that get ignored if they trigger inside an external macro. But I didn't find documentation about this feature. So all I know is that the binary_info_demo compiled without warnings on all compilers I tried.

PR #939, opened 2025-06-17
Expose watchdog reset reason

jackwilsdon:
This PR adds a `reason()` method to `rp235x_hal::watchdog::Watchdog`, returning the reset reason from the watchdog peripheral.

Screenshot from RP2350 datasheet showing watchdog REASON register

Both the force and timer bits could technically be set simultaneously, but I've not been able to reproduce this.

One thing to note is that some ways to reboot the RP2350 give unexpected reasons:

| Method | Reason |
|-|-|
| [`reboot(RebootKind::Normal, RebootArch::Normal)`](https://docs.rs/rp235x-hal/0.3.0/rp235x_hal/reboot/fn.reboot.html) | Timeout |
| [`explicit_buy`](https://docs.rs/rp235x-hal/0.3.0/rp235x_hal/rom_data/fn.explicit_buy.html) | Timeout |

The official SDK avoids this confusion by not exposing the reason directly and instead providing [`watchdog_caused_reboot`](https://www.raspberrypi.com/documentation/pico-sdk/hardware.html#group_hardware_watchdog_1ga65f5d24169045b8c7dc709e572535d94) (which checks reason and boot type) and [`watchdog_enable_caused_reboot`](https://www.raspberrypi.com/documentation/pico-sdk/hardware.html#group_hardware_watchdog_1ga455aa48ca6f11298e184d2ae0e81a085) (which checks reason and a scratch register set by `watchdog_enable`).

### DIFF
--- a/rp235x-hal/src/watchdog.rs
+++ b/rp235x-hal/src/watchdog.rs
@@ -47,6 +47,14 @@ pub struct Watchdog {
     load_value: u32, // decremented by 2 per tick (µs)
 }
 
+/// Watchdog reset reason
+pub enum ResetReason {
+    /// A manual watchdog trigger caused the reset.
+    Force,
+    /// A watchdog timeout caused the reset.
+    Timeout,
+}
+
 #[derive(Debug)]
 #[allow(missing_docs)]
 /// Scratch registers of the watchdog peripheral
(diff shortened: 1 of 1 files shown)


thejpster:
I wonder if we should do it how the Pico SDK does it, to avoid the confusion.

jackwilsdon:
We can implement `watchdog_caused_reboot` quite easily but we'd need to set a value in the scratch register on enable to implement `watchdog_enable_caused_reboot` - is this something we'd want to do? Happy to implement it in this PR if so.

thejpster:
I think that would be good. I'd we use a scratch that someone else is using, they can switch to one of yes others easily enough - or just call a different reboot function that doesn't set the register.

thejpster:
I'm currently using a scratch register in my project to catch whether we've had a Core 0 restart (thanks to probe-rs) or a full system restart. If I detect the former, I use the watchdog to force the latter, so that Core 1 also gets reset. I'd have no problem changing that if required.

PR #940, opened 2025-06-18
Fix rp235x USB example enumeration on Windows

### LINKED ISSUE #937: RP235x USB example fails to enumerate on Windows
Hi, I'm a newcomer to this ecosystem, so I appologise for any silly mistakes.

I've observed that the https://github.com/rp-rs/rp-hal/blob/main/rp235x-hal-examples/src/bin/usb.rs example does not enumerate on Windows.

## Steps to Repro
1. Load rp235x-hal-examples/src/bin/usb.rs to rp2350 device
2. Observe in Device Manager: "Unknown USB Device (Set Address Failed)"

## Configuration
- Windows 11 26100.4351
- Pimoroni Presto (RP2350)

I see that this is a [known issue](https://docs.rs/rp235x-hal/latest/rp235x_hal/usb/index.html#enumeration-issue-with-small-ep0-max-packet-size), and it has a workaround in the [rp 253x-hal usb poll](https://github.com/rp-rs/rp-hal/blob/38d28bd5e55298783284b4affcef2f0c4e01a357/rp235x-hal/src/usb.rs#L552). But the example uses usb-device which doesn't/can't have this workaround? A cursory glance suggests the RP2040 examples could also behave the same.

I worked around it as the advisory stated and set the `max_packet_size_0` to 64 in the example.

Minimal-C:
Increase max_packet_size_0 from 8 to 64 in example as workaround for https://docs.rs/rp235x-hal/0.3.0/rp235x_hal/usb/index.html (Enumeration issue with small EP0 max packet size)

Fixes #937 

Enumeration on Windows:
image

Cyme:
```
Bus 000 Device 021: ID 16c0:27dd Van Ooijen Technische Informatica CDC-ACM class devices (modems)
Negotiated speed: Full Speed (12Mbps)
```

### DIFF
--- a/rp235x-hal-examples/src/bin/usb.rs
+++ b/rp235x-hal-examples/src/bin/usb.rs
@@ -86,6 +86,8 @@ fn main() -> ! {
             .product("Serial port")
             .serial_number("TEST")])
         .unwrap()
+        .max_packet_size_0(64)
+        .unwrap()
         .device_class(2) // from: https://www.usb.org/defined-class-codes
         .build();
 


thejpster (approved):

thejpster:
Thank you!

PR #941, opened 2025-06-18
Update rand_core to 0.9.3

jannic:
We could easily implement both rand_core 0.6 and rand_core 0.9, if there's demand to do so. I prefer to only implement the latest version to avoid accumulating technical debt.


thejpster (approved):
I think this is fine.

PR #942, opened 2025-06-27
Fix clippy warnings generated by current beta (1.89.0-beta.2)

jannic:


rp235x-hal/src/reboot.rs
```diff
@@ -1,5 +1,6 @@
 //! Functions for rebooting the chip using the ROM.
 
+#[allow(unpredictable_function_pointer_comparisons)]
```
thejpster, line 3: Should be below the comment, like the other attribute

PR #943, opened 2025-06-28
Add HSTX support

OYTIS:
There is not much to abstract in HSTX, but it still nice to have it in HAL for resets, pins etc. There are many questionable decisions made here for sure (including, but not limited to pinout), and I would be happy to hear your feedback.

I have an example using this module in [my repo](https://github.com/OYTIS/rp235x-hstx-example), but it's too bulky for rp235x-hal-examples - maybe a rework of [ST7789 example from pico-examples](https://github.com/raspberrypi/pico-examples/blob/master/hstx/spi_lcd/hstx_spi_lcd.c) will work?

PIO coupling is out of scope for this PR.

### DIFF
--- a/rp235x-hal/src/hstx.rs
+++ b/rp235x-hal/src/hstx.rs
@@ -193,6 +193,196 @@ impl<
     type Bit7 = Opt7;
 }
 
+impl HstxPins<OptionTNone,
+              OptionTNone,
+              OptionTNone,
+              OptionTNone,
+              OptionTNone,
+              OptionTNone,
+              OptionTNone,
+              OptionTNone> {
+    /// Create an empty set of HstxPins
+    #[allow(clippy::new_without_default)]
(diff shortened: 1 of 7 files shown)


rp235x-hal/src/hstx.rs
```diff
@@ -0,0 +1,324 @@
...
+    /// Configure output shifter
+    pub fn configure_shifter(&self, shift: u8, n_shifts: u8) {
+        self.ctrl
+            .csr()
+            .modify(|_, w| unsafe { w.shift().bits(shift).n_shifts().bits(n_shifts) });
+    }
+
+    /// Configure expand shifter
+    pub fn configure_expand_shifter(
+        &self,
```
thejpster, line 121: These methods are changing shared mutable state in the hardware and so should require &mut I think.

thejpster (commented):
Just one comment. Looks great!

thejpster:
Apologies for the delay.

PR #947, opened 2025-08-12
Remove elf2uf2-rs in favour of picotool

robjwells:
elf2uf2-rs seems to be unmaintained, and as of Rust 1.89 [rejects valid ELF files](https://github.com/JoNil/elf2uf2-rs/issues/40) due to an overly strict check of the ELF header. Picotool is maintained by Raspberry Pi and does not have this problem.

This commit changes guide-level explanation in the root and RP2040 examples READMEs, and the runner in .cargo/config.toml files, to picotool.

Some wording around the flashing process as "creating a UF2 file" was changed as picotool can flash directly from an ELF file with the appropriate command-line flag. Information about deliberately creating a UF2 file has been kept.

In addition, the root README now directs people to rp-binary-info rather than stating that picotool compatible "binary info" is unsupported.

---

**A question**: Previously the root README said:

> For boards with USB Device support like the Raspberry Pi Pico, we recommend you use the UF2 process.

I have removed this explicit recommendation as the paragraph above it directs people to use picotool if they are not using a debug probe (or otherwise wish to flash over USB). Is this acceptable, or should an explicit recommendation (to use picotool over USB) be included?

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ RPI-RP2 mass storage device, use the `picotool uf2 convert` command on your
 compiled program with the `-t elf` argument.
 
 ```console
-$ picotool uf2 convert target/thumbv6m-none-eabi/release/pwm_blink -t elf
+$ picotool uf2 convert -t elf target/thumbv6m-none-eabi/release/pwm_blink
 pwm_blink.uf2
 ```
 
--- a/README.md
+++ b/README.md
@@ -67,14 +67,19 @@ rustup update stable
 rustup target add thumbv6m-none-eabi
 ```
 
-You may also want to install these helpful tools:
+You may also want to install probe-rs, to flash your device over the SWD pins
+with a debug probe:
 
 ```sh
-# Useful to creating UF2 images for the RP2040 USB Bootloader
-cargo install elf2uf2-rs --locked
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -190,6 +188,22 @@ $ cargo run --release --features "critical-section-impl,rt,defmt" --example pwm_
 (The `pwm_blink` example doesn't need all these feature flags. They are listed here
 so you can use the same command for all examples.)
 
+If you want to create a UF2 file, which is loaded by copying it over to the
+RPI-RP2 mass storage device, use the `picotool uf2 convert` command on your
+compiled program with the `-t elf` argument.
+
+```console
+$ picotool uf2 convert target/thumbv6m-none-eabi/release/pwm_blink -t elf
```
thejpster, line 196:
```suggestion
$ picotool uf2 convert -t elf target/thumbv6m-none-eabi/release/pwm_blink
```

In other cases, we put the `-t elf` before the file name. Just wanted to be consistent in case it causes confusion (it doesn't really matter where you put it).

thejpster (approved):

thejpster:
Thank you for these improvements!

PR #948, opened 2025-08-12
Don't emit binary info unless we are bare-metal.

thejpster:
The change for `#[used]` in Rust 1.89 causes linker errors when compiling this library for a Linux target (and probably other OSes as well). So now only emit this global variable when target_os is none.


thejpster:
Clippy issues are fixed by #942

PR #950, opened 2025-08-12
Make the top level README reflect that there are two HALs now.

thejpster:
Make sure we mention both HALs here, plus some of the other packages we have going on.

I've also dialled back on "Raspberry Silicon" because although that was used in the headline of the [launch blog](https://www.raspberrypi.com/news/raspberry-pi-silicon-pico-now-on-sale/), they don't seem to use the term much now.


thejpster:
> Could I ask you to sneak in a fix to a minor format problem I introduced

Done

PR #952, opened 2025-08-18
rp235x: fix names of directly mapped SRAM banks

robjwells:
This corrects the names of the directly mapped SRAM banks from 4/5 to 8/9, in line with section 2.2.3 of the RP2350 datasheet.

This affects three memory.x files; the change in the RP235x HAL itself is to a doc comment.

The incorrect names appear to just have been carried over from the RP2040, where 4/5 are the directly mapped banks. Originally spotted in [this Embassy PR](https://github.com/embassy-rs/embassy/pull/4404).

### DIFF
--- a/on-target-tests/memory_rp235x.x
+++ b/on-target-tests/memory_rp235x.x
@@ -17,8 +17,8 @@ MEMORY {
      * of access times.
      * Example: Separate stacks for core0 and core1.
      */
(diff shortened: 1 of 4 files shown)


thejpster (approved):

thejpster:
Thank you for this fix!

PR #953, opened 2025-08-26
Added wrappers for slices to allow for use of non-static memory for DMA

Gip-Gip:
Recently needed to use DMA to ensure safe transmission of data in the off event that an interrupt triggers mid-data transfer(can cause substantial packet loss in my use case). This type ensures that the DMA transfer is complete before it is dropped by producing a pointer that needs to be fed back into the wrapper.

I haven't documented but if the feedback is good I can work on that later.

### DIFF
--- a/rp-hal-common/Cargo.toml
+++ b/rp-hal-common/Cargo.toml
@@ -11,4 +11,6 @@ version = "0.1.0"
 
 # DO NOT LIST ANY PAC CRATES OR ARCHITECTURE CRATES HERE
 [dependencies]
+const-random = "0.1.18"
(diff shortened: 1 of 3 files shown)


Gip-Gip:
> Also, do y'all think this would be better PR'd to the embedded-dma crate?

My only concern with this is I don't think DMA implementations are required to return the ReadBuffer/WriteBuffer only after DMAs are complete, like the rp-hal implementation

thejpster:
I don't see anything here that is RP2x specific, so this might not be the best place for it.

The documentation should probably also discuss the ramifications of calling `std::mem::forget()` on the wrapper.

PR #963, opened 2025-09-16
Added support for pins_with_function!() and pin_with_name!()

MarkSwanson:
I submitted this to be used as a reference to save someone some time for when they decide to do a full implementation.

pub static PICOTOOL_ENTRIES: [rp_binary_info::EntryAddr; 2] = [
    rp_binary_info::pins_with_function!(
            PICO_DEFAULT_I2C_SDA_PIN,
            PICO_DEFAULT_I2C_SCL_PIN,
            GPIO_FUNC_I2C),

    rp_binary_info::pin_with_name!(
        25, // pin 25 is the LED
        c"The LED"
    ),
];

$ picotool info -a m.bin
Fixed Pin Information
 4:                   I2C0 SDA
 5:                   I2C0 SCL
 25:                  The LED

I found constants in these places:
pico-sdk/src/boards/include/boards/pico2_w.h
pico-sdk/src/common/pico_binary_info/include/pico/binary_info/structure.h pico-sdk/src/rp2350/hardware_structs/include/hardware/structs/io_bank0.h

### DIFF
--- a/rp-binary-info/src/types.rs
+++ b/rp-binary-info/src/types.rs
@@ -1,4 +1,8 @@
 //! Types for the Binary Info system
+use crate::consts::{TAG_RASPBERRY_PI};
+
+// pico-sck/src/common/pico_binary_info/include/pico/binary_info/structure.h
+const BI_PINS_ENCODING_MULTI: u32 = 2;
 
 /// This is the 'Binary Info' header block that `picotool` looks for in your UF2
 /// file/ELF file/Pico in Bootloader Mode to give you useful metadata about your
@@ -116,7 +120,7 @@ pub enum DataType {
     BlockDevice = 7,
     /// GPIO pins, with their function
     PinsWithFunction = 8,
-    /// GPIO pins, with their name
+    /// GPIO pins, with their name pico-sdk: BINARY_INFO_TYPE_PINS_WITH_NAME
     PinsWithName = 9,
     /// GPIO pins, with multiple names?
     PinsWithNames = 10,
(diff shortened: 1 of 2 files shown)


thejpster:
Can you reformat and squash?

rp-binary-info/src/types.rs
```diff
@@ -189,6 +194,62 @@ impl IntegerEntry {
...
+
+impl PinsWithFunctionEntry {
+    /// Create a new `PinsWithFunctionEntry`
+    pub const fn new(pin_0: u32, pin_1: u32, func: u32) -> PinsWithFunctionEntry {
+        PinsWithFunctionEntry {
+            header: EntryCommon {
+                data_type: DataType::PinsWithFunction,
+                tag: TAG_RASPBERRY_PI,
+            },
+            pin_encoding: BI_PINS_ENCODING_MULTI
```
thejpster, line 212: I think it's worth noting that this is a value with multiple fields, and raspberry pi seem to indicate that there are no more pins by setting two consecutive fields to the same value.

You could also consider using bitbybit::bitfield to describe this 32-bit value, and avoid needing to do a bunch of shifts using magic numbers. I [have examples you can refer to](https://github.com/rust-embedded/cortex-ar/blob/f1d8ad3c08e1917c1f2f2d08aa0072a787c96d0a/cortex-ar/src/register/dfsr.rs#L36)

thejpster: You could then change this to take a slice of pins, rather than only two pins.

rp-binary-info/src/types.rs
```diff
@@ -116,7 +120,7 @@ pub enum DataType {
     BlockDevice = 7,
     /// GPIO pins, with their function
     PinsWithFunction = 8,
-    /// GPIO pins, with their name
+    /// GPIO pins, with their name pico-sdk: BINARY_INFO_TYPE_PINS_WITH_NAME
```
thejpster, line 123:
```suggestion
    /// GPIO pins, with their name
    ///
    /// The Pico SDK calls this `BINARY_INFO_TYPE_PINS_WITH_NAME`
```

PR #965, opened 2025-09-24
unpredictable_function_pointer_comparisons lint is not known on MSRV

jannic:
Add #[allow(unknown_lints)] to fix the issue.

(I'd also be fine with ignoring this. It is only a warning, and our CI doesn't forbid warnings on MSRV, so nothing breaks if we keep it as it is.)


thejpster:
I'm fine with the warning to be honest. Users using the library won't see it.

PR #970, opened 2025-10-30
Bump MSRV to 1.82

thejpster:
CI currently fails because indexmap 2.12 requires 1.82 minimum


jannic:
> But users won't know to do that and might be confused when we don't build out of the box on the MSRV.

I don't think we can guard against that issue in any meaningful way. Even if 1.82 works today, one of our dependency might bump it's MSRV tomorrow and break the build again.

But then, we already did bump the MSRV since our last release, so it doesn't matter much if we go to 1.81 or to 1.82.

As discussed on matrix it may even be better to go directly to 1.84, so we can take advantage of the new `resolver.incompatible-rust-versions=fallback` setting.

thejpster:
This gets it fixed for now. We can go to 1.84 next time.

PR #971, opened 2025-11-16
docs: Fix broken "run examples" commands

DeflateAwning:
The old [now-removed] lines didn't work.

Pretty sure they're mostly right in the `*-examples/README.md` files (which duplicate this info).

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,9 +264,6 @@ cd rp2040-hal-examples  # Or: cd rp235x-hal-examples
 cargo run --release --bin pwm_blink
 ```
 
-Note that some examples require additional feature flags enabled:
-e.g., `--features "critical-section-impl,rt,defmt"`.
-
 If you want to create a UF2 file, which is loaded by copying it over to the
 RPI-RP2 mass storage device, use the `picotool uf2 convert` command on your
 compiled program with the `-t elf` argument.


README.md
```diff
@@ -260,11 +260,12 @@ specified 'runner'. As the 'runner' is picotool, it will flash your compiled
...
 ```sh
-cargo run --release --features "critical-section-impl,rt,defmt" --example pwm_blink
+cd rp2040-hal-examples  # Or: cd rp235x-hal-examples
+cargo run --release --bin pwm_blink
 ```
 
-(The `pwm_blink` example doesn't need all these feature flags. They are listed here
-so you can use the same command for all examples.)
+Note that some examples require additional feature flags enabled:
+e.g., `--features "critical-section-impl,rt,defmt"`.
```
9names: Is this true? I don't think there's any features in the rp*hal-examples directories any more. If we're going to fix this, you should not replace it with more incorrect info.

9names (changes requested):

README.md
DeflateAwning: Fair enough - I'm certain it gives an error if you try to run with the features where they are not required. I assumed maybe there were some places that needed it still. I'd be happy to remove them - need someone else to advise though.

DeflateAwning: Done, thanks!

DeflateAwning:
Fixed the feedback comments

PR #972, opened 2025-11-20
implement ValidSpiPinout for CS pin.

pfzetto:
Hi, 
currently there doesn't seem to be a way to configure the CS-pin of the `rp2040-hal::Spi` device.
This PR implements the `ValidSpiPinout` trait on `(Tx,Rx,Sck,Cs)` to allow setting the Cs pin of the SPI device.

I required setting the CS-pin when using the RP2040 as a SPI slave.

Cheers


jannic:
Looks good, thank you!
Can you apply the same change to `rp235x-hal/src/spi/pins.rs`?

pfzetto:
Of course, implemented in 4d7a63b.

PR #975, opened 2025-11-27
Port most of our examples to embedded-hal 1.0

jannic:
Many of our examples still used embedded-hal 0.2 traits. Migrate them to embedded-hal 1.0.

Only example left using embedded-hal 0.2 is `adc`. This is because embedded-hal 1.0 doesn't provide ADC support, and 
our hal implementation isn't usable without the embedded-hal traits. We should provide inherent methods and use them instead, but that can be done in a separate pull request.


9names (approved):
Yep, all sensible changes.
Thanks @jannic

PR #984, opened 2026-02-01
Bump hal crate version numbers in preperation for release

9names:


rp2040-hal-examples/Cargo.toml
```diff
@@ -36,9 +36,9 @@ pio = "0.3.0"
 # Should be fixed in e-h-b 0.3 via https://github.com/rust-embedded/embedded-hal/pull/607
 portable-atomic = {version = "1.7.0", features = ["critical-section"]}
 rp2040-boot2 = "0.3.0"
-rp2040-hal = {path = "../rp2040-hal", version = "0.11.0", features = ["binary-info", "critical-section-impl", "rt", "defmt"]}
+rp2040-hal = {path = "../rp2040-hal", version = "0.12.0", features = ["binary-info", "critical-section-impl", "rt", "defmt"]}
 # The examples use features not yet available in rp-binary-info 0.1.0,
```
thejpster: Is this still true?

rp235x-hal-examples/Cargo.toml
```diff
@@ -32,9 +32,9 @@ heapless = "0.8.0"
 nb = "1.0"
 panic-halt = "0.2.0"
 pio = "0.3.0"
-rp235x-hal = {path = "../rp235x-hal", version = "0.3.0", features = ["binary-info", "critical-section-impl", "rt", "defmt"]}
+rp235x-hal = {path = "../rp235x-hal", version = "0.4.0", features = ["binary-info", "critical-section-impl", "rt", "defmt"]}
 # The examples use features not yet available in rp-binary-info 0.1.0,
```
thejpster: Is this still true?

9names: Doesn't seem to be, it's pulling in 0.1.2 consistently even without.
Removing both the comment and the dep from both HALs, lets see what CI says.

9names:
I also updated the rp2040-hal-examples Cargo.toml comment about portable-atomic inclusion - we no longer need it for embedded-hal-bus, but we now need it for static-cell.
I think it's for the best to make that comment a permanent feature now.

PR #989, opened 2026-03-18
Add ReadPinHList and WritePinHList to PinGroup export

doublezerowld:
A tiny change that would allow users of the crate to define trait bounds using ``ReadPinHList`` and ``WritePinHList``, which **aren't publicly re-exported**.

For example, say a user wants to pass a ``PinGroup`` as a struct field, and use ``.set_u32()`` on the ``PinGroup``.

This is not possible, as the trait bounds aren't satisfied for ANY PinGroup. ``.set_u32()`` **only works on** ``PinGroup``s **with these bounds:**

```rs
impl<H, T> PinGroup<HCons<H, T>>
where
    HCons<H, T>: ReadPinHList + WritePinHList,
    H: AnyPin,
{
    ...
}
```

The issue is that, because both of those traits aren't public, **users cannot define trait bounds for their own code**:

```rs
use rp2040_hal::{self as hal, gpio::{PinGroup, ReadPinHList, WritePinHList}}; // <-- unresolved imports `rp2040_hal::gpio::ReadPinHList`, `rp2040_hal::gpio::WritePinHList`
use embedded_hal::digital::OutputPin;
use frunk::HCons;

pub struct DataBus<H, T, BDIR, BC1>
where
    HCons<H, T>: ReadPinHList + WritePinHList,
    BDIR: OutputPin,
    BC1: OutputPin,
{
    // ( code that involves .set_u32() )
    ...
}
```

### DIFF
--- a/rp2040-hal/src/gpio/mod.rs
+++ b/rp2040-hal/src/gpio/mod.rs
@@ -55,7 +55,7 @@ mod pull;
 
 pub use func::*;
 pub use pin::{DynBankId, DynPinId, PinId};
-pub use pin_group::PinGroup;
+pub use pin_group::{PinGroup, ReadPinHList, WritePinHList};
 pub use pull::*;
 
 /// The amount of current that a pin can drive when used as an output.


jannic:
I don't really like exposing type-level trickery like this HList in the public API. But that ship has sailed: We already exposed it by defining the type bounds on the public PinGroup impl. So it seems to be only logical to also make ReadPinHList and WritePinHList publicly available.

Or am I missing something?

As these traits are sealed, other crates still can't implement these traits on arbitrary types, so we don't lose any type safety.

jannic (approved):

ithinuel (approved):
I think that was actually an oversight. Thank you for fixing it.

PR #992, opened 2026-04-11
Fix future incompatibility warning about fN::BITS

jannic:
See https://github.com/rust-lang/rust/pull/154065


thejpster:
TL;DR: We used a trait to add a `BITS` constant to `f32` and `f64` and now those types are gaining the constant directly, meaning our use of `f32::BITS` is ambigious (is it the constant from the trait, or from the type?).

LGTM

thejpster (approved):

PR #994, opened 2026-04-16
powman: add set_vreg_voltage, get_vreg_voltage, VRegVoltage enum

txleafeye:
changed: rp235x-hal/src/lib.rs, rp235x-hal/src/powman.rs
added: rp235x-hal/src/vreg.rs

Borrowed from @Altaflux example: https://github.com/Altaflux/gb-rp2350/blob/5c089fc386ad599193e4101f31e2d71d39146cd2/src/main.rs#L144

Tested on RP2350A2.

### DIFF
--- a/rp235x-hal/src/vreg.rs
+++ b/rp235x-hal/src/vreg.rs
@@ -1,0 +1,111 @@
+//! VREG voltage control types.
+
+/// Voltage selection for the on-chip VREG.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum VRegVoltage {
+    /// 0.55V
+    V550mV = 0b00000,
+    /// 0.60V
+    V600mV = 0b00001,
+    /// 0.65V
+    V650mV = 0b00010,
+    /// 0.70V
+    V700mV = 0b00011,
+    /// 0.75V
+    V750mV = 0b00100,
+    /// 0.80V
+    V800mV = 0b00101,
+    /// 0.85V
+    V850mV = 0b00110,
+    /// 0.90V
+    V900mV = 0b00111,
+    /// 0.95V
+    V950mV = 0b01000,
+    /// 1.00V
+    V1000mV = 0b01001,
+    /// 1.05V
+    V1050mV = 0b01010,
+    /// 1.10V (default)
+    V1100mV = 0b01011,
+    /// 1.15V
+    V1150mV = 0b01100,
+    /// 1.20V
+    V1200mV = 0b01101,
+    /// 1.25V
+    V1250mV = 0b01110,
+    /// 1.30V
+    V1300mV = 0b01111,
+    /// 1.35V
+    V1350mV = 0b10000,
+    /// 1.40V
+    V1400mV = 0b10001,
+    /// 1.50V
+    V1500mV = 0b10010,
+    /// 1.60V
+    V1600mV = 0b10011,
+    /// 1.65V
+    V1650mV = 0b10100,
+    /// 1.70V
+    V1700mV = 0b10101,
+    /// 1.80V
+    V1800mV = 0b10110,
+    /// 1.90V
+    V1900mV = 0b10111,
+    /// 2.00V
+    V2000mV = 0b11000,
+    /// 2.35V
+    V2350mV = 0b11001,
+    /// 2.50V
+    V2500mV = 0b11010,
+    /// 2.65V
+    V2650mV = 0b11011,
+    /// 2.80V
+    V2800mV = 0b11100,
+    /// 3.00V
+    V3000mV = 0b11101,
+    /// 3.15V
+    V3150mV = 0b11110,
+    /// 3.30V
+    V3300mV = 0b11111,
+}
+
+impl VRegVoltage {
+    /// Convert from raw register value to [`VRegVoltage`].
+    pub fn from_vsel_bits(bits: u8) -> Option<Self> {
+        match bits {
+            0b00000 => Some(VRegVoltage::V550mV),
+            0b00001 => Some(VRegVoltage::V600mV),
+            0b00010 => Some(VRegVoltage::V650mV),
+            0b00011 => Some(VRegVoltage::V700mV),
+            0b00100 => Some(VRegVoltage::V750mV),
+            0b00101 => Some(VRegVoltage::V800mV),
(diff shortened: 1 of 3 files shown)


txleafeye:
Comments on #757 (same PR but for RP2040) may be relevant here. 

The enum I put in rp235x-hal/src/vreg.rs could be moved to one of rp235x-pac/src/inner/powman/vreg*.rs, but those appear to be automatically generated (?)

https://github.com/rp-rs/rp235x-pac/blob/main/src/inner/powman/vreg_ctrl.rs

https://github.com/rp-rs/rp235x-pac/blob/main/src/inner/powman/vreg.rs

rp235x-hal/src/vreg.rs
```diff
@@ -0,0 +1,111 @@
...
+            0b10111 => Some(VRegVoltage::V1900mV),
+            0b11000 => Some(VRegVoltage::V2000mV),
+            0b11001 => Some(VRegVoltage::V2350mV),
+            0b11010 => Some(VRegVoltage::V2500mV),
+            0b11011 => Some(VRegVoltage::V2650mV),
+            0b11100 => Some(VRegVoltage::V2800mV),
+            0b11101 => Some(VRegVoltage::V3000mV),
+            0b11110 => Some(VRegVoltage::V3150mV),
+            0b11111 => Some(VRegVoltage::V3300mV),
+            _ => None,
```
ithinuel: I think this should be an error rather than None.

Maybe using `TryFrom` would even be better here.

ithinuel (changes requested):
Hey there, thank you for the contribution.

I think there’s a little room for improvement on the conversion from integer to the Enum type.

Also, the Datasheet says the output voltage is limited to 1.3V unless the `disable_voltage_limit` in `VREG_CTRL` is set.
I haven’t read in details not tested to see what would happen if we try it when conditions are not met.
When setting to 3.3V:
- Is the register updated but voltage remains at 1.3V?
- Is the register updated to the 1.3V setting and the voltage accordingly
- Is the write ignored and the regulator remains at its current setting?

Either way, IMHO this function should check that and return an error if the conditions are not met.

rp235x-hal/src/vreg.rs
txleafeye: That's certainly a little more ergonomic. Will apply that change.

txleafeye: Does it make sense, though, to create a new error type which should never occur except in the case of catastrophic hardware damage or PAC corruption? 

In rp235x-pac, vsel(&self) is defined as `VSEL_R::new(((self.bits >> 4) & 0x1f) as u8)`. In other words, a value not covered by the match (such as 0b100100) cannot be returned unless hardware fails or the PAC is re-written to remove that `& 0x1f.` `None` seems more appropriate.

txleafeye:
> Also, the Datasheet says the output voltage is limited to 1.3V unless the `disable_voltage_limit` in `VREG_CTRL` is set. I haven’t read in details not tested to see what would happen if we try it when conditions are not met. When setting to 3.3V:
> 
>     * Is the register updated but voltage remains at 1.3V?
>
>     * Is the register updated to the 1.3V setting and the voltage accordingly
> 
>     * Is the write ignored and the regulator remains at its current setting?
> 
> Either way, IMHO this function should check that and return an error if the conditions are not met.

I had to poke this a bit more. 

1. We can write whatever value we want to VSEL and it will return the same up to a cutoff of 1300 mV when the limit is enabled. Values written above 1300 mV just leave a value of 1300 mV in the register. Also, what we write to VSEL sticks even if UNLOCK is not set; it just doesn't do anything. In other words, the unlock bit doesn't gate VSEL writes.

2. If we do write the unlock and reset bits, the voltage we write does actually get set (up to 1300 mV if the voltage limit is not disabled) **and is activated immediately.** I have confirmed this a couple of ways: attempt to overclock at 1100 mV, find a frequency which just fails to run, and then set VSEL to 1300 mV whereupon it will not fail to run.

3. The PR at present does not actually set DISABLE_VOLTAGE_LIMIT. Should we add a separate setter for this? If I add it by default within the VREG setter, users could fry their chip in two lines: `let powman = Powman::new(); powman.set_vreg_voltage(3.3V)`

4. The PR at present doesn't provide a method to set the VREG_LP_ENTRY or VREG_LP_EXIT registers which are largely the same (except for the MODE values) as VREG. We could add those.

rp235x-hal/src/vreg.rs
ithinuel: I don’t know what users may be doing with it or where the data is coming from 🤷‍♀️.

All I can tell is that I can call `VRegVoltage::from_vsel_bits(some_value_taken_from_somewhere_else)`, where the only valid values are in `0..=31`.
If `some_value_taken_from_somewhere_else` breaks this predicate, I’d expect an error in return (or, at worst, a panic :( ).

In the PAC, the reader takes its value from the hardware, which is not expected to ever contain any invalid or unknown values.
And a writer can only be created with invalid values using an unsafe fn (`bits` or `write_with_zero`).

If the function were private and only used within that crate with proper input validation, I’d argue that an `unreachable!()` would even be acceptable.
But because it’s public, anyone can call it, and None (absence of a value) does not reflect the intent.

txleafeye: That's a reasonable point. I'll change it to a TryFrom.

txleafeye:
I've added error checking for set_vreg_voltage. If the value the user provides isn't successfully read back from VREG as it was written, we return a VRegError. I don't have any particularly innovative ideas for the `DISABLE_VOLTAGE_LIMIT` bit at present, beyond adding a shallow setter method, so I'll leave that for another contributor.

ithinuel:
> Should we add a separate setter for this?

Yes, I think that if we were to add that, it’d need to be behind an unsafe function or at least something that makes it harder to do that by mistake :o

But it doesn’t need to be added in This PR IMHO.

ithinuel (approved):
Thank you :) 

LGTM, not tested myself.

jannic:
Tested on a pico2 that it behaves as expected:
- set some legal voltages
- read them back
- try set a value > 1.3V
- check for correct error code
- check that get_vreg_voltage still returns the previous value

I did not measure the actual voltage, nor did I do any overclocking experiments.

PR #996, opened 2026-05-18
Add a method to expose the number of transfer remaining during DMA

peckpeck:
With current hal, during DMA transfer, you can only know if a transfer is finished or not with `Transfer::is_done()`.
However, the rp2040 provides a more fine-grained information with the trans_count register.

With those methods, this information is exposed and can be used to estimate remaining time until the end of the transfer.


jannic:
Thanks for the pull request, this looks like a good addition!

Could you run `cargo fmt` to fix the CI error, and apply the same change to rp235x-hal, to keep the APIs as similar as possible?

peckpeck:
Sure, done

jannic (approved):

PR #997, opened 2026-05-18
Add test for powman voltage selection

jannic:
Also adds some derives for the error types, so `.expect()` can be used easily.

### DIFF
--- a/on-target-tests/tests/powman.rs
+++ b/on-target-tests/tests/powman.rs
@@ -1,0 +1,87 @@
+#![no_std]
+#![no_main]
+#![cfg(test)]
+
+use defmt_rtt as _; // defmt transport
+use defmt_test as _;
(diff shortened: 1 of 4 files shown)


on-target-tests/tests/powman.rs
```diff
@@ -0,0 +1,87 @@
...
+    use rp235x_hal::powman::{Powman, VRegError};
+    use rp235x_hal::vreg::VRegVoltage;
+
+    #[init]
+    fn setup() -> () {
+        unsafe {
+            crate::init::reset_cleanup();
+        }
+        let mut pac = pac::Peripherals::take().unwrap();
+        #[cfg(feature = "rp2040")]
```
thejpster: This test only runs on 2350?

thejpster (approved):

PR #998, opened 2026-05-24
Implement embedded_hal_timer::Timer

jannic:
Open questions:
- Naming
- Should we feature-gate this behind some `unstable` feature for now?
- This doesn't detect overflows of the u64 tick counter. Most likely not an issue in practice, for microsecond ticks. But it should at least be documented.
- Needs a MSRV bump because embedded-hal-timer requires edition2024. (See failing CI)

Before merging this, it should also be implemented for rp235x-hal.

### DIFF
--- a/rp2040-hal/src/timer.rs
+++ b/rp2040-hal/src/timer.rs
@@ -9,6 +9,7 @@
 //! See [Chapter 4 Section 6](https://datasheets.raspberrypi.org/rp2040/rp2040-datasheet.pdf) of the datasheet for more details.
 
 use core::sync::atomic::{AtomicU8, Ordering};
+use embedded_hal_timer::OverflowError;
 use fugit::{MicrosDurationU32, MicrosDurationU64, TimerInstantU64};
 
 use crate::{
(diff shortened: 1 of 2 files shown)


rp2040-hal/src/timer.rs
```diff
@@ -542,3 +543,81 @@ pub mod monotonic {
...
+
+    fn max_ticks(&self) -> u64 {
+        u64::MAX
+    }
+
+    fn max_nanos(&self) -> u64 {
+        u64::MAX
+    }
+
+    fn max_micros(&self) -> u64 {
```
thejpster, line 609: 2^64 microseconds is 584 millenia (or 584,542 years). I think we're fine to claim that overflows are impossible, and ignore the -1 here.

rp2040-hal/src/timer.rs
```diff
@@ -542,3 +543,81 @@ pub mod monotonic {
...
+
+/// Implementation of embedded_hal_timer::Timer
+pub struct TickCounter {
+    timer: Timer,
+    start_time: u64,
+}
+
+impl TickCounter {
+    /// Creates a new TickCounter.
+    pub const fn new(timer: Timer) -> Self {
```
thejpster, line 555: Does it need to own the timer? It seems like it would be OK to make a bunch of these, and have them unsafely conjure up the timer for reading whenever required? Or just have a freestanding function called `fn get_time() -> u64` that all these types can use?

PR #1000, opened 2026-06-09
add support for QFN 80 package in rp235x-hal

maxaudron:
The checks on pin numbers for the ADC were hardcoded to the values of the QFN 60 package. I factor this out into consts and introduce a feature flag to enable the QFN 80 pin numbers instead.

Open to some suggestions for handling the feature flags in a nicer way. Having to have them on each channel! etc. is a bit ugly


maxaudron:
Fixed the formatting issues

thejpster:
The datasheet says:

> Figure 108. ADC Connection Diagram for QFN-80. This package features eight external ADC inputs (0 through 7), on Bank 0 GPIOs 40 through 47. The internal temperature sensor connects to a ninth channel (channel 8).

So this looks good to me. And I'm fine with the use of `cfg`.

thejpster (approved):

PR #1001, opened 2026-07-16
Correct ranges of PLL dividers for both rp2040 and rp235x

ElectronicsTinkerer:
Currently in `main` the `POSTDIV` and `FBDIV` divider ranges for both the rp2040 and rp235x PLLs use a noninclusive range. Based on the reference code for calculating these parameters (sections 2.18.2.1 and 8.6.3.2, of their respective datasheets), these should be inclusive ranges on the upper bound.


thejpster (approved):
Good catch. Thanks!

Issue #993, opened 2026-04-11
Set MSRV to latest stable


thejpster:
We're fighting a losing battle with our dependencies because they keep needing a newer Rust than our MSRV.

I propose we just set the MSRV to latest stable and be done with it. And at the same time, set any clippy jobs to not "Deny all Warnings" because new warnings come along all the time and I don't a new warning to break the build (but I do want to know about them so I can choose to fix them if required).

jannic:
Isn't the dependency issue mostly solved with [this change](https://github.com/rp-rs/rp-hal/pull/991)?

In my opinion, MSRV should only be increased when it is needed to get some relevant advantage.
That could be a new feature we want to use. Either in the language itself, or in some updated dependency that needs the newer Rust version.

Of course, the old situation where CI broke every now and then because some dependency decided to bump the MSRV in a minor version was very unsatisfying. This should be solved by the new resolver mode. Sure, we may learn that it doesn't work as well in practice (e.g. when crates we depend on start to require newer Rust versions implicitly, without adjusting the declared MSRV). But I'd prefer to actually try that out and see how well it works.

Regarding the clippy warnings: Here I don't care too much. We should strive for a warning-free build either way, but I agree that it is really annoying when a new clippy lint breaks the build. Locally, I use the beta compiler by default, so I tend to notice new warnings before they reach our CI and can send fixes proactively. (E.g. https://github.com/rp-rs/rp-hal/pull/992) 
I looked at all the failed pipeline runs in https://github.com/rp-rs/rp-hal/actions?query=is%3Afailure where the logs are still available, and I didn't see a single case where a new lint broke the build. Is this really a problem we need to solve?

thejpster:
Ok, the new resolver might be enough. Happy to try it out for a while!

PR #1002, opened 2026-08-04
Add TRNG support for RP235x

cvvletter:
First time contributor here!

For a project, I am using the TRNG peripheral on the RP235x chip, but I noticed it only had PAC and not HAL support. Thus, I decided it might be useful to upstream this code. This PR adds TRNG support and implements `rand_core`'s traits just like already implemented for the ROSC ([`rp235x-hal/src/rosc.rs`](./rp235x-hal/src/rosc.rs)).

I debated using a typestate pattern like the ROSC, but decided against that for simplicity for now. Would that add a lot of value? E.g. the following code can now be run which will hang indefinitely:
```rust
let mut trng = Trng::new(pac.TRNG, &mut pac.RESETS);
trng.source_disable();
trng.next_u32();
```

Do let me know if any changes are necessary!


rp235x-hal/src/trng.rs
```diff
@@ -0,0 +1,152 @@
...
+            self.trng.ehr_data1().read().ehr_data1().bits(),
+            self.trng.ehr_data2().read().ehr_data2().bits(),
+            self.trng.ehr_data3().read().ehr_data3().bits(),
+            self.trng.ehr_data4().read().ehr_data4().bits(),
+            self.trng.ehr_data5().read().ehr_data5().bits(),
+        ];
+
+        // Convert 6 u32 values to 24 bytes (192 bits)
+        let mut out = [0u8; 24];
+        for (i, &word) in data.iter().enumerate() {
```
thejpster: I would prefer to see a u8 array created first, and filled in using array_chunks (or chunks_mut_exact if our MSRV is too low) rather than putting two arrays on the stack.

rp235x-hal/src/trng.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /// Fill a buffer with random bytes.
+    ///
+    /// This reads 192-bit chunks from the TRNG until the buffer is filled.
+    /// Each 192-bit read is guaranteed to be fresh data since reading the
+    /// last register clears all EHR_DATA registers.
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), Self::Error> {
+        let mut offset = 0;
+        while offset < dest.len() {
```
thejpster: I think you could use a chunks iterator here?

thejpster:
Looks great to me. Just some minor notes.

rp235x-hal/src/trng.rs
cvvletter: Done. A bit hacky maybe with the match statement but it _is_ nicer.

rp235x-hal/src/trng.rs
cvvletter: Good point, much better!

cvvletter:
Thank you. Feedback processed. What do you think about implementing the typestate pattern like the ROSC? And is the failing build due to my changes?

9names:
The build is failing because the newer randcore requires rust 2024 edition, but our minimum supported rust version (msrv) is too old for that.
It's probably time to increase our msrv.

cvvletter:
Ah, thanks. So that's probably related to https://github.com/rp-rs/rp-hal/issues/993 then, right?

thejpster:
Happy to see the MSRV bumped in this PR

cvvletter:
It took me some tries but I think I now updated the MSRV *and* the check for the MSRV. :laughing:

9names (approved):

cvvletter:
We merged a change into the `rp235x-pac` crate which will make the code here much nicer (no more need for the `match` statement). Once that code is in a new release of the crate, we will change the code here too.

PR #1003, opened 2026-08-22
Add a binary info macro for specifying program description from the user's Cargo.toml package description

twokilohertz:
Similarly to the other `rp_program_xxx` & `rp_cargo_xxx` macros: this macro allows for the specification of picotool binary info entries for the program's description expanded from the `CARGO_PKG_DESCRIPTION` environment variable from the user's `Cargo.toml`.

### DIFF
--- a/rp-binary-info/src/macros.rs
+++ b/rp-binary-info/src/macros.rs
@@ -146,6 +146,19 @@ macro_rules! rp_program_description {
     };
 }
 
+/// Generate a static item containing the `CARGO_PKG_DESCRIPTION` as the program description,
+/// and return its [`EntryAddr`](super::EntryAddr).
+#[macro_export]
+macro_rules! rp_cargo_description {
+    () => {
+        $crate::env!(
+            $crate::consts::TAG_RASPBERRY_PI,
+            $crate::consts::ID_RP_PROGRAM_DESCRIPTION,
+            "CARGO_PKG_DESCRIPTION"
+        )
+    };
+}
+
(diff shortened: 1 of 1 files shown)


rp-binary-info/src/macros.rs
```diff
@@ -81,6 +81,19 @@ macro_rules! rp_cargo_bin_name {
     };
 }
 
+/// Generate a static item containing the `CARGO_PKG_DESCRIPTION` as the program description,
+/// and return its [`EntryAddr`](super::EntryAddr).
+#[macro_export]
+macro_rules! rp_cargo_description {
```
thejpster: Can we also add one where the user can set the `ID_RP_PROGRAM_DESCRIPTION` property to a string literal? Most of the macros have both a "get this value from `cargo` through the environment" and a "use this literal string" versions, and I think that makes sense here too.

twokilohertz: Good shout I'll implement one of those too

twokilohertz: Implemented in [b41f948](https://github.com/rp-rs/rp-hal/pull/1003/commits/b41f94839c1b1b38af8aa7925f18673a43f35e73)

thejpster:
Oops. See CI failures.

twokilohertz:
> Oops. See CI failures.

There we are, missed the fact that we already had a `rp_program_description!` macro, oops

thejpster (approved):

thejpster:
Thanks!

PR #1004, opened 2026-09-01
Update pio version to 0.4.0

9names:
Update to pio 0.4.0 since it was just released.
Dropped pio_core as a dependency.
Need to think about if we want to re-export it like embassy-rp does, and what we would call it (since `pio` is already taken by the hal).
I guess we could just do a re-export of the re-export?


thejpster:
I'm happy to merge the update now and worry about re-exporting the crate later.

thejpster (approved):

PR #1005, opened 2026-09-03
update fugit dependency to 0.5.0

rene0:
Update the fugit dependency to 0.5.0 (latest version). The API changed somewhat, so adjust relevant code.

- Bug https://github.com/korken89/fugit/issues/50 was fixed, so the constuctor in rp235x-hal/src/powman.rs could be simplified? (not done here).
- only runtime-tested on an rp2040 (Raspberry Pi Pico, non WiFi-version) as I don't have an rp235x, but the vector_table example works fine.
- the "write" test of on-target-tests/tests/i2c_loopback.rs seems to hang on my rp2040 (wiring is correct, and using a second Pico as probe-rs host)


rp235x-hal/src/powman.rs
```diff
@@ -36,17 +36,18 @@ pub struct FractionalFrequency(u32);
 
 impl FractionalFrequency {
     /// Create a new Fractional Frequency from any `fugit::Rate`.
-    pub const fn new<const NOM: u32, const DENOM: u32>(
+    pub const fn new<const NOM: u64, const DENOM: u64>(
         rate: fugit::Rate<u32, NOM, DENOM>,
```
thejpster, line 40: I think this counts as a breaking change? I think it's fine, but something to note.

thejpster (approved):
Looks OK to me, but I think we'll need a major version bump for this.

rp235x-hal/src/powman.rs
rene0: This change was because fugit changed the types to u64 to acomodate for picosecond durations.

rene0:
- Not sure what the above conflicts in vector_table.rs are about, I don't see any local conflicts?
-  I still need to update the alarm code in the last commit to fix the rtic-monotonic feature , how much freedom is there (e.g. introducing functions)?
- In the meantime, fugit 0.6.0 has been released, should that be in its own pull request?

thejpster:
That might be because the examples got renamed from foo_bar.rs to foo-bar.rs. You'll need to rebase.

thejpster:
Adding functions is fine - we just don't want to take away functionality and cause a breaking change if we can avoid it.

I'm fine with going straight to fugit 0.6 if that makes sense.

rp235x-hal/src/powman.rs
rene0: The change was made in fugit 0.4.0. An u32 always fits in an u64 so existing consumers should be fine, or am I missing something?

PR #1006, opened 2026-09-06
Clean up new nightly warnings

thejpster:
- Remove the `package.homepage` property because Rust 1.100 generates a warning when it's the same a `package.repository`
- Rename all binaries to use `kebab-case` because Rust 1.100 generates a warning about using `snake_case`, and fix all references to those binaries

Includes #1004


9names (approved):
I'm happy with the changes. Might want to rebase on main and drop the pio commits that have already merged to main though?

I'll leave that up to you to decide.

thejpster:
rebased and merged.

PR #1011, opened 2026-09-15
Fix rp_program_version and doctests

mkj:
rp_program_version didn't work, the doctests now test it too.


thejpster (approved):
Well. I was clearly having a bad day when I wrote those examples.

thejpster:
Thank you for the fixes!

